PR #17, opened 2018-05-01
Updates for stacked coordinates.

jmilloy:
Updates convert_xarray_to_podpac and ReprojectedSource to handle stacked coordinates after the refactor.

These changes seem straightforward enough, but it is not in code that I've had much to do with or can test well.


mpu-creare:
I'll still need to test this more thoroughly, but at the root it doesn't seem to change functionality, only add to it, so I think it's pretty safe.

Issue #42, opened 2018-05-21
Fix lat_lon order in Coordinate


mpu-creare:
Its used to be ((lat_start, lon_start), (lat_end, lon_end), Number of points)
Now it's ((lat_start, lat_end), (lon_start, lon_end), Number of points)

mlshapiro:
this may be a little off topic, but related to the tile.

Shoud `lat_lon` be input as a `list(tuple(lat, lon))` ie. `[(lat, lon), (lan, lon)]` - it seems like the list of points is not necessarily immutable on the client (though certainly it is immutable within podpac). 

This is related to clearly bounding the inputs to each method. Feel free ignore if this is perpendicular to the focus.

jmilloy:
BTW `((lat_start, lon_start), (lat_end, lon_end), size)` makes a lot of sense... It would be a consistent `(start, stop, size)` for whether you're doing a single coordinate `lat` or stacked `lat_lon_time`.

As for `lat_lon` as an array, I would rather it accept an array in the shape (n, 2) -- like @mlshapiro has. I don't think we would need to force the argument to be a list of tuples, it could accept an array, a list of lists, a list of tuples, an array of tuples, whatever. But I would prefer that to a pair of lists or pair of arrays. Basically the same as above - more closely coupling the lat/lon pairs when stacked.

Thirdly, I don't really like how the stacked coordinates are stored as a structured array in the Coordinate object. It has really been frustrating to deal with, and I don't really see any benefits. It's related to the above because I would rather they were *also* an array of shape (n, 2).

Fourthly, I still suggest that we require Coordinate initialization to accept Coord objects, e.g. `Coordinate(lat_lon=(UniformCoord(...), UniformCoord(...)), time=Coord('2018-01-01'))` and handling the variety of user-friendly shortcuts in a separate method that is easy to document, test, and read. `unstack_dict` also muddles two separate tasks, and it would be simplified as well by this change. Maybe I'll throw together a pull request. I think it would simplify the user-facing aspects (usage, docs) of the Coordinate class and the internals (testing, actual implementation).

mpu-creare:
I agree with all of the above. 1 small point: 
* I don't like the structured array either, it's cumbersome to work with
* The structured array is for xarray's benefit. This allows xarray to do all its magic seamlessly
* When I started this, xarray didn't have a mechanism to define a coordinate from two coordinates
    * I think that's no longer the case? I think they now have some sort of stacking as well? 
[...]

PR #44, opened 2018-06-01
WIP Feature/simplify coordinate init

jmilloy:
@mpu-creare would you have a look at this and let me know what you think?

 * The `Coordinate` object is initialized only with a dictionary (OrderedDict in Python 2) that maps unstacked dims to a BaseCoord object and stacked dims to a tuple of BaseCoord objects. Generally we use the Coordinate object directly within podpac.

 * The `coordinate` function is for user-friendly coordinate creation. It takes named dimension arguments, parses the coordinate values, and returns a `Coordinate` object. Users will generally use this function to crete Coordinates -- it is the only thing from the coordinate submodule that is available in the top-level namespace as `podpac.coordinate`.

I think this makes the code, testing, and docs much simpler. I've updated much of the Coordinate testing.

The one thing I'm hesitant about is the names `Coordinate` and `coordinate` since they are so similar. I wondered if something like `MultiCoordinate` and 'coordinate' would be better to avoid confusion. Sort of like `np.array` vs `np.ndarray`, you know? What do you think about that?

### DIFF
--- a/podpac/core/algorithm/algorithm.py
+++ b/podpac/core/algorithm/algorithm.py
@@ -5,6 +5,7 @@ Algorithm Summary
 from __future__ import division, unicode_literals, print_function, absolute_import
 
 import inspect
+from collections import OrderedDict
 import numpy as np
 import xarray as xr
 import traitlets as tl
@@ -176,7 +177,7 @@ class CoordData(Algorithm):
        
         c = ec[coord_name]
         data = c.coordinates
-        coords = Coordinate(order=[coord_name], **{coord_name: c})
+        coords = Coordinate(OrderedDict(**{coord_name:c}))
         return self.initialize_coord_array(coords, init_type='data', fillval=data)
 
 
@@ -281,9 +282,9 @@ class Arithmetic(Algorithm):
         return d
         
 if __name__ == "__main__":
+    import podpac
     a = SinCoords()
-    coords = Coordinate(lat=[-90, 90, 1.], lon=[-180, 180, 1.], 
-                        order=['lat', 'lon'])
+    coords = podpac.coordinate(lat=[-90, 90, 1.], lon=[-180, 180, 1.], order=['lat', 'lon'])
     o = a.execute(coords)
     a2 = Arithmetic(A=a, B=a)
     o2 = a2.execute(coords, params={'eqn': '2*abs(A) - B'})
--- a/dist/aws/handler.py
+++ b/dist/aws/handler.py
@@ -79,7 +79,7 @@ def handler(event, context, get_deps=True, ret_pipeline=False):
     # Need to set matplotlib backend to 'Agg' before importing it elsewhere
     import matplotlib 
     matplotlib.use('agg')
-    from podpac import Coordinate
+    from podpac import coordinate
     from podpac.core.pipeline import Pipeline
     try:
         pipeline = Pipeline(source=pipeline)
@@ -87,7 +87,7 @@ def handler(event, context, get_deps=True, ret_pipeline=False):
         w, s, e, n = np.array(bbox, float)     
         dwe = (w-e)/width/2.
         dns = (n-s)/height/2.
-        coord = Coordinate(lat=(n-dns, s+dns, height),
+        coord = coordinate(lat=(n-dns, s+dns, height),
                            lon=(w+dwe, e-dwe, width),
                            time=np.datetime64(time), 
                            order=['time', 'lat', 'lon'])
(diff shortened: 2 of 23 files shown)


jmilloy:
Note I didn't change the spec for creating stacked coordinates in the `coordinate` function.

 - The basic format is `dim1_..._dimN=(dim1_values, ..., dimN_values)`, i.e. a tuple matching the number of stacked dimensions, with values that are parsed as coord_linspace args, UniformCoord args, MonotonicCoord args, or Coord args as appropriate and that must have the same size.
 - In addition, you can use the linspace shortcut `dim1_...dimN=((dim1_start, dim1_end), ..., (dimN_start, dimN_end), size)`, i.e. a tuple 1 element longer than the number of stacked dimensions.

mlshapiro:
@jmilloy i like the `coordinate` function as a shorthand definition. Its not much different than instantiating the class, but I do think that functions are more accessible to many non-technical users, which will help.

I also really like the py3.6 catching for order - this is more intuitive. I see your comment on #42 - it makes sense why you would want to keep the coordinate definitions for stacked coordinates the way it stands.

mlshapiro:
@jmilloy i'm trying to make an example notebook based on some of  our discussions, but I *think* I found a bug. 

I get the error below when trying to get one data source a a second's coordinate. Example here: [data-source.txt](https://github.com/creare-com/podpac/files/2069598/data-source.txt)  I may be at fault with the example, but if you can tell me what I'm doing wrong that would be just as helpful.

```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-25-c329a7e61be2> in <module>()
      3 
      4 Grid_at_Point = GridData.execute(point_coords)  # get the grid data at the nearest coordinate the point data is defined
----> 5 Point_at_Point = PointData.execute(point_coords)  # get the grid data at the grid data coordinates (?? Is this necessary ??)

~/computing/repositories/geospatial/podpac/podpac/core/data/data.py in execute(self, coordinates, params, output)
    106 
    107         # sets self.output
--> 108         self.interpolate_data(data_subset, coords_subset, self.evaluated_coordinates)
    109 
    110         # set the order of dims to be the same as that of evaluated_coordinates

~/computing/repositories/geospatial/podpac/podpac/core/data/data.py in interpolate_data(self, data_src, coords_src, coords_dst)
    292         elif 'lat_lon' in coords_src.dims or 'lon_lat' in coords_src.dims:
    293             return self.interpolate_point_data(data_src, coords_src,
--> 294                                                data_dst, coords_dst)
    295 
    296         return data_src

~/computing/repositories/geospatial/podpac/podpac/core/data/data.py in interpolate_point_data(self, data_src, coords_src, data_dst, coords_dst, grid)
    519             new_crds = coordinate(**{key: [coords_dst.unstack()[c].coordinates for c in key.split('_')]})
    520             tol = np.linalg.norm(coords_dst.delta[i]) * 8
--> 521             src_stacked = np.stack([c.coordinates for c in coords_src.stacked_dict[key]], axis=1)
    522             new_stacked = np.stack([c.coordinates for c in new_crds.stacked_dict[key]], axis=1)
    523             pts = KDTree(src_stacked)

AttributeError: 'Coordinate' object has no attribute 'stacked_dict'
```

podpac/core/algorithm/algorithm.py
```diff
@@ -176,7 +177,7 @@ def algorithm(self):
        
         c = ec[coord_name]
         data = c.coordinates
-        coords = Coordinate(order=[coord_name], **{coord_name: c})
+        coords = Coordinate(OrderedDict(**{coord_name:c}))
```
mpu-creare: Why not just 
`OrderedDict((coord_name, c))` ?

jmilloy: Yeah, that's better. I think you would need `OrderedDict([(coord_name, c)])`.

mpu-creare:
I'll check this out and try it... won't have time till next week. Nothing really jumps out at me. I think it looks fine in general. I can live with keeping track of both coordinate and `Coordinate` -- do that with numpy right now anyway. I do agree we should rename one of them.... how about `NDCoordinate` to follow numpy?

jmilloy:
NDCoordinate is great. But I think we should wait for the spec and can keep working in this branch to finish that out.

mpu-creare:
I was hoping to just merge this in and forge ahead, but I think you're correct. Let's sit on this one until we have a good spec. We'll also be able to merge in develop as unit tests are developed. That way we can finish off the `Coordinates` object, or at least backwards incompatible changes, in one go.

PR #55, opened 2018-06-08
Feature/traitlets refactor

mpu-creare:
This modifies the way we use Traitlets. It inegrates more tightly with traitlets, requiring traits to be tagged. This tagging can the be used to more accurately and robustly create pipelines. 

Also adding a large amount of documentation in the PR.

### DIFF
--- a/podpac/core/algorithm/algorithm.py
+++ b/podpac/core/algorithm/algorithm.py
@@ -16,42 +16,40 @@ except:
 
 from podpac.core.coordinate import Coordinate, convert_xarray_to_podpac
 from podpac.core.node import Node
+from podpac.core.node import COMMON_NODE_DOC
+from podpac.core.utils import common_doc
 
+COMMON_DOC = COMMON_NODE_DOC.copy()
 
 class Algorithm(Node):
-    """Summary
+    """Base node for any algorithm or computation node. 
     
-    Attributes
-    ----------
-    evaluated : bool
-        Description
-    evaluated_coordinates : TYPE
-        Description
-    output : TYPE
-        Description
-    params : TYPE
-        Description
+    Notes
+    ------
+    Developers of new Algorithm nodes need to implement the `algorithm` method. 
     """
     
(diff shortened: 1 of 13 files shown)


mpu-creare:
I'd like this one merged in sooner than later so we can start working using the same methodology.

jmilloy:
@mpu-creare This was the first thing I tried to do.

```
from podpac import Coordinate
from podpac.core.algorithm.algorithm import CoordData 
n = CoordData(coord_name='lat')
c = Coordinate(lat=[1, 2, 3, 4])
n.execute(c)
```

Here's the result:

```
  File "podpac/core/algorithm/algorithm.py", line 70, in execute
    result = self.algorithm()
  File "podpac/core/algorithm/algorithm.py", line 162, in algorithm
    coord_name = self._params['coord_name']
KeyError: u'coord_name'
```

Are the tagged attrs supposed to be loaded in the the `_params`, too? Does get_params need to be `if value.metadata.get('param', False) or value.metadata.get('attr', False)`? Or is it something else?

jmilloy:
@mpu Well, no, actually I'm guessing that either

 - `coord_name` should be tagged as a param, **or**
 - it should be `coord_name = self.coord_name` (instead of `coord_name = self._params['coord_name']`)

You *could* load attributes into the params safely if you did it in the right order, but I don't think that was your intent? Like this:

```
def get_params(self, params=None):
        p = {}
        for key, value in self.traits().items():
            if value.metadata.get('param', False):
                p[key] = getattr(self, key)
        if params: 
            p.update(params)
        for key, value in self.traits().items():
            if value.metadata.get('attr', False):
                p[key] = getattr(self, key)
        return p
```

mpu-creare:
Yep, that was another mistake on my part. `attrs` are NOT supposed to be loaded into the params dictionary. It should have be `self.coord_name`. I'll fix that and the ReprojectedCoords one.

mpu-creare:
Agreed. Feel free to start a doc stub, or assign that task to me.

PR #64, opened 2018-06-15
Feature/test node

mpu-creare:
Adding unit tests for the Node class. 

I wouldn't say this is complete, but its a start.


mpu-creare:
Merging. Checks failed because develop is currently failing.

PR #67, opened 2018-07-07
Feature/test data

mlshapiro:
@mpu-creare please check to make sure my changes to data.py and type.py don't affect critical infrastructure at this point

test_type definitely still needs work, especially a relevant example for WCS and ReprojectedSource. 

I would like to go through the DataSource/Interpolation spec and refactor process before I finish these tests. We're up to 76% for now, which really is >90% if you add in the lines that don't count much (imports, urllib3 implementation)


mpu-creare:
I think just fixing up the unit tests that are failing should suffice. 
I tested a few things and your changes didn't seem to break anything. 
I need to revisit the SMAP node anyway -- it might be slightly broken.

mlshapiro:
tests fixed - ready to merge?

mpu-creare:
Looks like the tests are still failing.

PR #75, opened 2018-07-26
Omit tests and datalib from coverage calculations

jmilloy:
I think this does what you want. We round out at 85%.

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,2 +1,0 @@
-[run]
-omit = */test_*.py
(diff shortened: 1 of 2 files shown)


mpu-creare:
Ok, so it looks like when we do not count the tests, the coverage actually decreases... ugh, great. Thanks.

jmilloy:
@mpu-creare I don't really get it, it correctly omitted the tests *and* the datalib in one of these coverage reports, but now it's only omitting the tests, but not omitting the datalib. When we get that working properly again, it will be at ~85%.

PR #95, opened 2018-08-21
WIP Implement Coordinate Spec

jmilloy:


jmilloy:
@mpu-creare We're removing the delta from `ArrayCoordinates1d` and `MonotonicCoordinates1d`, but I'm not sure what to do with the `area_bounds` which uses it. In fact, that made me wonder about the the 'segment' ctype for `ArrayCoordinates1d` and `MonotonicCoordinates1d` in general. To me those would only really make sense as 'point' ctypes, **or** if they are 'segments' then a `segment_width` should also specified along with the `segment_position`. Does that make sense? Should I implement the `segment_width` property or force these to be point types?

jmilloy:
@mpu-creare What do you think about

```
c1 = ArrayCoordinates1d(['2018-01-01', '2018-01-02'], name='lat')
c2 = ArrayCoordinates1d([0, 1, 2], name='time')
```

Obviously `c1` is wrong, should we validate that? And is `c2` wrong also, or is it okay to use numerical values to represent time in some cases?

mpu-creare:
Sure, let's validate on the lat-lon-alt dimensions. 
I'm ok with time being ints or floats. You could potentially have a local time system. We don't handle timezones or anything like that -- we require our users to take care of that themselves.

jmilloy:
@mpu-creare Where do you want the aliases? There's no risk of cluttering the namespace because they are underscores. Which of these do you want?

```
from podpac import _cu
from podpac.coordinates import _cu
from podpac.core.coordinates import _cu
from podpac.core.coordinates.coordinates1d import _cu
```

jmilloy:
@mpu-creare I guess maybe if you expect to use them after just importing podpac, the only ones that makes sense are these two

```
import podpac

podpac._cu
podpac.coordinates._cu
```

mpu-creare:
Yep, top level please. 

I'd almost vote for a second location so that you can do: 
`from podpac.coord_shortucts import *`
But I despise `import *`, so I won't vote for it.

mlshapiro:
my 2c: i'm not sure you need a `..._shortcuts` at this point, it feels like top level is fine for now.

jmilloy:
I think `from ... import *` doesn't import underscore-variables anyways. Top level it is.

jmilloy:
I renamed the feature branch without the spelling typo.

PR #97, opened 2018-08-22
Implement coordinate spec

jmilloy:


jmilloy:
Migrated. Notes:
 * validate dtype is numerical on lat, lon, and alt coordinates (either dtype is okay for time)
 * toplevel aliases, e.g. `podpac._cu_lat`

PR #101, opened 2018-08-30
WIP: Feature/data source

mlshapiro:
@mpu-creare I am almost finished with the DataSource refactor. This is largely semantical, but it has far reaching implications (mostly because of the changed from `self.evaluated_coordinates` to `self.requested_coordinates`)

I wanted to file this PR now so that you can review before you leave. What I still have left to do will not change the functionality or interfaces.  Tests are passing on my machine in py 35 - I will need to let the CI tell me if they are passing on py 27 since I have not yet set up a py 27 podpac environment locally (will do this shortly).

If you can test your consuming data sources (SMAP, etc) or point me towards people who can test this for me next week, I will make sure to merge after those data sources are working.

@jmilloy the only issue with merging with your coordinate refactor may be related to changing `self.evaluated_coordinates` all over the place. 

Still to do:

- Include more tests not that the structure of private methods has changed
- Move out `interpolate_...` methods function into `interpolator` - this is purely organizational to slim down the DataSource and get all relevant methods ready  in the `interpoloate.py` file.


mpu-creare:
This looks fine to me. Note that DataSource will also be affected by the Node refactor, but I think this can get merged in without any problem at this point.

mlshapiro:
Yes - I will fix on the node refactor and for @jmilloy's coordinate refactor. I will be in charge of the merge

PR #108, opened 2018-09-28
add `is_stacked` property to Coordinates1D

mlshapiro:
@jmilloy  i would find it useful to have this property on Coordinates1D when they are part of a larger Stacked Coordinates (a way to see if its stacked while looking at a single dimension) 

Did I do this correctly? It seems to work for my purposes


jmilloy:
I don't understand how you are able to set `c.is_stacked` that way if it is readonly, but I'll try it out, I guess I'm missing something. I also think it should only a property of Coordinates1d, not BaseCoordinates.

I'm curious when you need this, though, if you can point me to an example? I'd love to show do it a different way if possible.

mlshapiro:
Don't need this anymore - abandoning for now

Issue #104, opened 2018-09-17
Test and update SMAP with regard to Coordinates refactor


jmilloy:
**Description**
The Coordinates API was updated in #84 but SMAP has not been tested. I suppose it is possible that there are no issues.

**Steps to Reproduce**

Run the SMAP notebooks.

jmilloy:
@mpu-creare There are no SMAP tests. I haven't looked at the code, but I assume that there are unaddressed issues in the SMAP nodes. What is the best way for me to test SMAP at this point? Or would you like to assign this to someone else, anyways?

mpu-creare:
Best way to test SMAP usage is to run the notebooks.

jmilloy:
Currently the SMAPDateFolder(OrderedCompositor) uses `Coord` with a delta. `Coord` has been replaced by ArrayCoordinates1d which does not support a delta argument.

@mpu-creare wrote:

> There's this nasty bug that I never really fully resolved. When you request a single lat_lon coordinate that doesn't align exactly with one of the source coordinates, you sometimes get nothing. I.e. the compositor doesn't find an appropriate source.
> 
> What the delta does, is extend the 'bounds' of each coordinate, such that the intersection will be True if such a lat_lon coordinate is requested. So, instead of representing a single point, each source coordinate represents a rectangle. That's what this delta is doing.
> 
> This problem was very specific to the Smap-Sentinel product. So you could just take out all of this logic , request a lat_lon point that doesn't align with a tile (But should) and then try to fix the problem.

I will replace the Coord usage now, but closing this issue will require reproducing and fixing this case.

mpu-creare:
Does `ArrayCoordinates1d` support both `point` and `segment` ctypes? Because essentially this should be a `segment` ctype with a specified segment length.

jmilloy:
Yes, it supports segment ctypes, but there is not segment length attribute yet.

ArrayCoordinates1d handles non-monotonic and monotonic coordinate arrays. Segment ctypes are only supported for monotonic arrays, where the segment length is inferred from the distance between coordinate values. But segment ctypes are not supported for non-monotonic arrays. I would happily add a segment_length attribute in order to support segment ctypes (I think I proposed that at one point but we decided not to do it).

I want to reproduce, though... it seems possible to me that this could be handled though `extents`, ie that the only time the intersect returns nothing is when the requested point is outside the area_bounds. Currently the extents attribute is ignored for point ctypes, but that could be changed.

[...]

PR #109, opened 2018-10-04
Updates some SMAP Coordinates usage

jmilloy:


mpu-creare:
Does this close #104 ?

mpu-creare:
There's this nasty bug that I never really fully resolved. When you request a single lat_lon coordinate that doesn't align exactly with one of the source coordinates, you sometimes get nothing. I.e. the compositor doesn't find an appropriate source. 

What the `delta` does, is extend the 'bounds' of each coordinate, such that the intersection will be `True` if such a lat_lon coordinate is requested. So, instead of representing a single point, each source coordinate represents a rectangle. That's what this delta is doing. 

This problem was very specific to the Smap-Sentinel product. So you could just take out all of this logic , request a lat_lon point that doesn't align with a tile (But should) and then try to fix the problem.

PR #132, opened 2018-10-29
Feature/interpolators

mlshapiro:
To do before merge:

- @jmilloy review the Coordinates `__eq__` implementation. Do we wan to keep this here or move to a utils method (i.e. `coordinates_equal()`) . If we move, then i'll update Interpolation.interpolate to reflect
- Test SMAP - update for temporal tolerance
- Review and update spec (?) and documentation
- Review rasterio interpolation - notebook example is providing strange results


mpu-creare:
I think for now being very strict (including order) makes sense. 

I think your suggestion of implementing `__eq__` in child classes also makes sense (for optimization purposes).

mlshapiro:
Interesting - so Scipy RegularGridInterpolator works a little different than rasterio when you request data outside of the bounds of the native coordinates. In rasterio (and our) implementation we include the outer data point if the the requested data is outside the bounds of the native coordinates. Scipy fills this. I'm going to make a note of this but not let it stop me getting this implemened . 

See https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.RegularGridInterpolator.html#scipy-interpolate-regulargridinterpolator

mlshapiro:
note there is a nuance with how i've designed the interpolators that some interpolators may support params, while others will not.  For example, the `NearestNeighbor` interpolator supports spatial_tolerance and time_tolerance, but `Rasterio` and `ScipyGrid` don't currently support these parameters. 

There will be some holes to fill after this merge

mlshapiro:
alright tests are finally passing (again)

mlshapiro:
@mpu-creare the SMAP notebook is working again.

This is not feeling as complete as a I wanted, but its ready to merge

mlshapiro:
@jmilloy @mpu-creare I started #148 for `__eq__` - its been removed from my branch

mpu-creare:
I did some checks on this branch. Looks good to me. Merging.

Issue #43, opened 2018-05-21
node and node.attribute references in pipeline definitions


jmilloy:
**Description**
We need to be able to define node attributes in pipeline definitions by referring to another node or another node's attribute (without clashing with attributes that are strings, dicts, etc)

**Describe the solution you'd like**
Reference the node or attribute as a string: `"my_attr": "node_name"` or `"my_attr": "node_name.attr1.attr2"`. These references need to be in a special location in the node definition so that the pipeline parser knows to interpret these as node and attribute references rather than as a string.

(We have also discussed dictionaries and tuples in the past, which remain a fine option.)

**Tasks**
 - [x] Define new spec.
 - [x] Update tests and docs to match new spec.
 - [x] Update the pipeline parser to use the spec (probably after #36)

PR #136, opened 2018-10-30
Feature/node sub attributes

### LINKED ISSUE #10: *Proposal*: Parse sub-attributes in Pipelines
We want to be able to reference
 * node sub-attributes specifically in Algorithm `inputs` and Compositor `sources`
 * sub-attributes generally in `attrs`

e.g.

```
{
    "nodes": {
        "downscaled_sm_algorithm": {
            ...
       },
       "reprojected": {
            "node": "core.data.type.ReprojectedSource"
            "inputs": {"source": "downscaled_sm_algorithm.twi"}
            "attrs": {"reprojected_coordinates": "downscaled_sm_algorithm.solmst.native_coordinates"}
       }
    }
}
```

This is just a proposal for now, and needs to be thought through more carefully and discussed.

jmilloy:
Closes #10 and #43 

## Usage

The tests in core/pipeline/test/test_pipeline_util.py provide a good summary of usage.

You can reference other nodes in the pipeline or attributes of other nodes in the pipeline in
 - Compositor `sources`
 - Algorithm 'inputs'
 - DataSource `lookup_source`
 - any `lookup_attrs`
 - output `node`

Normal values (strings, integers, etc) are used in
 - DataSource `source`
 - DataSource and Compositor `interpolation`
 - `attrs`

## ReprojectedSource

You can now make and use pipelines using `reprojected_coordinates`.

The ReprojectedSource `coordinates_source` can still be used:  `"lookup_attrs": {"coordinates_source": "other_node"}`

But you can also use `reprojected_source` and provide another node's attribute: * `"lookup_attrs": {"reprojected_coordinates": "other_node.some_coord_attr"}`

The ReprojectedSource `definition` property will output `"attrs": {"reprojected_coordinates": <serialized coordinates>`, and ReprojectedSource deserialize the coordinates. Note that this uses `attrs` because there is no lookup.

### DIFF
--- a/podpac/core/coordinates/array_coordinates1d.py
+++ b/podpac/core/coordinates/array_coordinates1d.py
@@ -90,7 +90,7 @@ class ArrayCoordinates1d(Coordinates1d):
         return cls(x.data, name=x.name)
 
     @classmethod
-    def from_json(cls, d):
+    def from_definition(cls, d):
         coords = d.pop('values')
         return cls(coords, **d)
 
@@ -193,7 +193,7 @@ class ArrayCoordinates1d(Coordinates1d):
         return area_bounds
 
     @property
-    def json(self):
+    def definition(self):
         d = OrderedDict()
         if self.dtype == float:
             d['values'] = self.coords.tolist()
--- a/podpac/core/coordinates/coordinates.py
+++ b/podpac/core/coordinates/coordinates.py
@@ -201,24 +201,26 @@ class Coordinates(tl.HasTraits):
         return cls(coords, coord_ref_sys=coord_ref_sys, ctype=ctype, distance_units=distance_units)
 
     @classmethod
-    def from_json(cls, d):
-        if isinstance(d, str):
-            d = json.loads(d)
-
+    def from_definition(cls, d):
         coords = []
         for elem in d:
             if isinstance(elem, list):
-                c = StackedCoordinates.from_json(elem)
+                c = StackedCoordinates.from_definition(elem)
             elif 'start' in elem and 'stop' in elem and 'step' in elem:
-                c = UniformCoordinates1d.from_json(elem)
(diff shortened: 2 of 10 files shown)


jmilloy:
@mpu-creare The two failing tests are waiting for the `__eq__` method in #132

jmilloy:
I forgot to mention that I used "lookup_source" and "lookup_attrs" instead of "node_source" and "node_attrs" because it just felt confusing to me alongside "node", for example:

```
{
    "nodes": {
        "mydata": {
            "node": "data.DataSource",
            "source": "my_data_string"
        },
        "mydata2": {
            "node": "data.DataSource",
            "node_source": "A.source"
        }
    }
}
```

But we can easily change that to whatever you would prefer before merging this in.

mpu-creare:
Yep, I noticed the 'lookup_' nomenclature. Works for me.

mpu-creare:
@jmilloy Do you agree about which attribute to tag above?

jmilloy:
As a user, I really would rather the output pipeline definition use coordinates_source whenever possible because it would be a lot easier to read and modify. The coordinates definitions can be pretty unwieldy.

But yeah, I could get behind removing coordinates_source altogether in fact. When writing pipeline nodes, coordinates_source is now superfluous because we can use sub-attrs. When caching, two identical ReprojectedSource nodes could have different hashes if one used reprojected_coordinates and one used coordinates_source, so always using reprojected_coordinates would be better in that regard, too.

mpu-creare:
Let's get rid of `coordinates_source` altogether. As a user you can now just do: 
```json
lookup_attr: "reprojected_coordinates": "some_node.native_coordinates"
```

versus

```json
lookup_source: "coordinates_source": "some_node"
```

jmilloy:
804937c

PR #137, opened 2018-10-30
Feature/coordinates group

### LINKED ISSUE #2: CoordinateGroup
@mpu I think you know this, but there is an initial implementation with very basic tests in the repository. Here's a summary.

Both Coordinate and CoordinateGroup inherit from BaseCoordinate, which mainly just defines a common API (currently stack, unstack, and intersect).

The CoordinateGroup wraps a list of Coordinate objects in a private attribute. Maybe you would prefer that this be a numpy array, not sure.

**Implemented**
 - `len(group)`
 - `for c in group`
 - indexing `group[1]`
 - slicing `group[1:4]`
 - multi-indexing `group[1, 'lat']` to get the underlying Coord objects
 - `groupA + groupB`
 - `groupA += groupB`
 - `groupA.append(coordinate)`
 - `group.dims` gives a set of unstacked dims, shared by all coordinates in the group but not necessarily in the same order or stacking.
 - `group.stack`, `group.unstack`, and `group.intersect` just map to each coordinate in the list and return a CoordinateGroup.

**Not implemented**
 - `groupA + coordinate` - I didn't know if it was better to fail or to append.  
 - `add_unique`
 - `iterchunks`
 - `latlon_bounds_str`

**Next Steps**
 - initial feedback
 - how to execute a node given a CoordinateGroup and what the output should be

jmilloy:
Closes #2 

These are meant to be usable as native_coordinates. They could also be used as the return value for find_coordinates. They should support whatever is most useful in that regard, for example so that you could call intersect on a Coordinates object or a GroupCoordinates object in the same way. So, while we can always loop through and map a function to the group's items, please feel free to update in the future with the most appropriate additional methods.


mpu-creare:
It seems like we couln't use `GroupCoordinates` as the return from find_coordinates because `find_coordinates` may have a mix-matching of dims, while every coordinate in `GroupCoordinates` need matching dims (even though they can be stacked. 

But it's an interesting idea nonetheless.

PR #139, opened 2018-10-31
Feature/cache refactor 0.2.0

drc-creare:
I think the cache refactor is ready to merge into develop. We can start using it for saving data to disk (and finding the bugs my tests don't cover).


podpac/core/cache/cache.py
```diff
@@ -0,0 +1,428 @@
...
+            If the data is not in the cache.
+        '''
+        mode = self._determine_mode(mode)
+        for c in self._cache_stores:
+            if c.cache_modes_matches(set([mode])):
+                if c.has(node=node, key=key, coordinates=coordinates):
+                    return c.get(node=node, key=key, coordinates=coordinates)
+        raise CacheException("Requested data is not in any cache stores.")
+
+    def rem(self, node=None, key=None, coordinates=None, mode=None):
```
mpu-creare: With the node refactor we went with `del`. Would you mind changing `rem` to `del` so everything matches?

podpac/core/cache/cache.py
```diff
@@ -0,0 +1,428 @@
...
+            The cached data.
+        
+        Raises
+        -------
+        CacheError
+            If the data is not in the cache.
+        '''
+        raise NotImplementedError
+
+    def rem(self, node=None, key=None, coordinates=None):
```
mpu-creare: Again `rem` --> `del` please.

podpac/core/cache/cache.py
drc-creare: I get an error when trying to define a del method for a class in ipython. I think because of the built in python del method:

```
n [1]: class C:
   ...:     def del():
  File "<ipython-input-1-74da07aeaab4>", line 2
    def del():
          ^
SyntaxError: invalid syntax
```

drc-creare:
I'd like to do the merge first and then hook into the Node stubs. I'll create an issue for that.  I can go either way on the documentation, but might not get to it until tomorrow. I'll add an issue.

podpac/core/cache/cache.py
mpu-creare: Ok, we'll have to fix that in the Node then.

PR #144, opened 2018-11-01
Feature/aws

dsully1992:


podpac/core/managers/aws_lambda.py
```diff
@@ -0,0 +1,150 @@
+"""
+Lambda is `Node` manager, which executes the given `Node` on an AWS Lambda
+function.
+"""
+import json
```
mpu-creare: You're missing the `from future imports` that make this python2 compatible.

podpac/core/managers/aws_lambda.py
```diff
@@ -0,0 +1,150 @@
...
+            coordinates.hash,
+            self.source_output.format)
+        waiter.wait(Bucket=self.s3_bucket_name, Key=filename)
+        # After waiting, load the pickle file like this:
+        resource = boto3.resource('s3')
+        with BytesIO() as data:
+            # Get the bucket and file name programmatically - see above...
+            resource.Bucket(self.s3_bucket_name).download_fileobj(filename, data)
+            data.seek(0)    # move back to the beginning after writing
+            self._output = cPickle.load(data)
```
mpu-creare: Technically we'll only set self._output if settings.DEBUG == True (in the future). 
This is OK for now.

podpac/core/managers/aws_lambda.py
dsully1992: Good point. It's worth noting our Handler is designed for a Python 3.6 environment, although this could run in Python2 (with a couple of changes). Will do.

podpac/core/managers/aws_lambda.py
dsully1992: Thanks for letting me know. Will Nodes just return the output when not setting `_output`?

podpac/core/managers/aws_lambda.py
dsully1992: As far as unit tests go, it's worth discussing a bit offline. It would be possible to test _if_ the S3 settings are configured, though that would still require mocking some of the handler functionality. Alternatively, we could just test one side of the communication. In other words, we could test that a pipeline can be created from the aggregated definition in `aws_lambda`. We'd get decent coverage in this simplest case, by splitting portions of `eval` into private, testable methods.

podpac/core/managers/aws_lambda.py
mpu-creare: you bet. They'd just return the output (after being cached somehow). Then on subsequent evaluates, the node might just hit the cache instead of recomputing.

podpac/core/managers/aws_lambda.py
mpu-creare: I'm kinda fine with the handler being python 3 only... it's not the user's python environment. You don't want to mess with the user's environment, but the environment on a server is far enough removed -- it's just a black box.

But we do want Python2.7 users to be able to use this feature

dsully1992: Agreed, I'll make this Python2.7 compatible today.

dsully1992: See [issue](#147), [commit](https://github.com/creare-com/podpac/commit/ce05484e7a6f6ee98d4d1978220d1d8fbbc4c896)

PR #150, opened 2018-11-02
HOTFIX: The ImageOutput image trait type is reverted to Bytes.

jmilloy:
The try-except that hid the actual error is also removed. I think that user code that needs to catch that exception should have it's own try-except. Is that okay?

### DIFF
--- a/podpac/core/pipeline/output.py
+++ b/podpac/core/pipeline/output.py
@@ -154,16 +154,12 @@ class ImageOutput(Output):
         Description
     """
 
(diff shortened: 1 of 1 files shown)


mpu-creare:
Sounds like we needed the tl.Union([BytesIO, Bytes]) or something like that to handle the flag change... and then some unit tests. No worries. 

Thanks for sorting this out for this release.

jmilloy:
I expect we can simply remove the `base64_encode` flag and force it to be one way or the other.

PR #161, opened 2018-11-13
Feature/pipeline traits usage

### LINKED ISSUE #141: Fix up traitlets usage in pipeline node
**Description**
There are a couple of issues in the pipeline node. 

* Generally you shouldn't set any attributes in traitlets validators
* When you create a pipeline with anything but the json the `hash` returns 0 (not correct)

jmilloy:
- Removes assignments in Pipeline validators
 - The Pipeline json attribute is defined when the Pipeline is initialized using a `path` or `definition` argument
 - Units tests are added to cover these cases
 - The AWS workaround is removed

Closes #141

### DIFF
--- a/podpac/core/pipeline/pipeline.py
+++ b/podpac/core/pipeline/pipeline.py
@@ -5,12 +5,12 @@ Pipeline Summary
 from __future__ import division, unicode_literals, print_function, absolute_import
 
 from collections import OrderedDict
-import json
+import json as json
 
 import traitlets as tl
 
 from podpac.core.node import Node
-
+from podpac.core.utils import OrderedDictTrait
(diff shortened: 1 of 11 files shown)


podpac/core/pipeline/pipeline.py
```diff
@@ -5,12 +5,12 @@
 from __future__ import division, unicode_literals, print_function, absolute_import
 
 from collections import OrderedDict
-import json
+import json as json
```
mpu-creare: ??? `import json` then rename as `json`?

jmilloy: Oops, I'll fix that tomorrow real quick.

Issue #120, opened 2018-10-15
Add segment length property to Coordinates1d


jmilloy:
**Description**
Non-monotonic coordinates cannot use segment ctype. In particular, area_bounds and intersection with outer=True do not work.

**Describe the solution you'd like**
Add a segment_length property, which can be used to implement area_bounds and outer intersections.

**Sample Code**

```
>>> c = ArrayCoordinates1d([10, 4, 6], ctype='center', segment_length=2)
>>> c.area_bounds
[3, 11]
```

jmilloy:
This issue was discussed in #109 and will replace the workaround in 7811d6c.

mpu-creare:
Incidentally, I suppose the segment_length should be allowed to be an array the same size as the coordinates. That way you can set the size of each segment in a non-monotonic system.

jmilloy:
Sure thing. What about UniformCoordinates? Should segment_length (value or array) be supported in the same way, or is the segment length always going to be the step?

mpu-creare:
Mmm.. that's a good point. Although, it would be an odd case to have different segment lengths for uniform coordinates.... but it could happen. 

I think I'd rather not try to support this functionality because I doubt anyone will ever use it. I think calculating segment_length (i.e. == step) in UniformCoordinates is fine. If we ever do have a use-case, I don't think the refactor would be horrendous. 

Waffling here... if it helps to make the interface more uniform, I also have no problem supporting this functionality in UniformCoordinates. So I leave it up to you what you want to do.

jmilloy:
@mpu-creare I started to look at this again this morning. I don't think that segment lengths are required for non-monotonic outer intersection. I think it is only the area_bounds that would use segment lengths, but area_bounds already uses extents if they were provided, so the segment lengths wouldn't be adding any functionality. I'm going to implement the non-monotonic outer intersection and submit a PR. I don't want to spend time on a general segment lengths feature if it is not yet needed, but I want to check with you because there is maybe something that I am missing.

mpu-creare:
* In the monotonic case, we actually know the bounds, and they should come from the neighboring elements:
``` python
In [19]: c = ArrayCoordinates1d([1, 2, 5, 7], extents=[0, 8])

In [20]: c[2].coordinates
Out[20]: array([5.])

In [21]: c[2].extents
Out[21]: array([3., 6.])
```
So this suggests maintaining the segment lengths

* Same goes for the strided selection:
```python
[...]

Issue #162, opened 2018-11-14
fix area bounds in non-monotonic coordinates


jmilloy:
**Description**
The area bounds of non-monotonic coordinates is incorrect when ctype is not 'point'.

**Steps to Reproduce**
```
c = ArrayCoordinates1d([20., 50., 60., 90., 40., 10.])
c.area_bounds
```

**Observed Behavior**
```
>>> c.area_bounds
array([ 5., -5.])
```

**Expected Behavior**
Either
 * the `area_bounds` should be inferred reasonably from the coordinates, or
 * an exception should be raised

**Additional Notes**
The area_bounds also works correctly when using ctype 'point`. For segment ctypes, the area_bounds works correctly when the `extents` are provided. It may not be possible to reasonably infer the area_bounds from non-monotonic coordinates, so the solution may be to require extents when defining  non-monotonic segment coordinates.

Alternatively, a reasonable inference for area_bounds would be to sort the coordinates and use the same value we would use for monotonic coordinates. In the example case, that would be `array([  5., 105.])`.

jmilloy:
See also #120

jmilloy:
@mpu-creare will you let me know if the inference or the exception is better? Both are quick to implement.

mpu-creare:
Please do the inference. 
[min - segment_length_of_min / 2, max + segment_length_of_max / 2] for not point
[min , max ] for point

If the user does not specify `segment_length` for ctype `not 'point'`, raise an exception with a clear message. 

Thanks!

jmilloy:
Okay, so this depends on #120.

jmilloy:
Implemented in #163

Issue #111, opened 2018-10-04
Read-only Node traits


jmilloy:
**Description**
Some node traits, especially the tagged Node `attrs`, should be read-only.

**Describe the solution you'd like**
We should look through the Node classes and set `readonly=True` to the appropriate traits. Note that these can still be set when necessary with `set_traits`.

jmilloy:
Note: I did this for coordinates in #163

jmilloy:
It will not be sufficient to add read_only=True to the traits; we'll have to explicitly call `set_trait` in the Node init methods.

jmilloy:
See #315

Issue #103, opened 2018-09-17
Handle Coordinates Properties consistently


jmilloy:
**Description**
Coordinate properties include ctype, coordinate reference system, and units. These can currently be specified during manual `Coordinates1d` creation and are generally passed when copying coordinates and checked when appropriate. We also need to be able easily set defaults during `StackedCoordinates` and `Coordinates` creation that can still be overridden by the `Coordinates1d` if necessary, and add checks where appropriate in methods for both of these classes.

There is a placeholder section in the User Guide for this, and there are not yet tests for this, either.

PR #163, opened 2018-11-14
Improve non-monotonic array coordinates

jmilloy:
This PR includes
 - #120 segment length property
 - #162 non-monotonic area bounds
 - non-monotonic outer select
 - #168 default ctype for singleton coordinates
 - #103 proper handling of coordinate properties
 - read-only traits in Coordinates1d, ArrayCoordinates1d, UniformCoordinates1d, StackedCoordinates (see also #111)
 - enforces crs to be the same in all dimensions for StackedCoordinates and Coordinates.
 - `divide_delta` utility function (similar to `add_coord`)
 - #211 partially implemented custom JSONEncoder, used by Coordinates


mpu-creare:
Do you want to include #162 in this PR?

jmilloy:
I've broadened this PR, and marked it as WIP

jmilloy:
@mpu-creare This feature is implemented, but SMAP should be updated. It may fail currently, or it may have workarounds that are no longer needed. Do you want me to take a look at that or can you do it?

jmilloy:
@mlshapiro I've updated the coordinates for this feature and there is one compositor test failing: `test_compositor_implemented_functions`. Can you take a look at that? You may be able to fix it, or we just need to comment out that assertion for now.

mpu-creare:
@jmilloy if you have time, please have a look... I'm swamped...

mpu-creare:
@jmilloy I just confirmed that this still works with SMAP. I'm merging, and I'll make a new issue to update SMAP to use this functionality.

mpu-creare:
Of course I'll figure out the failing unit test first...

jmilloy:
So we should be able to close #120, #162, and #168.

(But I want to spend an hour or so on #211, for the next release)

PR #170, opened 2018-11-16
FEATURE: add initial logging capability

mlshapiro:
This is basic, but it should get us started.

```python
from podpac import log

log.info('test')   # goes to console and file podpac.log
```

```python
from podpac import logger
from popac import log

logger.set_level(20, handler='file')
logger.set_level(40, handler='console')

log.info('test)  # goes to file, not console

logger.set_filename('mylogfile')
logger.disable('file')   # only log to console
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,6 +11,8 @@ __pycache__
 *.egg-info
 podpac/cache/
 
+# log
+podpac.log
 
 # test
 .cache
--- a/podpac/__init__.py
+++ b/podpac/__init__.py
@@ -16,6 +16,7 @@ from podpac.core.coordinates import Coordinates, crange, clinspace
 from podpac.core.node import Node, NodeException
 import podpac.core.authentication as authentication
 from podpac import settings
+from podpac.logger import log
 
 # Organized submodules
 # These files are simply wrappers to create a curated namespace of podpac modules
--- a/podpac/logger.py
+++ b/podpac/logger.py
@@ -1,0 +1,116 @@
+"""
+PODPAC Logging
+
+Logging settings are configured in :attr:`podpac.settings.LOG`
+See https://docs.python.org/3/library/logging.config.html#logging.config.dictConfig
+and https://docs.python.org/3/howto/logging-cookbook.html#an-example-dictionary-based-configuration
+"""
+
+import logging
+import logging.config
+
+from podpac import settings
+
+
+CONFIG = {
+    'version': 1,
+    'formatters': {
+        'simple': {
+            'format': '[%(levelname)s] %(message)s'
+        },
+        'verbose': {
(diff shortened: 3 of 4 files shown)


jmilloy:
Libraries should should not initialize any handlers. User code should do that.

We should also get the logger by name, best practice is to use `__name__` but `'podpac'` or `'creare.podpac'` might be acceptable as well.

mpu-creare:
Marc and I talked. It sounds like  you know a lot more about this. When you have a moment in the future (no rush) please fix up this branch how you want it and give us a few examples of how to use it.

jmilloy:
[Configuring Logging for a Library](https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library) from the Python docs:

> It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers. This is because the configuration of handlers is the prerogative of the application developer who uses your library. The application developer knows their target audience and what handlers are most appropriate for their application: if you add handlers ‘under the hood’, you might well interfere with their ability to carry out unit tests and deliver logs which suit their requirements. 

[Logging in a Library](https://docs.python-guide.org/writing/logging/#logging-in-a-library) from Kenneth Reitz's Python guide.

>Because the user, not the library, should dictate what happens when a logging event occurs, one admonition bears repeating: *It is strongly advised that you do not add any handlers other than NullHandler to your library’s loggers.*

> Best practice when instantiating loggers in a library is to only create them using the __name__ global variable: the logging module creates a hierarchy of loggers using dot notation, so using __name__ ensures no name collisions.

Novice users can use `logging.basicConfig`. [Here's another short guide](http://eric.themoritzfamily.com/learning-python-logging.html) that explains how and why, with some examples using a library that has logging. I found it [here](https://stackoverflow.com/questions/25629985/python-logging-best-practice-for-reusable-modules-intended-to-be-included-with-o) on SO.

> How your loggers are configured is an application level setting. Your modules have no reason to worry about how to configure loggers.

> logging.basicConfig() is a really handy function. I find myself using basicConfig() often because most of the time I simply want to output to stderr or to a single file rather than wanting complex routing using Handlers.

We can use basicConfig to log to 'podpac.log' at the top of all of our Jupyter notebooks. Matt, for development outside of a notebook, you can start the interpreter with file logging on in one line:

```
$ python -i -c "import logging; logging.basicConfig(filename='podpac.log')"
```

I don't think a setting is any easier than basicConfig.

There are a few other useful best practices with explanations [here](https://fangpenlin.com/posts/2012/08/26/good-logging-practice-in-python/). In addition to what's covered elsewhere, we should be careful about getting the podpac logger when using throughout podpac:

> Do not get logger at the module level unless disable_existing_loggers is False

mlshapiro:
closing in favor for #174

PR #174, opened 2018-11-21
FEATURE: second attempt at the log - see contributing.md "## Logging"

mlshapiro:
This includes one util method for creating podpac only log handler - i'm open to removing this, but this seems like a good middle road.

BasicConfig will work for most people, but unfortuantely there is no way to configure this to be a podpac only logger (i.e. you get all other module logging as well). This util method will create a log file with a specific format that has podpac ONLY logs.


mlshapiro:
merged

PR #175, opened 2018-11-27
WIP: Hook up Caching 

mpu-creare:
Add the caching functionality from @drc-creare  into the nodes.


podpac/core/node.py
```diff
@@ -748,3 +751,47 @@ def clear_disk_cache(self, attr='*', node_cache=False, all_cache=False):
...
+    """
+
+    cache_key = 'output'
+
+    @functools.wraps(fn)
+    def wrapper(self, coordinates, output=None):
+        if self.debug:
+            self._requested_coordinates = coordinates
+
+        cache_coordinates = coordinates.transform(sorted(coordinates.dims)) # order agnostic caching
```
mpu-creare: This should probably go into the caching framework directly? i.e. has_cache, put_cache, and get_cache, would work with any order, doing the transpose for you each time.

mpu-creare:
Ah yes, that would make sense to have that `cache_key` optionally set. Optional arguments in decorators is a little complicated. https://realpython.com/primer-on-python-decorators/#both-please-but-never-mind-the-bread I'll leave that till later.

mpu-creare:
Hey @drc-creare . This PR is for hooking up the caching functionality. There are also a few other improvements in here that I made during SUSMAP. I'd like to merge this into develop tomorrow so that I can make the 0.2.1 release for the SUSMAP folks and upload the windows install to S3. 

Could you please review? 

I still have to test all of the notebooks, but I can do that on develop.

podpac/__init__.py
```diff
@@ -11,6 +11,27 @@
     Description
 """
 
+
+# Monkey match os.makedirs for Python 2 compatibility
+import sys
+import os
+_osmakedirs = os.makedirs
+def makedirs(name, mode=511, exist_ok=False):
```
mpu-creare: I saw @mlshapiro use the `exist_ok` param and really liked it. Turns out it was Python 3 only, hence the monkey patch.

podpac/core/cache/cache.py
```diff
@@ -7,6 +7,7 @@
 import os
 from glob import glob
 import shutil
+from hashlib import md5 as hash_alg
```
mpu-creare: @drc-creare @dsully1992 Note that Python 3's default `hash` function uses a random seed. That probably why you didn't get consistent hashes on the Lambda functions, David. Dave, this change was required to actually make the disk cache useable across Python sessions. A little tricky to test in a unittest, so I didn't do it.

podpac/core/cache/cache.py
```diff
@@ -140,19 +141,20 @@ def has(self, node, key, coordinates=None, mode=None):
...
-        return self.get_hash_val(hashable_repr)
+        hashable_repr = 'None' if coordinates is None else coordinates.hash
+        return hashable_repr 
 
     def hash_key(self, key):
-        hashable_repr = str(repr(key))
-        return self.get_hash_val(hashable_repr)
+        #hashable_repr = str(repr(key)).encode('utf-8')
+        #return self.get_hash_val(hashable_repr)
+        return key
```
mpu-creare: I didn't hash the key-- felt it was useful to be human-readable string on the filesystem. It also helped me debug something...

podpac/core/cache/cache.py
```diff
@@ -140,19 +141,20 @@ def has(self, node, key, coordinates=None, mode=None):
 class CacheStore(object):
 
     def get_hash_val(self, obj):
-        return hash(obj)
+        return hash_alg(obj).hexdigest()
```
mpu-creare: Note, Hashlib cannot hash `None` or unencoded Unicode strings.

podpac/core/node.py
```diff
@@ -95,17 +104,39 @@ class Node(tl.HasTraits):
 
     units = Units(default_value=None, allow_none=True)
     dtype = tl.Any(default_value=float)
+    cache_output = tl.Bool(True)
+    cache_update = tl.Bool(False)
```
mpu-creare: I need this flag for when I cache a completely nan tile, for when I'm offline and request something (which I can't get from SMAP). So, when I'm back online, I want to update the cache instead of retrieving nans from it.

podpac/core/node.py
```diff
@@ -95,17 +104,39 @@ class Node(tl.HasTraits):
 
     units = Units(default_value=None, allow_none=True)
     dtype = tl.Any(default_value=float)
+    cache_output = tl.Bool(True)
```
mpu-creare: Presumably you don't always want to cache the output for all nodes. YOu can choose not to decorate the `eval` function for classes of `Nodes` but this gives you instance-level control.

podpac/core/node.py
```diff
@@ -95,17 +104,39 @@ class Node(tl.HasTraits):
...
+    cache_update = tl.Bool(False)
     cache_type = tl.Enum([None, 'disk', 'ram'], allow_none=True)
+    cache_ctrl = tl.Instance(cache.CacheCtrl, allow_none=True)
+
+    @tl.default('cache_ctrl')
+    def _cache_ctrl_default(self):
```
mpu-creare: The default should really probably be a ram store. I'm not sure what to do about SMAP for example...

podpac/core/node.py
```diff
@@ -74,8 +77,14 @@ class Node(tl.HasTraits):
 
     Attributes
     ----------
+    cache_output: bool
+        Default if True. Should the node's output be cached? 
+    cache_update: bool
+        Default is False. Should the node's cached output be updated from the source data? 
     cache_type : [None, 'disk', 'ram']
```
mpu-creare: We'll have to think a little more about this attribute... what if I want ram and disk?

podpac/core/node.py
```diff
@@ -426,9 +457,9 @@ def get_cache(self, key, coordinates=None):
         if not self.has_cache(key, coordinates=coordinates):
             raise NodeException("cached data not found for key '%s' and cooordinates %s" % (key, coordinates))
 
-        # return cache.get(self, data, key, coordinates=coordinates)
+        return self.cache_ctrl.get(self, key, coordinates=coordinates)
 
-    def put_cache(self, data, key, coordinates=None, overwrite=False):
+    def put_cache(self, data, key, coordinates=None, overwrite=False, raise_no_cache_exception=True):
```
mpu-creare: I really went back and forth on this one. I need the `raise_no_cache_exception` for when I REALLY want to make sure I cache something (Then I can set it to true). but I don't want my nodes to fail whenever I don't have a cache defined. The decorator will still try to put... perhaps I should handle this elsewhere? But I think it suffices for now.

podpac/core/node.py
```diff
@@ -485,266 +525,160 @@ def del_cache(self, key=None, coordinates=None):
...
+        if self.debug:
+            self._output = data
 
-    def clear_disk_cache(self, attr='*', node_cache=False, all_cache=False):
-        """Helper function to clear disk cache.
+        return data
+    return wrapper
 
-        WARNING: This function will permanently delete cached values
+def cached_property(key, depends=None, raise_no_cache_exception=False):
```
mpu-creare: I figured out how to do the dependencies on this decorator by setting up observers on traits. It complex, and the final implementation feels a little hacky, but it works. 
Turns out if you tag an attr, it will automatically become a `depends` for any cached property. This is because changing this attr changes the definition of the node, which gets used in the cache key. I think this is appropriate behaviour, but took me a while to figure out what was going on.

podpac/core/test/test_node.py
```diff
@@ -215,208 +215,283 @@ def test_create_output_array_dtype(self):
...
+
+        self.node.rem_cache()
+        assert not self.node.has_cache('a')
+        assert not self.node.has_cache('b')
+        assert not self.node.has_cache('a', coordinates=self.coords)
+        assert not self.node.has_cache('c', coordinates=self.coords)
+        assert not self.node.has_cache('c', coordinates=self.coords2)
+        assert not self.node.has_cache('d', coordinates=self.coords)
+
+    @pytest.mark.skip('BUG: Need to fix this.')
```
mpu-creare: Can we try to fix this for 0.2.2?

podpac/core/test/test_node.py
```diff
@@ -215,208 +215,283 @@ def test_create_output_array_dtype(self):
...
+        self.node.rem_cache(key='a')
+
+        assert not self.node.has_cache('a')
+        assert not self.node.has_cache('a', coordinates=self.coords)
+        assert self.node.has_cache('b')
+        assert self.node.has_cache('c', coordinates=self.coords)
+        assert self.node.has_cache('c', coordinates=self.coords2)
+        assert self.node.has_cache('d', coordinates=self.coords)
+
+    @pytest.mark.skip('BUG: Need to fix this.')
```
mpu-creare: Same with this test, 0.2.2?

podpac/core/test/test_node.py
```diff
@@ -215,208 +215,283 @@ def test_create_output_array_dtype(self):
...
+        assert t.a2() == 4
+        t.b = 2
+        assert t.b2() == 4  # This happens because the node definition changed
+        t.rem_cache()
+        assert t.c2() == 2  # This forces the cache to update based on the new node definition
+        assert t.d2() == 2  # This forces the cache to update based on the new node definition
+        t.c = 2
+        assert t.c2() == 4  # This happens because of depends
+        t.d = 2
+        assert t.d2() == 2  # No depends, and doesn't have a tag
```
mpu-creare: This test actually really highlights a user error... but this is the specified behaviour. Somewhat similar to the @tl.default behaviour -- only called once, not updated.

podpac/datalib/smap.py
```diff
@@ -190,29 +196,39 @@ def _infer_SMAP_product_version(product, base_url, auth_session):
...
-                with open(SMAP_BASE_URL_FILE, 'w') as fid:
-                    fid.write(r)
-            except Exception as e:
-                warnings.warn("Could not overwrite SMAP url update on disk:" + str(e))
-except Exception as e:
-    warnings.warn("Could not retrieve SMAP url from PODPAC S3 Server. Using default." + str(e))
-
-SMAP_BASE_URL_REGEX = re.compile(re.sub(r'\d', r'\\d', SMAP_BASE_URL.split('/')[2]))
+_SMAP_BASE_URL = None
+def SMAP_BASE_URL():
```
mpu-creare: I would have loved to get rid of the `()`, but I don't know the syntax for module-level properties. Anyway, lazy-loading makes importing podpac faster.

drc-creare:
> 
> 
> Hey @drc-creare . This PR is for hooking up the caching functionality. There are also a few other improvements in here that I made during SUSMAP. I'd like to merge this into develop tomorrow so that I can make the 0.2.1 release for the SUSMAP folks and upload the windows install to S3.
> 
> Could you please review?
> 
> I still have to test all of the notebooks, but I can do that on develop.

@mpu Looking at it now.

podpac/core/cache/cache.py
drc-creare: ok, thats fine, but I'll have to override `hash_key()` in `DiskCacheStore` so that the key is cleansed (e.g. remove file path separators, periods, and the special strings 'kKeY' etc.).

basically run this function on it:

https://github.com/creare-com/podpac/blob/19554a37c20e31a669881ad53d1d67ed43427e40/podpac/core/cache/cache.py#L344

podpac/core/node.py
drc-creare: Agreed, `CacheCtrl` should probably be able to accept a collection of cache mode strings (list/set), then this could be a collection too.

podpac/core/node.py
drc-creare: When we add RAM cache we'll probably need to make the default cache_ctrl a global variable (or at least the RAM store) so that shared data structures can be used.

podpac/core/node.py
```diff
@@ -485,266 +525,160 @@ def del_cache(self, key=None, coordinates=None):
...
-        elif node_cache:
-            shutil.rmtree(self.cache_dir)
-        else:
-            for f in glob.glob(self.cache_path(attr)):
-                os.remove(f)
+    Examples
+    ----------
+    >>> class MyClass(Node):
+           value = 0
+           @cached_func('native_coordinates')
```
drc-creare: Example should be consistent with the name of the decorator (e.g. `@cached_property('native_coordinates')`). 

But maybe it should be  `@cached_func` so as not to be confused with python `@property` which allows for the sytanx of accessing a function like it was a attribute (e.g. without the parenthesis.

podpac/core/node.py
```diff
@@ -485,266 +525,160 @@ def del_cache(self, key=None, coordinates=None):
...
+           def square_value(self):
+               self.value += 1
+               return self.value
+           @cached_func('native_coordinates', depends='value')
```
drc-creare: same as https://github.com/creare-com/podpac/blob/19554a37c20e31a669881ad53d1d67ed43427e40/podpac/core/node.py#L619

podpac/core/node.py
drc-creare: Cool!! I didn't think you would get to doing this decorator.

podpac/core/test/test_node.py
drc-creare: I'll take a look at this test tomorrow.

podpac/core/test/test_node.py
drc-creare: and this one

podpac/core/node.py
mpu-creare: I meant to flag the fact that I wasn't totally happy with the name. I think `cache_func` is better. I will rename.

podpac/core/cache/cache.py
mpu-creare: Ah! Really need that function elsewhere... I forgot to flag it...

podpac/core/cache/cache.py
```diff
@@ -314,7 +316,8 @@ def cache_dir(self, node):
         basedir = self._root_dir_path
         subdir = str(node.__class__)[8:-2].split('.')
         dirs = [basedir] + subdir
-        return os.path.join(*dirs)
+        # TODO: Use a function to sanitize directory name 
+        return os.path.join(*dirs).replace('<', '').replace('>', '')
```
mpu-creare: Need to call  `cleanse_filename_str(self, s): ` here.

PR #176, opened 2018-11-27
Sets attribute __doc__ strings directly so that they will be inherited

jmilloy:
A wrote a class decorator to set the `__doc__` strings for `class_traits`, but that didn't work for a few reasons. What do you think of this? It is the basically the same as writing  `#:` or `""" """` docstrings that Sphinx picks up (and Sphinx displays them exactly the same), but they are inherited properly.

### DIFF
--- a/doc/source/developer/contributing.md
+++ b/doc/source/developer/contributing.md
@@ -50,9 +50,9 @@ Examples:
 - https://docs.scipy.org/doc/numpy/docs/howto_document.html#example-source
 
 Note that class attributes can be documented multiple ways.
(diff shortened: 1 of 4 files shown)


jmilloy:
Just guessing here. When building docs, Sphinx parses and uses the `#:` or `""" """` comment as the docstring if it exists and otherwise falls back on the attribute `__doc__` string. Traitlets has default `__doc__` strings for traits (and this is similar but distinct from the `info_text`). Of course, those `__doc__` strings are inherited. However, what Sphinx *doesn't* do is set the `__doc__` string from the docstring comment, so in child classes it can only fall back on the traitlets default.

PR #190, opened 2018-12-09
WIP: GFS

jmilloy:
Adds support for GFS from S3.
 * `GFSSource` - time-independent data source for a single forecast/fill, e.g. "SOIM/0-10 m DPTH/20181206/1200/003")
 * `GFS` - time-depedendent data source for all forecasts from a given date, e.g aggregates all files prefixed "SOIM/0-10 m DPTH/20181206/1200"
 * `GFSLastest` - the latest GFS node (based on local datetime)

### DIFF
--- a/podpac/core/compositor.py
+++ b/podpac/core/compositor.py
@@ -37,7 +37,7 @@ class Compositor(Node):
         Default is False. The source_coordinates do not have to completely describe the source. For example, the source
         coordinates could include the year-month-day of the source, but the actual source also has hour-minute-second
         information. In that case, source_coordinates is incomplete. This flag is used to automatically construct
-        native_coordinates
+        native_coordinates.
     n_threads : int
         Default is 10 -- used when threaded is True.
         NASA data servers seem to have a hard limit of 10 simultaneous requests, which determined the default value.
@@ -47,7 +47,7 @@ class Compositor(Node):
         The source is used for a unique name to cache composited products.
     source_coordinates : :class:`podpac.Coordinates`
         Description
-    sources : np.ndarray
+    sources : :class:`np.ndarray`
         An array of sources. This is a numpy array as opposed to a list so that boolean indexing may be used to
         subselect the nodes that will be evaluated.
     threaded : bool, optional
@@ -75,7 +75,7 @@ class Compositor(Node):
     sources = tl.Instance(np.ndarray)
     cache_native_coordinates = tl.Bool(True)
     
-    interpolation = interpolation_trait()
+    interpolation = interpolation_trait(default_value=None)
 
     threaded = tl.Bool(False)
     n_threads = tl.Int(10)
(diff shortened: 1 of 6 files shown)


mlshapiro:
@jxm i found i needed a new type of compositor (i.e. a new `composite` method for my compositor node - i am close to being finished and will commit tomorrow. This compositor will just stitch together my raster files.

Otherwise, i made a few updates to the Rasterio node that may help you too - for instance the rasterio node was reading in data upside down and without the `window` keyword arg input correctly. I was also getting back `nan` values, but the issue was in the way that the compositor is selecting sources. I added a new method to the compositor `select_sources` that I overwrite in my compositor so that I can use the requested coordinates to select the individual datasource nodes the use for evaluation.

Try creating a new branch and merging in my changes to see if any of that helps, but I imagine that the issue is in the compositor source selection (`select_sources`) or the `composite` method.

jmilloy:
@mlshapiro Okay, cool. My data was also upside down, so I think your rasterio updates will be helpful.

I'm not using a Compositor at all. I ended up compositing individual GFSSources manually I guess in a DataSource... it's just 4 lines. But I'll take a look at what you did and see if it would work for me instead.

jmilloy:
@mlshapiro No, even with your more recent commits to feature/terrain_tiles, my grib files are loaded upside down, the nodata values aren't handled correctly, and the interpolation doesn't work properly. I'm going to give xarray (through iris) and pygrib a try just to see what the data is like.

Btw, I did in-memory rasterio loading using a rasterio MemoryFile like this:

```
    @tl.default('dataset')
    def open_dataset(self):
        """Opens the data source"""

        cache_key = 'fileobj'
        with rasterio.MemoryFile() as f:
            if self.cache_ctrl and self.has_cache(key=cache_key):
                data = self.get_cache(key=cache_key)
                f.write(data)
            else:
                self._logger.info('Downloading S3 fileobj (Bucket: %s, Key: %s)' % (BUCKET, self._key))
                s3.Object(BUCKET, self._key).download_fileobj(f)
                f.seek(0)
                self.cache_ctrl and self.put_cache(f.read(), key=cache_key)
            
            dataset = f.open()

        return dataset
```

mlshapiro:
interesting. i'm pulling your branch now to take a look. I made some other changes to Rasterio DataSource last night, my data (and other test data) is being read the correct way now.

Thank you for the memory reading (and caching syntax) - i will implement that into TerrainTiles and then we can figure out how to abstract that for any S3 DataSource handling.

jmilloy:
@mlshapiro Okay, my mistake! The lat coordinates are fixed for GFS, too. I was getting cached data.

I think it might be more efficient to flip the native_coordinates than the data. All you have to do is switch `bottom` and `top` in get_native_coordinates, right? What do you think about that?

mlshapiro:
@jmilloy Good idea - I think that is more intuitive to use the native_coordinates rather than flipping the raster. Rasterio `get_data` uses the coordinates_index, so as long as the coordinates `intersect` method will send back the index with the slices in the right direction then it should all work out great.

jmilloy:
@mlshapiro I pulled your branch into feature/gfs and just pushed updates (removing my lat flipping). If you want to take a look and see if you have an idea what's going on with the interpolation, that would be cool. I'm going to put together a quick notebook now that just does native_coordinates evaluation at least.

mlshapiro:
@jmilloy sounds good - send me the notebook when you are done and I will test that rather than make my own.

jmilloy:
@mlshapiro Can you give me push access to podpac-examples? I tried to push a branch with my demo notebook but was denied.

Here's the notebook, in the meantime.

[gfs.ipynb.zip](https://github.com/creare-com/podpac/files/2664137/gfs.ipynb.zip)

mlshapiro:
done - i'll check it out now thanks

jmilloy:
https://github.com/creare-com/podpac-examples/blob/f86786d2e70b8c50e4f15e3fac4cd94fd8d21569/notebooks/demos/gfs.ipynb

jmilloy:
Rather, this version will remain up to date. This should be it for a bit though.

https://github.com/creare-com/podpac-examples/blob/feature/gfs/notebooks/demos/gfs.ipynb

@mpu-creare You can demo the `GFSSource` portion.

mlshapiro:
> 
> 
> @mlshapiro Okay, my mistake! The lat coordinates are fixed for GFS, too. I was getting cached data.
> 
> I think it might be more efficient to flip the native_coordinates than the data. All you have to do is switch `bottom` and `top` in get_native_coordinates, right? What do you think about that?

fixed in 21101025a2ec311ff732fce34a28d9d519477e98

PR #196, opened 2018-12-11
WIP: Features/cosmos stations

mpu-creare:
WIP for hooking up COSMOS station datasource.


podpac/core/data/interpolate.py
```diff
@@ -334,6 +335,8 @@ def can_interpolate(self, udims, source_coordinates, eval_coordinates):
 
         # confirm that udims are in both source and eval coordinates
         # TODO: handle stacked coordinates
+        # subselect unstacked coordinates for now
+        udims_subset = [d for d in udims_subset if d in source_coordinates.dims]
```
mpu-creare: This line allows multiple interpolators to be selected as part of the queue. Without it, only one interpolator that can handle all the dims will be selected.

podpac/core/data/interpolate.py
```diff
@@ -1093,15 +1096,19 @@ def _select_interpolator_queue(self, source_coordinates, eval_coordinates, selec
...
                 can_handle = getattr(interpolator, select_method)(udims, source_coordinates, eval_coordinates)
 
-                # if interpolator can handle all udims
-                if not set(udims) - set(can_handle):
-
+                # dims handled by this interpolator
+                this_handled_dims = set(can_handle) - set(handled_dims)
+                
+                # if interpolator can handle some udims
+                if this_handled_dims:
```
mpu-creare: Again, this block of code allows multiple interpolators to be selected as part of the queue. 

I just realized that the original code allowed a single interpolator to handle all the dimensions if possible. A little bit of logic should allow that again. 

I.e. A can handle time, B can handle lat-lon, C can handle time lat-lon and alt, so just use C for all. (instead of A for time, B for lat-lon, C for alt based on the current implementation.)

podpac/core/data/interpolate.py
```diff
@@ -1212,16 +1219,45 @@ def interpolate(self, source_coordinates, source_data, eval_coordinates, output_
         # for debugging purposes, save the last defined interpolator queue
         self._last_interpolator_queue = interpolator_queue
 
+
```
mpu-creare: The changes in the following two blocks enable incremental interpolation from multiple interpolators in the queue. The tricky bit is making intermediate datastructures and coordinates to store the intermediate results.

mpu-creare:
@mlshapiro I was hoping COSMOS would be easy to implement, but I don't think this will be ready for AGU because I had to touch the interpolator pretty heavily. I think I probably made changes that are counter your original design. It's also breaking a bunch of unit tests that I'm not going to dig into in this short time period. But there are a few things here we should discuss. 

As it stands, COSMOS "works". It will give you data, but the example in the `__main__` section overwrites data from all the stations with a single station's data. I thought originally it was a spatial_tolerance problem... which still might be the case, but I'm done being stubborn. I think my time is better spent working on messing with the terrain and GFS data. This has gone the way of the nextrad data...

mlshapiro:
@mpu-creare i will take a look at this in the afternoon - each of dataset lends new insight :)

podpac/core/data/interpolate.py
```diff
@@ -1212,16 +1219,45 @@ def interpolate(self, source_coordinates, source_data, eval_coordinates, output_
...
             # run interpolation
-            output_data = interpolator.interpolate(udims,
-                                                   source_coordinates,
-                                                   source_data,
-                                                   eval_coordinates,
-                                                   output_data)
-
-        return output_data
-
+            intermediate_output = interpolator.interpolate(udims,
```
mlshapiro: I had expected the interpolators to use `output_data`  as the `intermediate_output`, similar to how you are using it here. 

Unfortunately the interpolator handling is still pretty fragile (re `loop_helper`...) I think this is part of the iteration for the interpolator where it has first class support for composited data sources in addition to atomic data sources.

mlshapiro:
Looking at this for a moment I think it makes sense to wait until we can discuss these changes a little more.  You should have enough with TerrainTiles and GFS (and of course SMAP)

PR #198, opened 2018-12-11
Fix datasource eval when output is provided

jmilloy:
I think this is a pretty simple/solid PR, but I just wanted to run it by you. I thought I would need to transpose the coordinates *earlier* so that `self._requested_source_coordinates` would also be transposed, but tests seem to be working so maybe this is just fine.


mlshapiro:
looks good to me - merging now

PR #199, opened 2018-12-11
ArrayTrait is added; enables deserialization/serializationn of numpy array attrs.

jmilloy:
The terrain_tiles pattern matching notebook works with this branch.


mpu-creare:
looks good to me. Thanks!

PR #200, opened 2018-12-12
Feature/aws

dsully1992:
This includes updates to our AWS Lambda feature, which accommodate other Podpac changes since the creation of this feature.

The biggest changes here are:

- Install all Podpac dependencies to allow any `Node` type to be evaluated
- Integrate with changes to Podpac such as how `Pipeline` objects are defined, and `PodpacSettings` as a `dict`


setup.py
```diff
@@ -1,22 +1,24 @@
...
+from setuptools import find_packages, setup
+
+import version
+
+""" podpac module"""
+
 
 # get version
 sys.path.insert(0, 'podpac')
-import version
```
mpu-creare: `import version` has to go below this line.

mlshapiro: Just pushed this updated, thanks!

PR #206, opened 2019-01-28
Feature/import faster

mpu-creare:
Faster imports. Testing on my machine: 
2.2 seconds before lazy import
1.6 seconds after lazy import. 

Probably still a few places where it could be improved, but this seems to work pretty well for now. 

@jmilloy @mlshapiro @dsully1992 @drc-creare  
Please use the `optional_import` utility function for optional imports -- as it uses the lazy_import module.

### DIFF
--- a/podpac/core/utils.py
+++ b/podpac/core/utils.py
@@ -11,7 +11,6 @@ import functools
 import importlib
 from collections import OrderedDict
 import logging
-import lazy_import
 
 import traitlets as tl
 import numpy as np
--- a/podpac/core/utils.py
+++ b/podpac/core/utils.py
@@ -11,6 +11,7 @@ import functools
 import importlib
 from collections import OrderedDict
 import logging
+import lazy_import
 
 import traitlets as tl
 import numpy as np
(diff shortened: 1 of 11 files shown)


podpac/core/utils.py
```diff
@@ -133,22 +134,14 @@ def optional_import(module_name, package=None, module_attr=None, return_root=Fal
...
-        except ImportError:
-            module = None
+    if return_root:
+        module = lazy_import.lazy_module(module_name, level='base')
+        if module_attr:
+            raise Exception("Cannot defined 'module_attr' if 'return_root == True'")
+    else:
+        module = lazy_import.lazy_module(module_name)
+        if module_attr:
+            module = getattr(module, module_attr)
```
mpu-creare: I think this line will actually hard-load the module. Perhaps this requires a little more logic... or perhaps we should scrap the optional_import module completely and just use lazy_import.

mpu-creare: Let's scrap optional_import and just use lazy_import for all our optional imports.

Issue #100, opened 2018-08-27
Improve error messages


mpu-creare:
**Description**
We will need to have error messages percolate up to the user when using pipelines -- both on the web side and the Python side. 

**Describe the solution you'd like**
Full tracebacks and sane error messages are communicated to the user

This may also be related to logging #99

PR #208, opened 2019-01-29
ENH: Adding logging to SMAP. Also adding automatic file logging feature. 

mpu-creare:
This is towards #100 .

Adding logging to SMAP (remove warnings.warn and print statements). 

Also added feature to automatically log to file -- as this is something I want always.

### DIFF
--- a/podpac/core/settings.py
+++ b/podpac/core/settings.py
@@ -28,7 +28,9 @@ DEFAULT_SETTINGS = {
     'S3_BUCKET_NAME': None,
     'S3_JSON_FOLDER': None,
     'S3_OUTPUT_FOLDER': None,
-    'AUTOSAVE_SETTINGS': False
+    'AUTOSAVE_SETTINGS': False,
+    'LOG_TO_FILE': False,
+    'LOG_FILE_PATH': os.path.join(os.path.expanduser('~'), '.podpac', 'logs', 'podpac.log')
 }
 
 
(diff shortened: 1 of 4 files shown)


podpac/core/settings.py
```diff
@@ -28,7 +28,9 @@
     'S3_BUCKET_NAME': None,
     'S3_JSON_FOLDER': None,
     'S3_OUTPUT_FOLDER': None,
-    'AUTOSAVE_SETTINGS': False
+    'AUTOSAVE_SETTINGS': False,
+    'LOG_TO_FILE': False,
+    'LOG_FILE_PATH': os.path.join(os.path.expanduser('~'), '.podpac', 'logs', 'podpac.log')
```
mpu-creare: @mlshapiro  What do you think about the above two settings...

podpac/core/settings.py
```diff
@@ -270,8 +272,9 @@ def load(self, path=None, filename='settings.json'):
...
 
-
-
-
 # load settings dict when module is loaded
 settings = PodpacSettings()
+
+if settings['LOG_TO_FILE']:
+    from podpac.core.utils import create_logfile
+    create_logfile()
```
mpu-creare: @mlshapiro This is something that will be very useful to me. I'd forget to set this up every time, but I'd like to have accurate logs of when opendap is down.

Issue #213, opened 2019-02-12
Add resource limitations to current cache stores


drc-creare:
**Description**
We'd like users to be able to specify limits for RAM/disk-space/s3-storeage-usage for the existing cache stores. Ultimately the store(s) should have configurable policies for prioritizing cache entries based on the limits, but the simplest policy is to reject new items once a limit has been reached.

**Describe the solution you'd like**

**Sample Code**
```
If possible, code that demonstrates how the feature would be used.
```

**Additional context**

PR #218, opened 2019-02-13
Initial RAM Cache

jmilloy:
This is just a quick RAM cache

 * I branched off of feature/cache-s3-0.2.2, so you should merge that first.
 * I used `node.json` and `coordinates.json` directly in the key to avoid hash collisions.
 * The cache is thread-safe, but not yet accessible across multiple processes. We can make a new issue for that. This way we can at least start using it by default.
 * There is a stub for #213
 * There are tests passing in podpac/core/cache/test/test_ram_store.py
 * I factored out `cache_mode_matches`.


jmilloy:
BTW I think this will be passing once the optional_imports -> lazy_imports are fixed.

drc-creare:
> 
> 
> BTW I think this will be passing once the optional_imports -> lazy_imports are fixed.

Thanks @jmilloy, that does need to get fixed - in particular, after I merge in the develop branch changes that get rid of optional_imports. But, I think the current failure, at least for the s3 cache, is because I run a check on being able to access the s3 bucket when initializing the S3CacheStore. It fails when there are no legit S3 credentials. I should maybe change this to log a warning instead of throwing an error.

drc-creare:
Looks great!

PR #220, opened 2019-02-13
Removes deprecated NumpyArray, etc

jmilloy:
NumpyArray was marked for deprecation in 0.2.0 and I guess we missed that. I think we should remove it for 1.0.0.

The `Node.style` attribute is also marked for deprecation also, but I didn't remove it. Should we (can we) remove it?

I didn't find anything else.


mpu-creare:
Thanks for doing this. 
We should have a conversation about '.style'. I'm still using it on another project -- I just need a different mechanism...

Issue #205, opened 2019-01-21
Every node needs a default cache_ctrl based on the local settings


mpu-creare:
Right now cache_type has to be specified for each node individually. By default, each node should at least have a RAM control, and anything in addition based on the settings. 

So basically: 
* Default is RAM if nothing else is specified
* Default it based on settings files
* Default is modified by nodes based on what developer has implemented
* Node cache is modified by user upon node creation.

drc-creare:
This makes sense. Thanks @mpu-creare . 

We do not yet have a RAM cache store. Should we try to add this to 0.3.0 Release? 

Alternatively, we could could have the default cache_ctrl be no cache store. It seems to me that it would be reasonable to expect the cache_ctrl to "function" when initialized with an empty list of cache_stores. I think that's the way it is written.

mpu-creare:
Note, this requires a ram store to be implemented. @drc-creare  please review what @jmilloy does. 

Note, this doesn't have to be the final implementation.

drc-creare:
@jmilloy It might make sense for you to use the `feature/cache-s3-0.2.2` branch as a starting point. I've done a little reorganization and the documentation is a little better. I can also merge it into develop. Its basically complete. I plan on reorganizing (see issue #214), but after this feature branch is finished.

jmilloy:
@drc-creare 

Okay. I literally just finished implementing a quick ram cache and am about to test. Maybe I'll jump over to the branch for that, too.

jmilloy:
@mpu-creare Should the user be able to completely disable a cache store type in the settings?

mpu-creare:
Yes -- with the caveat that this may potentially break certain nodes.

jmilloy:
So each cache type has three potential states in the settings:

 * enabled, and nodes use this cache for outputs by default
 * enabled, but nodes don't use this cache for outputs by default
 * disabled

Does that sound right? We'll need settings for

 * RAM_CACHE: 'always', 'enabled', or 'disabled'
 * DISK_CACHE: 'always', 'enabled', or 'disabled'
 * S3_CACHE: 'always', 'enabled', or 'disabled'

Or would you prefer something like

 * RAM_CACHE_ENABLED: True or False
 * DISK_CACHE_ENABLED: True or False
 * S3_CACHE_ENABLED: True or False
 * CACHE_OUTPUTS_TO_DISK: True or False
 * CACHE_OUTPUTS_TO_S3: True or False
[...]

PR #222, opened 2019-02-15
Feature/cache defaults and settings

jmilloy:
Primarily addresses #205.

## Cache Settings and their Defaults

Settings and their Defaults

```
    DEFAULT_CACHE : list
        Defines a default list of cache stores in priority order. Defaults to `['ram']`.
    RAM_CACHE_MAX_BYTES : int
        Maximum RAM cache size in bytes. Defaults to ``1e9`` (~1G).
    DISK_CACHE_MAX_BYTES : int
        Maximum disk space for use by the disk cache in bytes. Defaults to ``10e9`` (~10G).
    DISK_CACHE_DIR : str
        Subdirectory to use for the disk cache. Defaults to ``'cache'`` in the podpac root directory.
    S3_CACHE_DIR : str
        Subdirectory to use for S3 cache (within the specified S3 bucket). Defaults to ``'cache'``.
    CACHE_ALL_OUTPUTS : bool
        Automatically cache node outputs to the default cache store(s). Outputs for nodes with `cache_output=False` will not be cached. Defaults to ``False``.
    RAM_CACHE_ENABLED: bool
        Enable caching to RAM. Note that if disabled, some nodes may fail. Defaults to ``True``.
    DISK_CACHE_ENABLED: bool
        Enable caching to disk. Note that if disabled, some nodes may fail. Defaults to ``True``.
    S3_CACHE_ENABLED: bool
        Enable caching to RAM. Note that if disabled, some nodes may fail. Defaults to ``True``.
```

## Updates

 * The default `cache_ctl` for all nodes is determined by settings['DEFAULT_CACHE'], but of course a specific `cache_ctrl` can also be specified. Valid values for the DEFAULT_CACHE are 
    - a list of cache store names, such as `[]` or [`ram`], or `['ram', 'disk']`
    - a single cache store name, such as `'ram'`
    - `None`, which translates to a CacheCtrl with no stores, e.g. `CacheCtrl([])`
 * Caching is turned off by default during testing. Tests requiring a cache have been updated to specify the necessary cache.
 * `SMAPDateFolder` appends a disk cache to its cache_ctrl.

## Bugfixes

 * The DiskCache set its `_root_dir_path` incorrectly during initialization whenever the `root_cache_dir_path` argument was not provided. It always assumed that the `DISK_CACHE_DIR` setting was relative, and it mis-used the `_path_join` method. These were masked because the Node always instantiated the DiskCacheStore with a `root_cache_dir_path`, which is now removed.

## Notes

 * The `CACHE_RAM_MAX_BYTES` and `CACHE_DISK_MAX_BYTES` are just placeholders. They may need to be implemented as an attribute of each RamCacheStore and DiskCacheStore, rather than globally. See #213.
 * Branched of of feature/cache-ram, which in turn is branched off feature/cache-s3-0.2.2.
 * The `cache_types` shortcut has been removed. We may want to add something like that back in. It would be fed as an argument into the `make_cache_ctrl` method.


jmilloy:
@mpu-creare How should the SMAPDateFolder `_cache_ctrl_default` behave?

 * use a DiskCacheStore as the only store (ie, ignore the default cache setting)
 * add the DiskCacheStore to the cache_ctrl (ie, keep the default cache setting and prepend/append the disk cache to the list of cache stores if it is not already there)

mpu-creare:
Option number 2. Append (last option after default).

jmilloy:
This is implemented, but I still need to test, update unit tests, and document.

mpu-creare:
Let me know when this is ready.

jmilloy:
@mpu-creare I just tested and it is working now (after two typo fixes). I need to update existing tests before we can merge, and should write additional tests before we close the issue.

jmilloy:
@mpu-creare There are some tests that fail now that caching is on by default. I think we can write tests that are correct regardless of whether caching is on or off by manually specifying `cache_outputs=True` and `cache_outputs=False` in cases where it matters. But I think that would mean setting `cache_outputs=False` in a lot of places in the tests. And possibly forgetting in the future. What do you think about turning caching off by default in the CI test environment?

I'm not sure if we already have a local settings.json in the CI environment or how hard it is to set that up. We would just need to use `'CACHE_OUTPUTS_DEFAULT':False`, and then tests wishing to use the cache would create nodes with `cache_outputs=True` (or they could temporarily change the `CACHE_OUTPUTS_DEFAULT` setting).

mpu-creare:
There are some strange failures in there. Are they really all due to the cache? I think a package update might be breaking things?

mpu-creare:
I think turning off caching by default in the CI tests are fine -- as long as we have unittests that test the caching.

jmilloy:
I guess I haven't really looked at the CI, but I have the tests passing now with caching off on my machine, so that's a start. I'll sort it out on CI if I can.

jmilloy:
@mpu-creare All yours.

mpu-creare:
When I retrieve something from the RAM cache, is it a copy of the data in the cache? Or a reference. If the latter, when I change the reference, it will probably change the value in the cache...

podpac/core/settings.py
```diff
@@ -19,8 +17,15 @@
 # Settings Defaults
 DEFAULT_SETTINGS = {
     'DEBUG': False,
-    'CACHE_DIR': None,
-    'CACHE_TO_S3': False,
+    'DEFAULT_CACHE': ['ram'],
+    'CACHE_OUTPUT_DEFAULT': True,
+    'RAM_CACHE_MAX_BYTES': 1e9, # ~1GB
+    'DISK_CACHE_MAX_BYTES': 10e9, # ~10GB
```
mpu-creare: @jmilloy @drc-creare  Are these limits implemented?

podpac/core/node.py
```diff
@@ -657,7 +638,7 @@ def cached_function():
                     out = self.get_cache(key)
                 except NodeException:
                     out = func(self)
-                    self.put_cache(out, key, raise_no_cache_exception=raise_no_cache_exception)
```
mpu-creare: What happens now without this flag?

jmilloy:
I added a test and fixed the RAM cache bug in aa4bd1b

podpac/core/settings.py
jmilloy: No, I guess we should remove them or comment them out for now.

podpac/core/node.py
jmilloy: There is always a cache_ctrl, but it may be empty. If the cache_ctrl is empty, then `cache_ctrl.put` will do nothing.

I think if a node requires a cache, it should check that the cache_ctrl is not empty, probably when it is created. Truthfully, if a node requires a cache, that's not what a cache is. We're conflating a local data store with a cache, I guess because it is convenient. If a node wants to use a a disk store, another way it could do that would be to instantiate a DiskCacheStore object directly and then use it, rather than (ab)using the cache. Really, the DiskCacheStore is misnamed, it should just be a DiskStore, which the caching uses and other things can also use.

But I guess I could put `raise_no_cache_exception` back in and make the default `False`.

(I removed cache_ctrl=None because it was redundant. An empty CacheCtrl is the same: if a Node wanted to raise an exception if there was no cache ctrl, it would also want to raise an exception if the cache_ctrl was empty, right?)

mpu-creare:
Ok. That makes sense. I'm find with this behavior. Nodes that require caching or disk storage really should handle that themselves. I fully agree with that. 

Ok. I hope to try this out sometime next week.

mpu-creare:
This looks good. I'll merge when check complete. A note:

* When you delete the cache for the SMAP object, you still avoid reaching back to opendap 
    * This is because the `SMAPDateFolder` add its own disk cache control -- and this does not get propagated down the chain for the compositor
    * I think eventually we'll want an "ARCHIVE_CACHE_DEFAULT" property that will cache the result of `get_data` calls from `DataSource` nodes. 
    * For now the behavior is ok, though not ideal

Issue #177, opened 2018-11-28
Interpolation attributes in the DataSource should be json serializable


mpu-creare:
**Description**
When I define

``` python
node.interpolation = {'method': 'nearest', 'time_tolerance': numpy.timedelta64(1)}
```
Then I get the following error:
``` python
>>> srcs[0].json
​
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
<ipython-input-40-f9c4c4d6806f> in <module>()
----> 1 srcs[0].json

C:\Creare_Data\PODPAC_INSTALL\podpac\podpac\core\node.py in json(self)
    413         share algorithms and pipelines, or run code on cloud services.
    414         """
--> 415         return json.dumps(self.definition)
    416 
    417     @property

C:\Creare_Data\PODPAC_INSTALL\miniconda\envs\podpac\lib\json\__init__.py in dumps(obj, skipkeys, ensure_ascii, check_circular, allow_nan, cls, indent, separators, default, sort_keys, **kw)
    229         cls is None and indent is None and separators is None and
    230         default is None and not sort_keys and not kw):
--> 231         return _default_encoder.encode(obj)
    232     if cls is None:
    233         cls = JSONEncoder

C:\Creare_Data\PODPAC_INSTALL\miniconda\envs\podpac\lib\json\encoder.py in encode(self, o)
    197         # exceptions aren't as detailed.  The list call should be roughly
    198         # equivalent to the PySequence_Fast that ''.join() would do.
--> 199         chunks = self.iterencode(o, _one_shot=True)
    200         if not isinstance(chunks, (list, tuple)):
    201             chunks = list(chunks)

C:\Creare_Data\PODPAC_INSTALL\miniconda\envs\podpac\lib\json\encoder.py in iterencode(self, o, _one_shot)
    255                 self.key_separator, self.item_separator, self.sort_keys,
    256                 self.skipkeys, _one_shot)
--> 257         return _iterencode(o, 0)
    258 
    259 def _make_iterencode(markers, _default, _encoder, _indent, _floatstr,

C:\Creare_Data\PODPAC_INSTALL\miniconda\envs\podpac\lib\json\encoder.py in default(self, o)
    178         """
    179         raise TypeError("Object of type '%s' is not JSON serializable" %
--> 180                         o.__class__.__name__)
    181 
    182     def encode(self, o):

TypeError: Object of type 'timedelta64' is not JSON serializable
```

Problem comes from:
```python
srcs[0].definition
[...]

PR #229, opened 2019-03-22
ENH: add support for interpolation definition via pipeline

mlshapiro:
@jmilloy please review this small, but important change to allow datasource (and interpolation) to be json-ified in pipeline.

I'm not a fan of how this is implemented right now, but it will get the job done until I separate out interpolators into different files


mpu-creare:
See also #177

jmilloy:
Yes, thanks Matt. I think we should stick to the `definition` and `from_definition(...)` convention rather than, which is what I originally suggested in #177. I had also suggested something different like `to_string()` above, which would be okay but not as consistent.

mlshapiro:
@jmilloy  this is again ready to be merged

Issue #152, opened 2018-11-05
Add ability to automatically deal with different CRS and Projections


mpu-creare:
**Description**
One of PODPAC's core features is to automatically deal with geospatial coordinate reference systems and projections in a seamless manner. Right now, all of the datasets have to be in the same CRS and have the same units to work together (interpolation is automatic). 

**Describe the solution you'd like**
Need a mechanism to:
* [ ] specify the `proj4` CRS in coordinates
* [ ] automatically convert CRS in one dataset to the requested CRS

jmilloy:
See also #103

jmilloy:
Note: I've enforced that all 1d coordinates in a Coordinates be the same crs in #163

mlshapiro:
@mpu-creare i am overestimating the time on this one because i think this is one of the more important features to get right. This will be the first thing I work on starting this week.

mlshapiro:
Seems like we use `crs` and `coord_ref_sys` in the coordinates and node objects. Based on proj4 and rasterio, I am going to change these references to `crs` and refer the coordinate `coord_ref_sys` to self.crs

mlshapiro:
hm on second thought, `coord_ref_sys` is everywhere. i'm going to keep this for now and refer `crs` to `coord_ref_sys`.

After @jmilloy finishes the coordinates updates, I will merge those in and then make the vernacular change to `crs`

PR #230, opened 2019-04-02
ENH: allow stacked dimensions to be set from Coordinates or StackedCoordinates

mlshapiro:
This would be helpful to me since I would not have to figure out how dimensions are stacked in order to set them. It seems like there was a good reason to keep the StackedCoordinates components read only - feel free to reject this PR if this goes against a design consideration

Example:

stack_coords = Coordinates([(np.linspace(-10, 10, 21), np.linspace(-30, -10, 21))], dims=['lat_lon'])

stack_coords['lat'] = np.linspace(-50, -10, 21)
stack_coords['lat_lon']['lat'] = np.linspace(-50, -10, 21)  # also works

### DIFF
--- a/podpac/core/coordinates/stacked_coordinates.py
+++ b/podpac/core/coordinates/stacked_coordinates.py
@@ -43,7 +43,7 @@ class StackedCoordinates(BaseCoordinates):
         Coordinates
             lat_lon[lat]: ArrayCoordinates1d(lat): Bounds[0.0, 2.0], N[3], ctype['midpoint']
             lat_lon[lon]: ArrayCoordinates1d(lon): Bounds[10.0, 30.0], N[3], ctype['midpoint']
-            time: ArrayCoordinates1d(time): Bounds[2018-01-01, 2018-01-02], N[2], ctype['midpoint']    
+            time: ArrayCoordinates1d(time): Bounds[2018-01-01, 2018-01-02], N[2], ctype['midpoint']
 
     Parameters
     ----------
@@ -58,7 +58,7 @@ class StackedCoordinates(BaseCoordinates):
         
     """
 
-    _coords = tl.Tuple(trait=tl.Instance(Coordinates1d))
+    _coords = tl.Tuple(trait=tl.Instance(Coordinates1d), read_only=True)
 
     def __init__(self, coords, coord_ref_sys=None, ctype=None, distance_units=None):
         """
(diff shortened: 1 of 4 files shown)


podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -249,6 +249,24 @@ def __getitem__(self, index):
         else:
             return StackedCoordinates([c[index] for c in self._coords])
 
+    def __setitem__(self, dim, c):
+        if not dim in self.dims:
+            raise KeyError("Cannot set dimension '%s' in StackedCoordinates %s" % (dim, self.dims))
+
+        # try to cast to ArrayCoordinates1d
+        if not isinstance(c, BaseCoordinates):
```
jmilloy: This needs to be `if not isinstance(c, Coordinates1d)`.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -58,7 +58,7 @@ class StackedCoordinates(BaseCoordinates):
         
     """
 
-    _coords = tl.Tuple(trait=tl.Instance(Coordinates1d), read_only=True)
+    _coords = tl.Tuple(trait=tl.Instance(Coordinates1d))
```
jmilloy: You can change this underlying attribute to be an `OrderedDict` if it makes other things easier.

I think it should still be `read_only`, and you should use `set_trait('_coords', ...)` to modify it).

jmilloy:
@mlshapiro Totally fine with this. I'll have to modify a tad for the rotated coordinates, but that should be easy enough.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -249,6 +249,28 @@ def __getitem__(self, index):
...
+
+        if dim in self.dims:
+            idx = self.dims.index(dim)
+            coords = list(self._coords)
+            coords[idx] = c
+
+            self._check_sizes([c.size for c in coords])
+            self._check_names([c.name for c in coords])
+            self._check_coord_ref_sys(coords)
+
```
jmilloy: I think this is why I didn't implement it from the beginning... I didn't know whether it would ever be useful to change the StackedCoordinates coordinates for one dimension if you could only do so without changing the size, and it's a bit tricky to set it up to allow you change the size by setting one dimension at a time and then only checking the size, etc afterwards.

podpac/core/coordinates/stacked_coordinates.py
mlshapiro: The tuple is fine since the dims define which is which. If we want to make this an OrderedDict in the future for consistency or other reasons, then we can.

Switched it back to `read_only`

podpac/core/coordinates/stacked_coordinates.py
mlshapiro: To me, it makes a lot of sense that you can only change the size if you change the whole stacked coordinates (i.e. `stack_coords['lat_lon'] = ... `) This would still work because its happening at the `Coordinates` level.  If you change only one dim of a stacked coordinate, you have to check that its consistent

mlshapiro:
pending tests passing, this is ready to merge

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -268,8 +290,8 @@ def name(self):
 
         Stacked dimension names are the individual `dims` joined by an underscore.
         """
-
-        return '_'.join(dim or '?' for dim in self.dims)
+        if any(self.dims):
+            return '_'.join(dim or '?' for dim in self.dims)
```
jmilloy: This is a fine change, can you just change the test to match it?

mlshapiro: sorry, I should have caught that - i'm having trouble getting some tests to pass locally so I missed this one

mlshapiro:
looks like we're all passed

mlshapiro:
well at the end of the day I am going to change the _coords of StackedCoordinates from a tuple to an OrderedDict. See #152

PR #231, opened 2019-04-04
WIP: convert stacked coords to OrderedDict.

mlshapiro:
This is probably 75% done, but has resulted in some unintended consequences:

- `StackedCoordinates` now requires dims/names on creation, otherwise we can't keep track. I could use some index values as names (i.e. 0, 1), but that would be strange if they were never set. I can't use None (like the other implementing BaseCoordinates) since we have multiple dimensions using the same None key overwriting each other. For now, I required name. 
- Because of the name requirements, `clinspace` is not as intuitive for stacked coordinates. I removed `clinspace` support for `StackedCoordinates`. 
- `StackedCoordinates` end up looking like a mixture between `Coordinates` and `BaseCoordinates`. I think this is intuitive, but it may not be since its really multi-indexed.

Note I removed `read_only`, but after talking with @mpu-creare I will add back in.

At the end of the day, I want the stacked coordinates to be intuitive. Before I go deeper lets decide if this is the right path.

Alternatively, I can just change the `StackedCoordinates._coords` type to be a list, which will let me implement the transform work. That will have less consequences, though I've already gone through the effort to implement this, I feel like its worth considering.

### DIFF
--- a/podpac/core/coordinates/cfunctions.py
+++ b/podpac/core/coordinates/cfunctions.py
@@ -36,7 +36,7 @@ def crange(start, stop, step, name=None):
 
 def clinspace(start, stop, size, name=None):
     """
-    Create uniformly-spaced 1d or stacked coordinates with a start, stop, and size.
+    Create uniformly-spaced 1d coordinates with a start, stop, and size.
 
     For numerical coordinates, the start and stop are converted to ``float``. For time coordinates, the start and stop
     are converted to numpy ``datetime64``. For convenience, podpac automatically converts datetime strings such as
@@ -44,10 +44,10 @@ def clinspace(start, stop, size, name=None):
 
     Arguments
     ---------
-    start : float, datetime64, datetime, str, tuple
-        Start coordinate for 1d coordinates, or tuple of start coordinates for stacked coordinates.
-    stop : float, datetime64, datetime, str, tuple
-        Stop coordinate for 1d coordinates, or tuple of stop coordinates for stacked coordinates.
+    start : float, datetime64, datetime, str
+        Start coordinate for 1d coordinates
+    stop : float, datetime64, datetime, str
+        Stop coordinate for 1d coordinates
     size : int
         Number of coordinates.
     name : str, optional
@@ -67,15 +67,9 @@ def clinspace(start, stop, size, name=None):
         raise ValueError("Size mismatch, 'start' and 'stop' must have the same size (%s != %s)" % (
             np.array(start).size, np.array(stop).size))
     
-    # As of numpy 0.16, np.array([0, (0, 10)]) no longer raises a ValueError
-    # so we have to explicitly check for sizes of start and stop (see above)
-    a = np.array([start, stop])
-    if a.ndim == 2:
-        c = StackedCoordinates([UniformCoordinates1d(start[i], stop[i], size=size) for i in range(a[0].size)])
-    else:
-        c = UniformCoordinates1d(start, stop, size=size)
+    c = UniformCoordinates1d(start, stop, size=size)
 
     if name is not None:
         c.name = name
-        
+    
     return c
--- a/podpac/core/coordinates/coordinates.py
+++ b/podpac/core/coordinates/coordinates.py
@@ -101,6 +101,10 @@ class Coordinates(tl.HasTraits):
         if len(dims) != len(coords):
             raise ValueError("coords and dims size mismatch, %d != %d" % (len(dims), len(coords)))
 
+        for i, dim in enumerate(dims):
+            if not dim:
+                raise ValueError("Dim for coordinates at position %d is not defined" % (i))
+
         # get/create coordinates
         dcoords = OrderedDict()
         for i, dim in enumerate(dims):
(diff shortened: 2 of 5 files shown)


podpac/core/coordinates/cfunctions.py
```diff
@@ -36,18 +36,18 @@ def crange(start, stop, step, name=None):
 
 def clinspace(start, stop, size, name=None):
```
jmilloy: I think the stacked linspace is important, and I think we can use the name argument here to support it. Something like this

```
    ...
    dims = name.split('_')
    if np.array(start).size != len(dims):
        raise size mismatch

    a = np.array([start, stop])
    if a.ndim == 2:
        c = StackedCoordinates([UniformCoordinates1d(start[i], stop[i], size=size) for i, dim in enumerate(dims)])
    else:
        c = UniformCoordinates1d(start, stop, size=size, name=name)
    return c
```

podpac/core/coordinates/coordinates.py
```diff
@@ -784,7 +802,7 @@ def udrop(self, dims, ignore_missing=False):
                 if c.name not in dims:
                     cs.append(c)
             elif isinstance(c, StackedCoordinates):
-                stacked = [s for s in c if s.name not in dims]
+                stacked = [s for s in c.values() if s.name not in dims]
```
jmilloy: Let's try to use either `c.values()` or `c._coords.values()` throughout; I think we have both at the moment.

podpac/core/coordinates/coordinates.py
```diff
@@ -110,8 +114,20 @@ def __init__(self, coords, dims=None, coord_ref_sys=None, ctype=None, distance_u
             if isinstance(coords[i], BaseCoordinates):
                 c = coords[i]
             elif '_' in dim:
-                cs = [val if isinstance(val, Coordinates1d) else ArrayCoordinates1d(val) for val in coords[i]]
-                c = StackedCoordinates(cs)
+                # make sure components of stacked coordinates have dim names from stacked dim
```
jmilloy: Hmmm. This is messy.

Is this a new check that we weren't doing before (to check for a name mismatch with existing Coordinates1d), or is this just because we need to set it now (instead of later) so we have to the check now?

I think you also need to check that `len(sdims) == len(coords[i])`.

podpac/core/coordinates/coordinates.py
```diff
@@ -866,7 +884,14 @@ def unique(self):
             New Coordinates object with unique, sorted coordinate values in each dimension.
         """
 
-        return Coordinates([c[np.unique(c.coordinates, return_index=True)[1]] for c in self.values()])
+        coords = list(self.values())
```
jmilloy: What happened here? I'm curious what didn't work about the previous version.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -222,15 +227,42 @@ def copy(self):
             Copy of the stacked coordinates, with provided properties and name.
         """
 
-        return StackedCoordinates(self._coords)
+        return StackedCoordinates(list(self._coords.values()))
```
jmilloy: This is a sort of a good argument *against* using a dictionary for _coords.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -240,14 +272,11 @@ def __iter__(self):
     def __len__(self):
         return len(self._coords)
 
-    def __getitem__(self, index):
-        if isinstance(index, string_types):
-            if index not in self.dims:
-                raise KeyError("Dimension '%s' not found in dims %s" % (index, self.dims))
-            return self._coords[self.dims.index(index)]
+    def __getitem__(self, dim):
```
jmilloy: Ahh. I use the indexing I had here a lot. Why did you remove this part

```
    else:
        return StackedCoordinates([c[index] for c in self._coords])
```

podpac/core/coordinates/coordinates.py
jmilloy: Ahh, I see, the indexing is removed. I think you need to put that back, and then you can revert this.

jmilloy:
I'm sort of lukewarm on this. It seems like it simplifies `__getitem__` and `__setitem__` *slightly* and complicates initialization *slightly*. If you could accomplish the same thing by just changing the tuple to a list, I think that would make a lot of sense.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -260,16 +289,39 @@ def __setitem__(self, dim, c):
...
+            return False
+
+        return True
+
+    def __delitem__(self, dim):
+        if not dim in self.dims:
+            raise KeyError("Cannot delete dimension '%s' in StackedCoordinates %s" % (dim, self.dims))
 
-        self.set_trait('_coords', coords)
+        del self._coords[dim]
```
jmilloy: I think you need to check that this doesn't leave the StackedCoordinates with only one dimension.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -222,15 +227,42 @@ def copy(self):
...
         """
 
-        return StackedCoordinates(self._coords)
+        return StackedCoordinates(list(self._coords.values()))
 
     # ------------------------------------------------------------------------------------------------------------------
-    # standard methods, tuple-like
+    # standard dict-like methods
     # ------------------------------------------------------------------------------------------------------------------
+    def keys(self):
```
jmilloy: I probably wouldn't include `keys`, `items`, `get` and `update` in the public StackedCoordinates interface. It does make sense to have `values` in the public interface, though I would probably leave `iter` with the original meaning (now implemented as `return self._coords.values()`) and then you can use that instead `values`. But I think that's a small thing. I like

```
c = StackedCoordinates([<lat>, <lon>])
lat, lon = c
```

But I can see that this might be clearer.

```
c = StackedCoordinates([<lat>, <lon>])
lat, lon = c.values()
```

mlshapiro:
@jmilloy i greatly appreciate the view - i went with changing the `_coords` to being a list and that works, and it also makes intuitive sense.  i'm going to leave this through the release then we'll likely prune this branch

PR #232, opened 2019-04-05
Feature/crs2

mlshapiro:
Enables coordinate transformations using the PROJ4 syntax.

Automatically handles coordinate transformations in DataSource nodes. The data source eval process entails:

- first convert evaluated `coordinates` in the same crs as the `native_coordinates` in the data source node.
- Run the DataSource eval process, including coordinate intersection, interpolation coordinate down-selection, interpolation of data, and output data creation all the in the CRS of the native coordinates
- Just before returning the output, transform the coords of the output UnitsDataArray into the CRS of the original evaluated `coordinates`
- Return UnitsDataArray

There is ONE potential caveat that I haven't probed deeply. If an output UnitsDataArray is included in the `DataSource.eval(coordinates, output=output)`, it will work correctly assuming the output array is in the crs of the evaluated coordinates.  There is no way to verify what crs since we don't track the crs on the UnitsDataArray.  

## Example Usage:

```python
c = Coordinates([[0, 1], [10, 20], ['2018-01-01', '2018-01-02']], dims=['lat', 'lon', 'time'])  # uses default crs
tc = c.transform('EPSG:2193')   # use a gdal string
tc = c.transform('+proj=merc +lat_ts=56.5 +ellps=GRS80')  # use a proj4 string
```

works for stacked coordinates:

```python
stack_coords = Coordinates([(np.linspace(-10, 10, 21), np.linspace(-30, -10, 21))], dims=['lat_lon'])
tc = c.transform('EPSG:2193')   # use a gdal string
tc = c.transform('+proj=merc +lat_ts=56.5 +ellps=GRS80')  # use a proj4 string
```

use in DataSource node:

```python
c = Coordinates([[0, 1], [10, 20], ['2018-01-01', '2018-01-02']], dims=['lat', 'lon', 'time'])
node = Array(source=np.ones((2,2,2)), native_coordinates=c)
ct = c.transform('EPSG:2193')
node.eval(ct)

>> <xarray.UnitsDataArray (lat: 2, lon: 2, time: 2)>
array([[[1., 1.],
        [1., 1.]],

       [[1., 1.],
        [1., 1.]]])
Coordinates:
  * lat      (lat) float64 3e+07 2.987e+07
  * lon      (lon) float64 -3.203e+05 -1.523e+06
  * time     (time) datetime64[ns] 2018-01-01 2018-01-02
Attributes:
    layer_style:  <podpac.core.style.Style object at 0x1c29f5dfd0>
    units:        None
```

[...]

PR #234, opened 2019-04-09
Implements Dependent Coordinates, Rotated Coordinates, and additional indexing and selection

jmilloy:
This PR primarily implements `DependentCoordinates` and one of its specific subclasses `RotatedCoordinates`. It also implements selection in `Coordinates` and `StackedCoordinates`, and indexing in `Coordinates`.

I have added an `idims` property to all coordinates classes and modified the `coords` property to use it so that we can handle unstacked, stacked, and dependent coordinates in the same way.

## DependentCoordinates

```
lat = np.linspace(0, 1, 12).reshape(3, 4)
lon = np.linspace(10, 20, 12).reshape(3, 4)
c = DependentCoordinates([lat, lon], dims=['lat', 'lon'])
coords = Coordinates([c], dims=['lat,lon'])

# can coerce, similar to stacked coordinates
coords = Coordinates([[lat, lon]], dims=['lat,lon'])
```

I know this seems silly. You would never create DependentCoordinates like this, because you would use flat StackedCoordinates instead. (Note: we could actually remove StackedCoordinates altogether and only use DependentCoordinates!) But generic DependentCoordinates is useful as a base class, and it is necessary for implementing multidimensional indexing of dependent coordinates class such as RotatedCoordinates. This is similar in nature to the need for UniformCoordinates1d to sometimes return ArrayCoordinates1d when indexed.

## RotatedCoordinates

```
# using upper left corner and a step/scale
c = RotatedCoordinates(shape=(3, 4), theta=np.pi/4, ulc=[10, 20], step=[1.0, 2.0], dims=['lat', 'lon'])
coords = Coordinates([[lat, lon]], dims=['lat,lon'])

# using upper left corner and lower right corner
c = RotatedCoordinates(shape=(3, 4), theta=np.pi/4, ulc=[10, 20], lrc=[15, 17], dims=['lat', 'lon'])
coords = Coordinates([[lat, lon]], dims=['lat,lon'])

# to / from gdal geotransform
c = RotatedCoordinates(shape=(3, 4), theta=np.pi/4, ulc=[10, 20], step=[1.0, 2.0], dims=['lat', 'lon'])
c2 = RotatedCoordinates.from_geotransform(c.geotransform, c.shape, dims=c.dims)
```

### RotatedCoordinates selection and intersection

These should return a `RotatedCoordinates` object when possible. They do not yet. They currently fall back on the `DependentCoordinates` selection and intersection.

### RotatedCoordinates area_bounds

The area_bounds needs to use rotated segments. It does not yet.

## Coordinates Indexing

A tuple of indexes is mapped to the appropriate dimension:

```
lat = [0, 1, 2, 3, 5]
lon = [10, 20, 30, 40]
time = ['2018-01-01', '2018-01-02', '2018-01-03']
c = Coordinates([lat, lon, time], dims=['lat', 'lon', 'time'])
I = [2, 1, 3]
J = slice(1, 3)
K = 1
assert c[I, J, K].shape == 3, 2, 1

# implied indices
assert c[[I, J]].shape == (3, 2, 3)
assert c[I].shape == (3, 4, 3)
```

Coordinates with stacked or dependent coordinates work as well. The implementation regarding dependent coordinates is tricky -- let me know if you want to know more about that. In particular, note that dependent coordinates indexing will sometimes unravel the coordinates, and I've chosen to return a `StackedCoordinates` object in that case, though a 1d `DependentCoordinates` object *should* work as well.

## Selection

You can now pass a dictionary of bounds to the `select` method in Coordinates and underlying `BaseCoordinates` objects.

```
lat = [0, 1, 2, 3, 5]
lon = [10, 20, 30, 40]
time = ['2018-01-01', '2018-01-02', '2018-01-03']
c = Coordinates([lat, lon, time], dims=['lat', 'lon', 'time'])
c.select({'lat': [1, 3]})
c.select({'lat': [1, 3], 'lon': [15, 25]})
c.select({'alt': [1, 3]}) # no affect
```

Again, this is a bit tricky, so let me know if you want to know more about the implementation, but basically it should work as expected. In particular, note that in dependent and stacked coordinates, selection is a logical AND of the selection in each dimension. And, as above, this is where the dependent coordinates will generally return stacked coordinates after selection.

### bounds and area_bounds

Along with this, I added `bounds` and `area_bounds` properties to the Coordinates, StackedCoordinates, and DependentCoordinates objects. These are dictionaries of the bounds for each dimension.

## ArrayCoordinatesNd

In order to facilitate intersection (and for general consistency), `DependentCoordinates` need to return a `Coordinates1d` object when indexed by dimension name. I added `ArrayCoordinatesNd` which is mostly like `ArrayCoordinates1d` but for coordinates for a single real dimension but with a shape. It's is only used internally and for display.

## Coordinates Properties

I refactored the coordinates properties as part of my effort to avoid cases in Coordinates objects for different underlying BaseCoordinates types. This is one place we may have to be careful merging the units and crs work.

## PolarCoordinates

I also partially implemented `PolarCoordinates` as another practical example for subclassing DependentCoordinates. It is only lightly tested.


jmilloy:
@mpu-creare This is a pretty large PR. If you don't have time, will you review the examples in the description to make sure that you are happy with the API?

@mlshapiro Same thing, I'm not sure how much time you should spend on the gnarly details of this, but if you could look over anything you might be concerned about.

I'm happy to do the merge once you review.

mlshapiro:
@jmilloy i will review this afternoon after 2pm!

I'll take a look at what it takes to merge, but it may make sense for you to merge if you have the deeper changes.

I'll be finishing my branch today shortly too, so we can swap

podpac/coordinates.py
```diff
@@ -7,7 +7,8 @@
 
 from podpac.core.coordinates import Coordinates
 from podpac.core.coordinates import crange, clinspace
-from podpac.core.coordinates import Coordinates1d, ArrayCoordinates1d, UniformCoordinates1d, StackedCoordinates
+from podpac.core.coordinates import Coordinates1d, ArrayCoordinates1d, UniformCoordinates1d
+from podpac.core.coordinates import StackedCoordinates, DependentCoordinates
```
mlshapiro: Did you mean to include `RotatedCoordinates` here?

jmilloy: Yes df6f1c0

podpac/core/settings.py
```diff
@@ -19,6 +19,7 @@
 # Settings Defaults
 DEFAULT_SETTINGS = {
     'DEBUG': False,
+    'DEFAULT_COORD_REF_SYS': 'WGS84',
     'CACHE_DIR': None,
```
mlshapiro: note this will be overwritten by changes in #232

podpac/core/settings.py
```diff
@@ -64,6 +65,8 @@ class PodpacSettings(dict):
 
     Attributes
     ----------
+    DEFAULT_COORD_REF_SYS : str
```
mlshapiro: this will be `DEFAULT_CRS` in #232 

I would actually just revert the changes in this file, or overwrite them with the changes in #232

jmilloy:
I see that the `CoordinateReferenceSystem` Trait will need to be removed in the merge.

podpac/core/coordinates/coordinates.py
```diff
@@ -525,24 +539,10 @@ def dims(self):
...
-    def ndim(self):
-        """:int: Number of dimensions. """
-        
-        return len(self.dims)
-
-    @property
-    def size(self):
-        """:int: Total number of coordinates."""
+    def idims(self):
+        """:tuple: Tuple of dimension index names."""
```
mlshapiro: It would be helpful to have an example here. We should document the use cases for dims, udims, and idims.  I put some in the Coordinates notebook

podpac/core/coordinates/coordinates.py
```diff
@@ -580,15 +580,50 @@ def udims(self):
...
     def coords(self):
         """
         :xarray.core.coordinates.DataArrayCoordinates: xarray coords, a dictionary-like container of coordinate arrays.
         """
 
-        x = xr.DataArray(np.empty(self.shape), coords=[c.coordinates for c in self._coords.values()], dims=self.dims)
+        coords = OrderedDict()
+        for c in self._coords.values():
+            coords.update(c.coords)
+        # TODO just return coords?
```
mlshapiro: I think its good to return the DataArray coords object, unless they provide a simpler means for constructing their coordinates. It shouldn't cost too much and it ensures we are always keeping up to date with xarray changes

podpac/core/coordinates/coordinates.py
```diff
@@ -830,13 +865,61 @@ def intersect(self, other, outer=False, return_indices=False):
...
+    def select(self, lat=None, lon=None, alt=None, time=None, return_indices=False, outer=False):
+        """
+        Get the coordinate values that are within the given bounds for each dimension.
+
+        The default selection returns coordinates that are within the bounds::
+
+            In [1]: c = Coordinates([[0, 1, 2, 3], [10, 20, 30, 40], dims=['lat', 'lon'])
+
+            In [2]: c.select(lat=[1.5, 2.5])
+            Out[2]: TODO
```
mlshapiro: want to fill in the TODO items for select - these have errors, see next comment

podpac/core/coordinates/coordinates.py
```diff
@@ -830,13 +865,61 @@ def intersect(self, other, outer=False, return_indices=False):
...
+        -------
+        selection : :class:`Coordinates`
+            Coordinates object with coordinates within the given bounds.
+        I : list of indices (slices/lists)
+            index or slice for the selected coordinates in each dimension (only if return_indices=True)
+        """
+
+        bounds = {'lat':lat, 'lon':lon, 'alt':alt, 'time':time}
+        selections = [c.select(bounds, outer=outer, return_indices=return_indices) for c in self._coords.values()]
+        return _make_selected_coordinates(selections, return_indices)
```
mlshapiro: _make_selected_coordinates() --> self._make_selected_coordinates()

mlshapiro: _make_selected_coordinates() --> self._make_selected_coordinates()

podpac/core/coordinates/coordinates.py
mlshapiro: I also see a different error when running your PR in the notebook:

It looks like we would want this to be **dict:

```
c.select({'lat': [1, 3]})

>>

KeyError                                  Traceback (most recent call last)
<ipython-input-71-d7c9df0f2790> in <module>()
----> 1 c.select({'lat': [1, 3]})

~/computing/creare/podpac/podpac/podpac/core/coordinates/coordinates.py in select(self, lat, lon, alt, time, return_indices, outer)
    909 
    910         bounds = {'lat':lat, 'lon':lon, 'alt':alt, 'time':time}
--> 911         selections = [c.select(bounds, outer=outer, return_indices=return_indices) for c in self._coords.values()]
    912         return _make_selected_coordinates(selections, return_indices)
    913 

~/computing/creare/podpac/podpac/podpac/core/coordinates/coordinates.py in <listcomp>(.0)
    909 
    910         bounds = {'lat':lat, 'lon':lon, 'alt':alt, 'time':time}
--> 911         selections = [c.select(bounds, outer=outer, return_indices=return_indices) for c in self._coords.values()]
    912         return _make_selected_coordinates(selections, return_indices)
    913 

~/computing/creare/podpac/podpac/podpac/core/coordinates/coordinates1d.py in select(self, bounds, return_indices, outer)
    478                 return self._select_full(return_indices)
    479 
--> 480         bounds = make_coord_value(bounds[0]), make_coord_value(bounds[1])
    481 
    482         # full

KeyError: 0
```

podpac/core/coordinates/dependent_coordinates.py
```diff
@@ -0,0 +1,452 @@
...
+        return d['value']
+
+    @tl.observe('dims', 'idims', 'ctypes', 'units', 'coord_ref_sys', 'segment_lengths')
+    def _set_property(self, d):
+        self._properties.add(d['name'])
+
+    def _set_name(self, value):
+        # only set if the dims have not been set already
+        if 'dims' not in self._properties:
+            self.set_trait('dims', value.split(','))
```
mlshapiro: Can we support 'lat,lon' and 'lat, lon'?  i.e. after the split map each value to get rid of preceding whitespace? I know this is annoying, but its the first thing I did when I tried to use this.

podpac/core/coordinates/rotated_coordinates.py
```diff
@@ -1,11 +1,210 @@
...
 import traitlets as tl
-from podpac.core.coordinates.base_coordinates import BaseCoordinates
+import lazy_import
+rasterio = lazy_import.lazy_module('rasterio')
+
+from podpac.core.utils import ArrayTrait
+from podpac.core.coordinates.dependent_coordinates import DependentCoordinates
+
+class RotatedCoordinates(DependentCoordinates):
+    theta = tl.Float(read_only=True)
```
mlshapiro: do we want a docstring here?

podpac/core/coordinates/rotated_coordinates.py
```diff
@@ -1,11 +1,210 @@
...
+        theta = d['theta']
+        ulc = d['ulc']
+        kwargs = {k:v for k,v in d.items() if k not in ['shape', 'theta', 'ulc']}
+        return RotatedCoordinates(shape, theta, ulc, **kwargs)
+
+    # ------------------------------------------------------------------------------------------------------------------
+    # standard methods
+    # ------------------------------------------------------------------------------------------------------------------
+
+    def __repr__(self):
```
mlshapiro: this is a different __repr__ than DependentCoordinates which it wraps. Just making sure this is on purpose

mlshapiro:
@jmilloy I've updated the podpac-examples repository with a notebook with these examples. All run and make sense to me. Took me a little bit to figure out the DependentCoordinates, but the implementation actually makes a ton of sense. Opens up flexibility for basically anything that can be mapped to a matrix. Very cool.

jmilloy:
@mlshapiro Thanks. I will improve the docstrings, docs, and examples before merging. I think it was worth getting a once over on this before investing too much time on that, but I forgot to mention that.

Looks like I also overlooked tests for `Coordinates.intersect` and `Coordinates.select`.

podpac/core/coordinates/rotated_coordinates.py
jmilloy: Yes, intentional, but I should go back and check that this all looks reasonable when it is in alone and when it is in a Coordinates object.

PR #239, opened 2019-04-16
Feature/basic cache limits

drc-creare:
Basic implementation of caching limits. Throws error when limits are reached. S3 and Disk cache store sizes are determined by iterating through objects (files) and summing up their sizes. Things that are being put into cache are temporarily added to determine size and then removed if limit is exceeded. RAM cache store size is determined via the amount of memory the python process is currently using (via psutil). This is an overestimate but possibly the desired behavior anyway. 

It would be nice to have the S3 and Disk cache stores use ledger/index objects/files in order to keep track of the contents. This would improve the speed of calculating the cache size and would help with caching policies that automatically remove things based on priorities like size and timestamps. However, I may not have time to add this before the release.

@jmilloy I added you as a reviewer since you have worked with the cache system and wrote the RAM cache.

@mpu-creare I added you because you may want to push this through to free me up for other work.

### DIFF
--- a/podpac/core/cache/cache.py
+++ b/podpac/core/cache/cache.py
@@ -762,8 +762,7 @@ class FileCacheStore(CacheStore):
         else:
             self.save_new_container(listings=[listing], path=path)
 
-
-        if self.max_size and self.size() >= self.max_size:
+        if self.max_size is not None and self.size() >= self.max_size:
         #     # TODO removal policy
             self.rem(node=node, key=key, coordinates=coordinates)
             raise CacheException("Cache is full. Remove some old entries and try again.")
@@ -915,7 +914,7 @@ class DiskCacheStore(FileCacheStore):
         else:
             self._root_dir_path = self._path_join([settings['ROOT_PATH'], settings['DISK_CACHE_DIR']])
 
-        if max_size:
+        if max_size is not None:
             self.max_size = max_size
         elif 'DISK_CACHE_MAX_BYTES' in settings and settings['DISK_CACHE_MAX_BYTES']:
             self.max_size = settings['DISK_CACHE_MAX_BYTES']
@@ -1068,7 +1067,7 @@ class S3CacheStore(FileCacheStore):
                                              aws_secret_access_key=aws_secret_access_key)
         self._s3_bucket = s3_bucket
 
-        if max_size:
+        if max_size is not None:
             self.max_size = max_size
         elif 'S3_CACHE_MAX_BYTES' in settings and settings['S3_CACHE_MAX_BYTES']:
             self.max_size = settings['S3_CACHE_MAX_BYTES']
@@ -1238,7 +1237,7 @@ class RamCacheStore(CacheStore):
         if not settings['RAM_CACHE_ENABLED']:
             raise CacheException("RAM cache is disabled in the podpac settings.")
 
-        if max_size:
+        if max_size is not None:
             self.max_size = max_size
         elif 'RAM_CACHE_MAX_BYTES' in settings and settings['RAM_CACHE_MAX_BYTES']:
             self.max_size = settings['RAM_CACHE_MAX_BYTES']
@@ -1281,8 +1280,7 @@ class RamCacheStore(CacheStore):
         if full_key in _thread_local.cache:
             if not update:
                 raise CacheException("Cache entry already exists. Use update=True to overwrite.")
-
-        if self.max_size and self.size() >= self.max_size:
+        if self.max_size is not None and self.size() >= self.max_size:
         #     # TODO removal policy
             raise CacheException("RAM cache full. Remove some old entries and try again.")
 
--- a/podpac/core/cache/cache.py
+++ b/podpac/core/cache/cache.py
@@ -13,6 +13,7 @@ from hashlib import md5 as hash_alg
 import six
 import fnmatch
 from lazy_import import lazy_module
+import psutil
 try:
     import cPickle  # Python 2.7
 except:
@@ -317,6 +318,12 @@ class CacheStore(object):
         if len(self.cache_modes.intersection(modes)) > 0:
             return True
         return False
+
+    def size(self):
+        """Return size of cache store in bytes
+        """
+        raise NotImplementedError
+
 
     def put(self, node, data, key, coordinates=None, update=False):
         '''Cache data for specified node.
@@ -729,6 +736,11 @@ class FileCacheStore(CacheStore):
             If True existing data in cache will be updated with `data`, If False, error will be thrown if attempting put something into the cache with the same node, key, coordinates of an existing entry.
         '''
         self.make_cache_dir(node)
+
+        if self.size() >= self.max_size:
+        #     # TODO removal policy
+            raise CacheException("RAM cache full. Remove some old entries and try again.")
+
         listing = CacheListing(node=node, key=key, coordinates=coordinates, data=data)
         if self.has(node, key, coordinates): # a little inefficient but will do for now
             if not update:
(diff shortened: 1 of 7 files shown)


podpac/core/settings.py
```diff
@@ -19,8 +19,9 @@
     'DEBUG': False,  # This flag currently sets self._output on nodes
     'DEFAULT_CACHE': ['ram'],
     'CACHE_OUTPUT_DEFAULT': True,
-#    'RAM_CACHE_MAX_BYTES': 1e9, # ~1GB    TODO
-#    'DISK_CACHE_MAX_BYTES': 10e9, # ~10GB TODO
+    'RAM_CACHE_MAX_BYTES': 1e9, # ~1GB    TODO
+    'DISK_CACHE_MAX_BYTES': 10e9, # ~10GB TODO
+    'S3_CACHE_MAX_BYTES': 10e9, # ~10GB TODO
```
mpu-creare: Let's make this 100 GB, and remove the TODO's on the comments.

mpu-creare: Actually, 10GB is fine as a default (more conservative). Just need to remove the TODOS.

mpu-creare:
This looks fine to me.

podpac/core/cache/cache.py
```diff
@@ -1180,6 +1254,10 @@ def _get_key(self, obj):
     def _get_full_key(self, node, coordinates, key):
         return (self._get_key(node), self._get_key(coordinates), key)
 
+    def size(self):
+        process = psutil.Process(os.getpid())
+        return process.memory_info().rss # this is actually the total size of the process
```
jmilloy: I think this is okay, but do you want to measure the actual ram cache size, instead? We can do that.

drc-creare: That's a good question. For the use cases that come to mind, I think limiting the entire process would be desirable, but limiting via cache put requests doesn't necessarily accomplish this. Someone could easily exceed the limits without even using the cache.

How do we limit the cache size? I couldn't find a good universal way to determine the size of arbitrary data objects. Would we need to restrict the types of objects that get put into the cache? 

We could serialize the objects (with pickle) and determine the size of the resulting string, but this will generally take up more space then the actual object. Then again storing the objects serialized would allow for inter-process sharing, but we would take CPU/RAM hit serializing/de-serializing everything.

@mpu-creare  any thoughts?

podpac/core/settings.py
drc-creare: Yeah, thanks. One thing about how this is working, the user would have to explicitly set these limits in the settings to `None` in order to achieve unlimited storage. If they leave the setting out it will go to the defaults. There is a performance hit to using any limit (especially with S3).

I should put a note about `None`. Would this go in the `settings.py` comments?

podpac/core/cache/cache.py
jmilloy: > I couldn't find a good universal way to determine the size of arbitrary data objects.

I think you can do good enough, see maybe https://stackoverflow.com/a/40880923/484488

> That's a good question. For the use cases that come to mind, I think limiting the entire process would be desirable, but limiting via cache put requests doesn't necessarily accomplish this. Someone could easily exceed the limits without even using the cache.

I don't exactly disagree, but this hardly accomplishes that. If we want to limit the entire process RAM usage, we need to (1) name the setting `RAM_MAX_BYTES` and (2) somehow also check and prevent *non-cache* objects from being added to RAM.

podpac/core/settings.py
drc-creare: Added not about `None` to `PodpacSettings` class.

podpac/core/cache/cache.py
jmilloy: Oh, I misread your comment, even as I quoted it! I don't agree that *"limiting the entire process would be desirable"*, but I do agree that *"limiting via cache put requests doesn't ~~necessarily~~ accomplish this"*.

drc-creare: Thanks @jmilloy , let me check in with @mpu-creare about priorities. He's in a meeting. I''l hold for now on the merge.

mpu-creare: I like the current approach as implemented. We just need to carefully document it. 
I agree that it might be a little surprising to users, but I think it will be more maintainable in the long run and it's what I prefer for my applications as well. 

I don't like the other approach because in my experience it can be pretty fragile. Especially with all of the in-memory/out-of-memory operations that xarray does with different file formats. 

Also, by limiting the cache based on the entire process RAM, you have a hope of limiting to total RAM of the process (at least the cache won't make it worse than you've specified).

drc-creare: I'll update the [settings docs](https://github.com/creare-com/podpac/blob/develop/doc/source/user/settings.md) before merging.

jmilloy: It is the simplest implementation, and I think it's okay.

But imagine that you have "delete oldest" as your cache policy. You exceed the max ram settings outside of the cache. Then you try to put something in the cache, and the result is that your *entire* RAM cache is deleted. Now your executes very slowly. It would be very difficult to figure out why.

> the in-memory/out-of-memory operations that xarray does with different file formats.

Does this not effect either approach?

In any case, the default should be much larger, maybe the 100MB that you originally suggested, or even 1GB, if we're going to use the setting this way.

mpu-creare:
Its a good idea, but I think @mlshapiro has a way of doing that programmatically -- which will help with keeping it fresh... we should ask him how to do that and use that approach...

drc-creare:
Yeah, that would be much better.

jmilloy:
This just occurred to me. I think raising an exception when `put_cache` fails is a problem. I think we should just issue a warning and return `False` if the cache is full.

As it stands, a node will unexpectedly fail in the middle of evaluation if the cache gets full, and I think that's particularly problematic because we cache all node outputs to RAM by default. User code should continue to work regardless of the cache.

Another option would be to catch the CacheException in the `node_eval` decorator where we call `put_cache`. But I do think the desired behavior in most cases -- certainly all genuine cache usage -- would be to catch the exception so that you can ignore it, so that seems silly. Returning False would allow specific code to check and require put_cache success.

podpac/core/cache/cache.py
```diff
@@ -753,6 +761,12 @@ def put(self, node, data, key, coordinates=None, update=False):
         # if file for listing does not already exist, we need to create a new container, add the listing, and save to file
         else:
             self.save_new_container(listings=[listing], path=path)
+
+        if self.max_size is not None and self.size() >= self.max_size:
+        #     # TODO removal policy
+            self.rem(node=node, key=key, coordinates=coordinates)
+            raise CacheException("Cache is full. Remove some old entries and try again.")
```
jmilloy: This exception (or warning) should advise the user to increase the cache size. Maybe it could suggest both increasing the cache size or removing some old entries. A "full" RAM cache could have no entries in the cache to remove!

drc-creare: Thanks @jmilloy . These are good points. I agree the cache should issue a warning and not an error, and I think the suggestion should be to increase the cache size. 

Specifically, I think that the suggestion should be to modify the limit in the settings module and that the cache store classes should be changed so that the limit is dynamically looked up from the settings module - unless an over-ride limit was given when the cache store object was created (e.g. to `__init__()`). Does this make sense and sound like a good plan?

mpu-creare: Maybe the warning should also document how we ended up implementing the process RAM limitation. Something like
`"Warning: Process is using more RAM than the specified limit in settings.MAX_RAM_CACHE. No longer caching. Consider increasing this limit and restarting the Python process or try clearing the cache (e.g. node.rem_cache(key='*', mode='RAM', all_cache=True) to clear ALL cached results in RAM)"`

If you can easily do a dynamic lookup of the settings that would be preferred over having to restart the process. But otherwise, just make the warning informative.

PR #241, opened 2019-04-18
In pipelines, nodes are copied when debugging is on, to help when debugging.

### LINKED ISSUE #110: Pipelines should create unique instances for nodes that are re-used
**Description**
When a node is re-used in a Pipeline, it may be evaluated more than once with different coordinates. The pipeline should still evaluate correctly, but debugging becomes unreliable. All of the nodes in a Pipeline should be unique instances.

Nodes should be copied if necessary during Pipeline instantiation.

**Additional context**
We may additionally want to ensure that Algorithms and Compositors also don't re-use nodes, either throwing an exception (or warning) or just copying when necessary.

jmilloy:
The change here is very small, and may be sufficient to close #110. See `test_debuggable` in the pipeline tests and the change in `_get_subattr`. I am only copying nodes if debugging is on, but we could also just always deepcopy if you want.

This PR also includes also bugfixes so that pipelines load json definition as OrderedDict.

### DIFF
--- a/podpac/core/pipeline/pipeline.py
+++ b/podpac/core/pipeline/pipeline.py
@@ -9,10 +9,12 @@ import warnings
 import importlib
 from collections import OrderedDict
 import json
+from copy import deepcopy
 
 import traitlets as tl
 import numpy as np
 
+from podpac.core.settings import settings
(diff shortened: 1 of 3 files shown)


mpu-creare:
This is great. Merging.

PR #242, opened 2019-04-18
Automatically copies node traits when debug is on.

jmilloy:
This is a proposed additional fix for #110.

Instead of only copying nodes in pipelines, this copies nodes in all cases when debugging is on.

This example is from `test_debuggable_source` in the `test_signal.py`:

Normal usage:

```
        a = Arange()
        node = Convolution(source=a, kernel=[[1, 2, 1]], kernel_ndim=2)
        node.eval(coords)

        # doesn't copy
        assert node.source is a
```

Debugging:

```
        settings['DEBUG'] = True

        a = Arange()
        node = Convolution(source=a, kernel=[[1, 2, 1]], kernel_ndim=2)
        node.eval(coords)

        # copies
        assert node.source is not a

        # a._requested coordinates are not effected
        assert node._requested_coordinates is not None
        assert node.source._requested_coordinates is not None
        assert a._requested_coordinates is None
```

### DIFF
--- a/podpac/core/algorithm/algorithm.py
+++ b/podpac/core/algorithm/algorithm.py
@@ -22,6 +22,7 @@ from podpac.core.node import NodeException
 from podpac.core.node import COMMON_NODE_DOC
 from podpac.core.node import node_eval
 from podpac.core.utils import common_doc
+from podpac.core.utils import NodeTrait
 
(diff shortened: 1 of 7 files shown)


mpu-creare:
This rocks! Nice work.

Issue #247, opened 2019-04-24
pip installed podpac (v0.3.0) missing nsidc_smap_opendap_url.txt


trey-stafford:
**Description**
Version 0.3.0 of podpac installed via pip does not include datalib/nsidc_smap_opendap_url.txt

**Steps to Reproduce**
```
$ pip install podpac[all]
$ pip show podpac
Name: podpac
Version: 0.3.0
Summary: Pipeline for Observational Data Processing, Analysis, and Collaboration
Home-page: https://github.com/creare-com/podpac
Author: Creare
Author-email: UNKNOWN
License: APACHE 2.0
Location: /home/trst2284/miniconda3/envs/podpac/lib/python3.7/site-packages
Requires: scipy, matplotlib, requests, xarray, lazy-import, pint, traitlets, numpy
Required-by:
```

**Expected Behavior**
I expect to be able to use `podpac.datalib.SMAP` to retrieve data from NSIDC's OpenDAP instance.

**Observed Behavior**
Observed in https://github.com/creare-com/podpac-examples/blob/develop/notebooks/basic_examples/010-retrieving-SMAP-data.ipynb

when running cell 5:

```
WARNING:podpac.datalib.smap:Could not retrieve SMAP url from /home/trst2284/miniconda3/envs/podpac/lib/python3.7/site-packages/podpac/datalib/nsidc_smap_opendap_url.txt: [Errno 2] No such file or directory: '/home/trst2284/miniconda3/envs/podpac/lib/python3.7/site-packages/podpac/datalib/nsidc_smap_opendap_url.txt'

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/miniconda3/envs/podpac/lib/python3.7/site-packages/traitlets/traitlets.py in get(self, obj, cls)
    527         try:
--> 528             value = obj._trait_values[self.name]
    529         except KeyError:

KeyError: 'base_url'

During handling of the above exception, another exception occurred:

UnboundLocalError                         Traceback (most recent call last)
~/miniconda3/envs/podpac/lib/python3.7/site-packages/podpac/datalib/smap.py in SMAP_BASE_URL()
    215         if 'https://' in r and 'nsidc.org' in r:
--> 216             if rf != r:
    217                 _logger.warning("Updating SMAP url from PODPAC S3 Server.")

UnboundLocalError: local variable 'rf' referenced before assignment

During handling of the above exception, another exception occurred:

NameError                                 Traceback (most recent call last)
[...]

PR #249, opened 2019-04-24
Hotfix - SMAP url fixes

jmilloy:
See #247

I don't feel confident fully vetting these changes to SMAP on my own.

I'm not sure if we should create a 0.3.1 release for this hotfix, since 1.0.0 is coming soon, and there is a workaround mentioned in the issue. Nonetheless, this fix should be tested and applied to develop, as the issues are still outstanding there.


mpu-creare:
This looks fine. I think we'll skip the 0.3.1 since 1.0.0 is so close. 

Thanks for the fix.

PR #250, opened 2019-04-26
Replace BaseCoordinates units with Coordinates alt_units, and integrate with transform

### LINKED ISSUE #244: Implement or replace Coordinates units
**Description**
The coordinates `units` are not integrated with the crs transform (otherwise used) and may be the incorrect type. The `crs` contains the lat-lon units and *may* also contain the altitude (`vunits`). The `units` provides a way to specify the altitude units separately from the crs.

Currently `units` is a podpac `Units` object defined on each of the underlying coordinates. I don't think this adequately serves our needs. We do not need units for time, lat, or lon coordinates; we only need units for alt at the moment. These units need to match (or not contradict) the crs and need to be integrated into the transform method.

*Describe the solution you'd like**

I propose an `alt_units` trait defined on the `Coordinates` object. It should be a string that is compatible with proj4 vunits (or it can be None). If the crs contains a complete proj4 string with vunits, the `alt_units` should match (or be None). The transform method `alt_units` argument will correspond clearly with the Coordinates `alt_units` trait. The current `units` trait defined on all underlying coordinates would be removed.

**Additional notes**

 * We can add additional units for additional dimensions when and if they are added in the future.
 * The `distance_units` argument to `Coordinates` and `StackedCoordinates` should be removed regardless, as lat and lon units seem to be handled in the crs.

jmilloy:
closes #244

### DIFF
--- a/podpac/core/cache/test/test_disk_store.py
+++ b/podpac/core/cache/test/test_disk_store.py
@@ -1,332 +1,332 @@
-import pytest
-import numpy as np
-import os
+# import pytest
+# import numpy as np
+# import os
 
-from numpy.testing import assert_equal
+# from numpy.testing import assert_equal
 
-from podpac.core.cache.cache import CacheException
-from podpac.core.cache.cache import CacheCtrl
-from podpac.core.cache.cache import CacheStore
-from podpac.core.cache.cache import DiskCacheStore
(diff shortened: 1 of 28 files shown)


jmilloy:
This is failing because of various cache test failures.

PR #255, opened 2019-04-29
Feature/intake

mlshapiro:
Preliminary support for intake catalogs for the 1.0 release.

This is only a datalib, so it does not affect any of the podpac codebase.

I would merge this directly into develop except the intake dependency did not install 1st try on my machine. There was an issue installing `python-snappy`.  If you have no issues installing, then this should be good to merge.

This is a very preliminary implementation and has been tested with the Intake CSV examples.  More support to come.

See demo here: https://github.com/creare-com/podpac-examples/blob/develop/notebooks/demos/intake.ipynb (Unfortunately intake does not support CSV files outside of the current directory hierarchy so I had to put data in the demos folder)


mlshapiro:
well it looks like the CI has the same issue with python-snappy. We have to specify a specific version of stdc when installing.  I don't know how we can install intake without this, so maybe we should wait. 

Since this doesn't affect the library we can add this as an intermediate release at some point.

mlshapiro:
well should have looked - intake only supports python > 3.6.

mpu-creare:
Given the issues with snappy and only supporting 3.6, I think we'll shoot for the 1.1.0 release for this one. Thanks for pushing to try and get that done, but I don't want to rush it. 

Neither of the above are show-stoppers, we just want to be careful about how we handle it...

mlshapiro:
this still only supports py >=3.6, but its an optional dependency for us. I think its pretty harmless to merge this in at this point.

mlshapiro:
perhaps the right way to deal with this is to remove the `intake` key from setup.py and just tell the user to install intake if we fail to import it...

mpu-creare:
We could add a test for python 3 in our setup.py file and remove intake in case of python 2.7. That also won't break our devall install. How does that sound?

mpu-creare:
@mlshapiro are you waiting for me to review and merge this?

mlshapiro:
I was waiting until I could get the tests to pass. I just merged in Develop to reconcile the setup.py issues. This is not waiting on anything, so I'll merge it in

PR #257, opened 2019-04-29
Release/1.0.0

mlshapiro:
Initiate 1.0.0 release procedure


mlshapiro:
Note to update README copy (potentially) with new stability message

jmilloy:
We have a coverage issue. The two primary culprits are the caching and the data types, though the missed caching coverage is due to the S3 Cache, which may be the desired behavior. @mpu-creare Do we need to fix this so it is above 90% before we can release?

jmilloy:
The S3CacheStore has been excluded from the coverage; there are tests but we do not run them in CI.

mlshapiro:
Lambda function needs to get updated (see fix/aws)

dsully1992:
If I'm not mistaken, the Lambda function update was the final roadblock for this release. That is done, so I believe we are good to go.

PR #259, opened 2019-04-30
S3 test coverage

jmilloy:
Uses `# pragma: no cover` to exclude some S3 code from the coverage report:
 * `S3CacheStore`

### DIFF
--- a/podpac/core/cache/cache.py
+++ b/podpac/core/cache/cache.py
@@ -1235,7 +1235,6 @@ class S3CacheStore(FileCacheStore):
 
 
 _thread_local = threading.local()
-_thread_local.cache = {}
 
 class RamCacheStore(CacheStore):
     """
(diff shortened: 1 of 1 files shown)


jmilloy:
@drc-creare @mpu-creare Is this appropriate? Are there additional classes and methods (the `S3` Node) that should be included as well? Or should we be actually writing tests for this instead?

jmilloy:
Note that the cache commit is already pushed to release/1.0.0. Not sure why it is showing here, maybe that will go away at some point.

drc-creare:
@jmilloy Thanks, I think this is appropriate given the testing strategy for AWS resources (namely they aren't included in the automated testing). The tests exist for the S3 cache, they are just run manually. We could included them in the automated testing. But worth noting that they are slow, would incur some AWS charges, and we would have to add AWS credentials to the docker container that runs the tests. Maybe we would want them run less frequently.

Issue #62, opened 2018-06-13
stats Reduce nodes do not work for all chunk_sizes


jmilloy:
**Description**
The stats Reduce nodes return incorrect output for some chunk_sizes.

**Steps to Reproduce**
```
coords = Coordinate(
    lat=(0, 1, 10),
    lon=(0, 1, 20),
    time=('2018-01-01', '2018-01-30', '1,D'),
    order=['lat', 'lon', 'time'])

a = np.random.random(coords.shape)
source = NumpyArray(source=a, native_coordinates=coords)
data = source.execute(coords)
node = Mean(source=source)

output = node.execute(coords)
output_10 = node.execute(coords, {'iter_chunk_size': 10})
output_20 = node.execute(coords, {'iter_chunk_size': 20})
output_1000 = node.execute(coords, {'iter_chunk_size': 1000})
```

**Expected Behavior**
The output should be the same for all chunk sizes and for non-chunked execution.

**Observed Behavior**
 - chunk_size 10 works correctly (but is very slow)
 - chunk_size 1000 now works correctly (and is fast)
 - chunk_size 20 does not work correctly (and is also slow)

**Additional Notes**
 * This is not specific to the Mean node, but to the Reduce node in general.
 * It's not surprising that the small chunk size execution is slower, but it is much slower than I would expect. That should probably be a separate issue.

jmilloy:
Fixed in https://github.com/creare-com/podpac/pull/68

jmilloy:
This issue is partially fixed, but some cases still do not work. The description is updated.

PR #273, opened 2019-05-25
Stats (reduce nodes) test updates

jmilloy:
I haven't been able to replicate the bug in #62, and I believe the slowness there is surprising but not an error (at least not due to the chunking code).


jmilloy:
@mpu-creare @mlshapiro Any idea why this is causing the unrelated units/image test failures?

jmilloy:
I make a test branch that is identical to develop and it also failed on the CI. I'm just going to merge this and we can fix the other issue separately.

PR #278, opened 2019-06-26
non-pickle caching

jmilloy:
Data objects are cached to file using different formats (and extensions) based on type:
 * xarray Dataset and DatArray uses `to_netcdf` and `open_dataset`
 * podpac Coordinates uses JSON
 * otherwise, JSON is when possible
 * pickle is still used as a fallback

Note that this currently does not handle hash collisions.


jmilloy:
@mpu-creare not ready yet.

mpu-creare:
That is one strange unit test failure...

jmilloy:
I think these unit test failures also fail on develop. There are other issues with this branch, though. Should be ready this afternoon.

mpu-creare:
Ah, right you are. Looks like a unicode error in the unittest probably...

jmilloy:
@drc-creare How do I the test_s3_store tests? Is it not working because I don't have access to the podpac-internal-test bucket? Can you run them for me?

jmilloy:
@mpu-creare I haven't tested the S3Store, but the DiskStore with these changes is working. You could give this a try.

jmilloy:
Hash collisions are very unlikely, but because of the sort of hidden filename cleansing, I think you can create filename collisions quite easily. We discussed using a metadata file for each data file to handle this, but that isn't implemented here. Let me know if you want me to do that before you merge. (Alternately, we can change the filename cleansing so that it isn't a problem. I think we should do this anyways, but I know it may be lower priority).

podpac/core/cache/cache.py
```diff
@@ -738,39 +540,49 @@ def put(self, node, data, key, coordinates=None, update=False):
...
+        
+        if isinstance(data, xr.DataArray):
+            path = rootpath + '.netcdf'
+            s = data.to_netcdf()
+        elif isinstance(data, np.ndarray):
+            path = rootpath + '.npy'
+            with io.BytesIO() as f:
+                np.save(f, data)
+                s = f.getvalue()
+        elif isinstance(data, podpac.Coordinates):
```
mpu-creare: you could also add an entry for nodes, which also have the JSON property...

podpac/core/cache/cache.py
```diff
@@ -795,16 +607,45 @@ def get(self, node, key, coordinates=None):
...
+
+        if path is None:
+            raise CacheException("Cache miss. Requested data not found.")
+
+        # read
+        s = self._load(path)
+        
+        # deserialize
+        if path.endswith('.netcdf'):
+            data = xr.open_dataarray(s)
```
mpu-creare: This `data` should be fed to a `UnitsDataArray` because we have some special deserialization that happens for units that may be part of the metadata.

podpac/core/cache/test/test_disk_store2.py
```diff
@@ -0,0 +1,272 @@
...
+        
+        store.has(NODE1, 'mykey1') is False
+        store.has(NODE1, 'mykey2') is False
+        store.has(NODE1, 'mykeyA', COORDS1) is False
+        store.has(NODE1, 'mykeyB', COORDS1) is False
+        store.has(NODE1, 'mykeyA', COORDS2) is False
+        store.has(NODE2, 'mykey1') is False
+        store.has(NODE2, 'mykeyA', COORDS1) is False
+
+    def test_cache_xarray(self):
```
mpu-creare: could you also please add a test for a `UnitsDataArray`

podpac/core/cache/cache.py
jmilloy: I left it out originally because it seems like a weird thing to cache, but I pushed this. The implementation is weird because we don't have a `from_json` for nodes in general, we only have it for `Pipeline` nodes.

mpu-creare: Well, I certainly didn't think that through... well, now that it's in I think it's fine to leave. We'll likely fin d a use for it sometime.

mpu-creare:
I missed the comment about the filename caching. Aren't the hashes inside the filenames? so if the hashes do not clash, the filenames won't either?

jmilloy:
@mpu-creare I'm looking at it more closely, now. The code is complicated. I think it used to cleanse the input key (which is in the filename directly), so that say`cache.put(node, data, 'my:key')` and `cache.put(node, data, 'my_key')` would use same filename. Also say `cache.put(node, data, 'xxxx')` and `cache.put(node, data, 'cKey')`. I think now (in this PR) it cleanses the key on rem but not on get, has, and put anymore which is a mistake on my part. But I think cleansing the key is the wrong thing to do anyways: we should just hash the key, too (or disallow certain characters in the key if we really have to). A lot of the cache implementation doesn't make sense to me, so I'm sort of struggling to make the necessary changes without rewriting more than necessary. For example, the filenames have a prefix that needs to be cleansed as well, but why is there a prefix at all? Why are there special strings (`nKey`, `kKey`, and `cKey`) in the filenames, which cause other cleansing issues? Why does `cache.rem(node='*', key='mykey')` ignore the key and remove the entire cache? etc.

drc-creare:
@jmilloy , maybe I can help answer some of these questions. I can see why you were confused. It is complicated and I am certainly open to simplifying it where we can. Some knowledge behind the motivation will help us retain any desirable behavior.

1. filename cleansing should have happened after the hash integer was calculated for each of the three hashed entities (node, coordinates, and key). This way 'my_key' and 'my:key' would hash to different integers. If that wasn't happening, it was probably a mistake on my part. It was just a way to avoid illegal characters in filenames and to enforce that splitting the filename on underscore would yield integer strings that corresponded to the three hashed components (for searching purposes). 

2. The prefix existed because at one point there was a desire to allow for human readable parts of the filenames so that the cache filesystem would be less opaque to a human (possibly debugging). I think we could do away with it.

3. `cache.rem(node='*', key='mykey')` should not silently ignore the key. At the least it should throw a warning and probably error. Thanks for noting this. I think this was a mistake on my part. The three key components (node, key, coordinates) are designed to be hierarchical. Nodes (same pipeline definition) get there own "playground" and can use whatever `key` naming scheme they want without regard to the scheme employed in other nodes. That's why I did not think of the intent behind `cache.rem(node='*', key='mykey')` as a valid input. But, I'm flexible on this if we want to facilitate a query like this. Otherwise, we should throw an error instead of silently clearing the cache. lastly, coordinates were the most specific of the key hierarchy to allow for the ability to add in geospatially smarter cache retrievals (e.g. requests for a sub region of a cached region). 

Thanks for working on this. Its good to have a second set of eyes.

drc-creare:
@jmilloy I'm putting together some credentials so you can run the S3 tests. I'll get in touch outside of this public forum.

mpu-creare:
Couple of things. I'm the one who wanted human-readable cached information... and I still want that :) . In fact, I'm likely responsible for the mistakes both @jmilloy  and @drc-creare  are trying to claim. But they're mine, MINE I tells ya! 

These were last-minute hacks to get demos working and then never revisiting. I'm glad we're catching them now. 

Could we: 
* hash everything and put it in the filename
* ALSO include plain-text but cleaned version for the `key` and the prefix. That way key_key and key:key would have different hashes, but show up as 'key_key' in both files.

jmilloy:
Ugh. Yes, I guess we can do both. So the format should be `<cleaned-node-prefix>-<cleaned-key>_<node-hash>_<coordinates-hash>_<key-hash>`. The full prefix (cleaned node prefix and cleaned key) will have underscores removed, so searching will be things like
 - `*_<node-hash>_<coordinates-hash>_<key-hash>`.
 - `*_<node-hash>_*_<key-hash>`.
 - `*_<node-hash>_<coordinates-hash>_*`.
 - `*_<node-hash>_*_*`.

jmilloy:
I thought `cache.rem(node='*', key='output') made sense because we have the `output` key used for all nodes. The way that the files are in subdirectories using the node means that implementing that is annoying, though. I think we should require the `node` in `cache.rem`, and then provide `cache.clear()` (which DRC already implemented and is already public) to clear the whole cache.

drc-creare:
These sound like good changes to me.

I like requiring `node` in `cache.rem()`, makes sense to make  clear cache a distinct function.

Though, If we want to allow for something like one rem call to clear all outputs across all nodes, we could flatten the directory structure and move class hierarchy info to the filename prefix. That said, some filesystems limit the number of files in a folder and also limit filename length.

jmilloy:
Alright, I pushed the new filenaming scheme, the `cache.rem` and `cache.clear` change, and ironed out some other wrinkles. I also separated `cache.py` into separate files, which I know @drc-creare has been wanting to do anyways. I still need to check the s3 caching. I'll do that now and then let you know when this is ready to co.

jmilloy:
@mpu-creare @drc-creare This is all set. I should write `cache_ctrl` tests, but it is ready to be merged and/or used.

mpu-creare:
How am I supposed to clear the entire cache now?

jmilloy:
cache_ctrl.clear()

mpu-creare:
So, I need to remove the `all_cache` flag from `Node.rem_cache()`

jmilloy:
We can either
 - re-institute `cache_ctrl.rem(node="*")`, with a better implementation (1 hour)
 - remove `all_cache` from `Node.rem_cache`
 - replace `self.cache_ctrl.rem('*', '*')` with `self.cache_ctrl.clear()` in `Node.rem_cache`

mpu-creare:
Cool, thanks for the options. I'll take option #2. I'll fix it, I'm finding a few more bugs...

jmilloy:
Okay, thanks. I'm happy to follow behind after your fixes and add unit tests that cover them.

podpac/core/cache/cache_ctrl.py
```diff
@@ -0,0 +1,264 @@
...
+        """
+        Clear all cached data.
+
+        Parameters
+        ------------
+        mode : str
+            determines what types of the `CacheStore` are affected: 'ram','disk','network','all'. Defaults to `node._cache_mode` or 'all'. Overriden by `self._cache_mode` if `self._cache_mode` is not `None`.
+        """
+
+        if not isinstance(node, podpac.Node):
```
drc-creare: I think `node` is undefined here (in `cache_ctrl.clear()`).

jmilloy: Yep, thanks. Need those cache_ctrl tests.

mpu-creare: I'm working on a commit that fixes this one.

drc-creare: thanks guys. I'll keep looking, and comment if I see anything else. This is big rewrite. I appreciate the restructuring into multiple files, `cache.py` was getting too big.

mpu-creare:
Seems to work well. 

Just a note: when `podpac.settings[''CACHE_OUTPUT_DEFAULT'] == True`, we cache output from a SMAP eval 4 times! One for the source node, two for the datefolders (why two?) and one for the main compositor. 

I think this is the correct behavior in general, but I think I should fix this for the SMAP node in particular... although the openDAP implementation may be deprecated by the EGI interface, so I'll wait on those.

PR #279, opened 2019-07-01
Feature/egi

mlshapiro:
Missing pieces:

- Need to query the granule size limit so that if someone requests a large temporal/spatial extent, we can get the data via pages (See [FAQ](https://developer.earthdata.nasa.gov/sdps/programmatic-access-faqs))
- Decide what to do the `authentication.EarthDataSession`. I chose not to use this in EGI, and I'm not sure there is a use case outside of SMAP and opendap
- ~~Fix coordinate transformation issue with SMAP Level 4 data~~
- Cache data 
- Python 2.7 support for Smap Level 3


mlshapiro:
aside from the image thats breaking tests, and pending the py27 test, this should be ready to go

podpac/core/data/datasource.py
```diff
@@ -398,12 +397,13 @@ def eval(self, coordinates, output=None):
...
-        # transform the output crs to the requested coordinates crs
-        # TODO: this could be made into method in/for UnitsDataArray 
+
+        # if requested crs is differented than native coordinates,
+        # fabricate a new output with the original coordinates and new values
         if self._evaluated_coordinates.crs != coordinates.crs:
-            t_coords = coordinates.transform(self._evaluated_coordinates.crs)
-            output = output.assign_coords(**t_coords.coords)
+            trans_output = self.create_output_array(self._evaluated_coordinates)
+            trans_output[:] = output[:].values
```
mlshapiro: @jmilloy @mpu-creare I decided to replace the output coordinates with the originally requested coordinates if the crs has to be changed. This avoids issues when returning Dependent Coordinates

podpac/core/node.py
```diff
@@ -544,7 +544,7 @@ def wrapper(self, coordinates, output=None):
             self._from_cache = False
 
         # transpose data to match the dims order of the requested coordinates
-        order = [dim for dim in coordinates.dims if dim in data.dims]
+        order = [dim for dim in coordinates.idims if dim in data.dims]
```
mlshapiro: @jmilloy this will help if the requested coordinates are Dependent. There is still an uncertain case when an `output` is provided but is asymmetric (Depedent/Not-Dependent) from the request or native coordinates

podpac/core/coordinates/array_coordinates1d.py
```diff
@@ -347,14 +347,21 @@ def _select(self, bounds, return_indices, outer):
...
             except ValueError as e:
-                gt = self.coordinates >= -np.inf
+                if self.dtype == np.datetime64:
+                    gt = ~np.isnat(self.coordinates)
+                else:
+                    gt = self.coordinates >= -np.inf
             try:
                 lt = self.coordinates <= min(self.coordinates[self.coordinates >= bounds[1]])
             except ValueError as e:
-                lt = self.coordinates <= np.inf
```
mpu-creare: Looks like some kind of special case bug you ran into?

podpac/core/coordinates/coordinates.py
```diff
@@ -1264,6 +1264,8 @@ def transform(self, crs=None, alt_units=None):
 
     def __repr__(self):
         rep = str(self.__class__.__name__)
+        if self.crs:
+            rep += ' ({})'.format(self.crs)
```
mpu-creare: That's a nice addition.

podpac/core/data/datasource.py
mpu-creare: I think this could be a one-liner... which may save some memory... I'll check.

setup.py
```diff
@@ -18,7 +18,7 @@
                     'pint>=0.8',
                     'scipy>=1.0',
                     'traitlets>=4.3',
-                    'xarray>=0.10',
+                    'xarray>=0.11',
```
mpu-creare: Are you just updating the version number, or do we need features from the newest xarray?

setup.py
```diff
@@ -30,7 +30,7 @@
 extras_require = {
     'datatype': [
         'beautifulsoup4>=4.6',
-        'h5py>=2.7',
+        'h5py>=2.9',
```
mpu-creare: Do we need new features from h5py for this to work?

setup.py
jmilloy: In this branch it was upgraded from 0.10 to 0.12 and then downgraded to 0.11. Wouldn't be surprised if the downgrade should have just been back to 0.10.

jmilloy:
I fixed the breaking image test in develop. I don't think this is necessary, but you could pull develop and verify that all tests pass before merging.

mpu-creare:
Thanks for fixing that Jeff!

setup.py
mpu-creare: The answer is yes.

mpu-creare:
Can you post a working example for me? I'm having trouble getting something that works.

mlshapiro:
> 
> 
> Can you post a working example for me? I'm having trouble getting something that works.

https://github.com/creare-com/podpac-examples/blob/develop/notebooks/demos/SMAP-EGI.ipynb

setup.py
mlshapiro: yup, i meant to put that back to 0.10, i'll push it back

podpac/core/data/datasource.py
mlshapiro: sounds good - i was more concerned with being clear, i'll update

podpac/core/coordinates/array_coordinates1d.py
mlshapiro: yes for certain versions of np, it can't compare dates with `np.inf`

podpac/core/data/datasource.py
mlshapiro: fixed in 82bf4d61a80160c1512cba1a59d8e49e50f1b25b

setup.py
mlshapiro: yes!

mlshapiro:
addressed issues, make sure the examples work and then we can merge this in

mlshapiro:
pending test passage

PR #282, opened 2019-07-03
Zarr Datasource and DroughtMonitorCategory datalib

jmilloy:
Zarr sample usage:

```
from podpac.data import Zarr

# local zarr store
node = Zarr(
    source="my/local/path.zarr", datakey="my_data_key", dims=["lat', "lon", "time"])

# custom dimension keys
node = Zarr(
    source="my/local/path", datakey="my_data_key", dims=["lat', "lon"],
    latkey="latitude", "lonkey"="longitude")

# support for cf-encoded times, gets decoded automatically
node = Zarr(
    source="my/local/path", datakey="my_data_key", dims=["lat', "lon", "time"],
    cf_time=True, cf_units=days since 2018-01-01 00:00:00",
    cf_calendar = "proleptic_gregorian")

# s3 zarr store
node = Zarr(
    source="s3://bucket/path.zarr", datakey="my_data_key", dims=["lat', "lon"])

# s3 zarr store with explicit credentials
node = Zarr(
    source="s3://bucket/path.zarr",
    datakey="my_data_key",
    dims=["lat', "lon"],
    access_key_id="MY_KEY_ID",
    secret_access_key="MY_SECRET_KEY",
    region_name="us-east-1")

# customizable dimension loading
class MyZarrSource(Zarr):
    def get_lat(self):
        values = self.group[self.latkey]
        return UniformCoordinates1d(values[0], values[1], values[2])
    def get_lon(self):
        values = self.group[self.lonkey]
        return UniformCoordinates1d(values[0], values[1], values[2])
```

DroughtMonitorCategory sample usage:

```
from podpac.datalib.drought_monitor import DroughtMonitorCategory

d0 = DroughtMonitorCategory(source="my/local/path.zarr", datakey="d0")
```


podpac/core/data/types.py
```diff
@@ -48,6 +49,8 @@
 h5py = lazy_module('h5py')
 boto3 = lazy_module('boto3')
 requests = lazy_module('requests')
+zarr = lazy_module('zarr')
+s3fs = lazy_module('s3fs')
```
jmilloy: @mpu-creare Should I add these to the setup.py under optional imports somewhere?

podpac/core/data/types.py
```diff
@@ -694,7 +697,163 @@ def _find_h5py_keys(obj, keys=[]):
...
+
+        Returns
+        -------
+        time : array-like or Coordinates1d
+            native latitude coordinates.
+        """
+
+        values = self.group[self.timekey]
+        if self.cf_time:
+            # values = cftime.num2date(values, self.cf_units, self.cf_calendar)
```
jmilloy: This works just as well, but requires the `cftime` library. Would you prefer to add that as a requirement for Zarr nodes, or use it if available and fallback on xarray if necessary, or just leverage xarray like this and be done with it?

jmilloy: @mpu-creare The cftime line works just as well as the xarray version, but requires the cftime library. Would you prefer to add that as a requirement for Zarr nodes, or use it if available and fallback on xarray if necessary, or just leverage xarray like this and be done with it?

podpac/datalib/drought_monitor.py
```diff
@@ -0,0 +1,36 @@
+
+from podpac.core.data.types import Zarr
+from podpac.core.coordinates import ArrayCoordinates1d
+
+class DroughtMonitorCategory(Zarr):
```
jmilloy: @mpu-creare Is this enough? Do you need any sort of compositor, too?

podpac/core/data/types.py
jmilloy: Answered this myself.

podpac/core/data/types.py
mpu-creare: Mmm... let's stick with fewer dependencies. ALso, looks like xarray is handling this anyway: 
https://github.com/pydata/xarray/blob/fbe53fe2a35b3eeda63e50f1f38c7cfc92134a0c/xarray/coding/times.py#L38

podpac/datalib/drought_monitor.py
mpu-creare: Nope, This looks great. Thanks!

podpac/core/data/types.py
mpu-creare: I agree with what you did, FWIW.

jmilloy:
Okay, assuming these new tests pass on CI, too, this is all set on my end.

podpac/datalib/drought_monitor.py
jmilloy: @mpu-creare Actually, one more note.

Reall, the DroughtMonitorCategory should implement its own get_time method in order to specify that the time coordinates are ctype 'left' and provide the segment lengths:

```
    def get_time(self):
        values = super(DroughtMonitorCategor, self).get_time()
        segment_lengths = ['%d,d' for val in values[1:]]
        segment_lengths.append('30,d') # for december
        return ArrayCoordinates1d(values, ctype='left', segment_lengths=segment_lengths)
```

We could even partially fix the precision issue here, something more like:

```
    def get_time(self):
        values = super(DroughtMonitorCategor, self).get_time()
        values = [convert_to_month_only(value) for value in values]
        return ArrayCoordinates1d(values, ctype='left', segment_lengths='1,M')
```

In theory, we want to get the June value for any datetime in June (rather than getting the nearest, which would be the July value for the second half of June if using ctype='midpoint' or ctype='point'), and this would be the way to indicate that. I'm not sure how the time interpolation is handling 'nearest' and if it is handling the ctypes correctly, though.

mlshapiro: just an observation:

@jmilloy i like the architecture using the `get_[dim]` in addition to the keys. We should push the `get_[dim]` to other similar DataSource types to allow the datalib to do some more customization of the key-val/tabular formats (CSV, Dataframe, hdf5)

podpac/core/data/types.py
```diff
@@ -694,7 +697,165 @@ def _find_h5py_keys(obj, keys=[]):
         keys = list(set(keys))
         keys.sort()
         return keys
-            
+
+class Zarr(DataSource):
+    source = tl.Unicode(allow_none=True)
```
mpu-creare: should `source` not be tagged? Even if it could be None? 

Why could source be None?

podpac/core/data/types.py
```diff
@@ -694,7 +697,165 @@ def _find_h5py_keys(obj, keys=[]):
...
+        return settings['AWS_SECRET_ACCESS_KEY']
+
+    @tl.default('region_name')
+    def _get_region_name(self):
+        return settings['AWS_REGION_NAME']
+
+    def init(self):
+        # check that source or group is provided
+        if self.source is None:
+            self.group
```
mpu-creare: Does this force `_open_group` to be called?

mpu-creare:
Seems like you can directly specify `group` with this implementation -- but in that case you cannot use this Node as part of a pipeline. Is that true?

podpac/core/data/types.py
mpu-creare: oh wait... it's a datasource node, so `source` is always automatically tagged.

podpac/datalib/drought_monitor.py
mpu-creare: We should probably stage the data better for now...

jmilloy:
Yeah, the source could be `None` if you specify a group directly, but it *has* to be a source to use in a Pipeline. Could remove that and always require the source. I would have no problem with that.

podpac/core/data/types.py
jmilloy: Yes. That's a light operation, and forces the Node to fail early (on init) rather than on eval. Actually, the subsequent checks use `self.group` anyways, so this isn't strictly necessary. I'd have no problem if you prefer to do less checks here and fail later.

podpac/datalib/drought_monitor.py
jmilloy: Do you want me to spend an hour restaging the data without using the cf conventions? That would've probably been easier that handling the cf decoding in the first place.

PR #290, opened 2019-07-11
FIX: use always_xy for lat/lon to handle geodetic to Cartesian transf

mlshapiro:
WIP: need to fix the tests

### DIFF
--- a/podpac/core/coordinates/stacked_coordinates.py
+++ b/podpac/core/coordinates/stacked_coordinates.py
@@ -429,7 +429,7 @@ class StackedCoordinates(BaseCoordinates):
             lat = coords[ilat]
             lon = coords[ilon]
             alt = coords[ialt]
-            tlat, tlon, talt = transformer.transform(lat.coordinates, lon.coordinates, alt.coordinates)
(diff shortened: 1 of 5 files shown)


podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -429,7 +429,7 @@ def _transform(self, transformer):
             lat = coords[ilat]
             lon = coords[ilon]
             alt = coords[ialt]
-            tlat, tlon, talt = transformer.transform(lat.coordinates, lon.coordinates, alt.coordinates)
+            tlon, tlat, talt = transformer.transform(lon.coordinates, lat.coordinates, alt.coordinates)
```
jmilloy: Should this also use `always_xy=True`?

jmilloy:
Looks good and works.

podpac/core/coordinates/stacked_coordinates.py
mlshapiro: no, the Tranformer object is instantiated with `always_xy` so all child `transformer`s will have it.

PR #296, opened 2019-07-18
Feature/drought monitor

mpu-creare:
Lots of bugfixes and a few enhancements needed for the drought monitor.


podpac/core/data/types.py
```diff
@@ -560,8 +560,88 @@ def get_band_numbers(self, key, value):
 
         return matches
 
+class SpecifyCoordinatedMixin(tl.HasTraits):
```
mpu-creare: I'll take suggestions on a better name here.

jmilloy: I wouldn't worry too much, it is fairly internal. You have typo, should be Coordinates not Coordinated?

But I would choose something more like `DatasetCoordinatesMixin` or `LoadCoordinatesMixin`, since you aren't actually specify the coordinates.

mpu-creare:
Pretty sure something went wrong with my previous commit... git thinks that I made a merge commit but I didn't... I'm closing this and fixing on another branch.

PR #300, opened 2019-07-25
Feature/drought monitor merge

mpu-creare:
The drought monitor contains multiple bug fixes and enhancements discovered during implementation of the drought-monitor application. 

ENH: 
* Lambda function more robust to bucket naming
* Making JSON format consistent with javascript (JS uses more compact spacing around separators)
* Handling EGI point queries
* Adding 'substitute_eval_coords' option to coord_select algorithms (substitutes back the original evaluated coordinates)
* Implemented the "to_format" method in UnitsDataArray
* Added handling of pagination to EGI
* Adding public API interface for managers (including aws_lambda manager)
* Adding node wrapper to EGI so that results are cached
* Enabling json output from pipelines
* Checking and filtering SMAP data based on quality flag
* Shortcircuit for AWS Lambda functions in case results for a request already exists
* Added from_url method for Coordinates and Nodes to allow creation from WMS/WCS requests
* Enabled WMS requests through AWS Lambda handler
* Applied Python Black formatting

BUG: 
* Various serialization and EGI bug fixes
* Various fixes to test cases
* A few fixes related to coordinate transforms

See #296 as well.


podpac/core/cache/test/test_cache_stores.py
```diff
@@ -358,7 +358,7 @@ def test_size(self):
         assert store.size == expected_size
 
 
-@pytest.mark.aws
+@pytest.mark.skipif(pytest.config.getoption("--ci"), reason="skip AWS tests during CI")
```
mpu-creare: @jmilloy When I was doing the black merge, I wasn't sure which form was "correct". Please fix this if I got it wrong.

mpu-creare:
@dsully1992 So I got the merge right with the AWS handler and etc. ? Thanks for checking.

dsully1992:
> @dsully1992 So I got the merge right with the AWS handler and etc. ? Thanks for checking.

It looks like it, yes. It's also not likely a big deal, since I'll be making changes to it in this sprint to build a new Lambda function.

podpac/core/cache/test/test_cache_stores.py
jmilloy: Should be the `@pytest.mark.aws` version going forwar. The `pytest.config` global is deprecated. I'll fix that.

mpu-creare:
Thanks for fixing the marking @jmilloy . Other than that, there was nothing specific. Thanks for taking a quick look. Most of these changes weren't structural -- more application specific, so I think we're ok.

dist/aws/handler.py
```diff
@@ -59,6 +59,7 @@ def handler(event, context, get_deps=True, ret_pipeline=False):
     from podpac.core.pipeline import Pipeline
     from podpac.core.coordinates import Coordinates
     from podpac.core.utils import JSONEncoder, _get_query_params_from_url
+    import podpac.datalib
```
dsullivan-creare: I don't see this import being used, what does this do?

dsullivan-creare:
I made one comment on a commit; otherwise, I'm still good with this.

dist/aws/handler.py
mpu-creare: datalib doesn't get imported by default when you import podpac (speed requirement). 

So, when a pipeline needs podpac.datalib.smap, it throws an error saying that datalib doesn't exist. 

This import takes care of that particular error. Should probably add a comment to that effect.

dsullivan-creare: 👍

PR #309, opened 2019-09-12
STALE: Nodes with multiple outputs (stale)

jmilloy:
## Todo

 * [x] rename `"data"` dimension to `"output"`
 * [x] implement the `output` selection in `node_eval` as a fallback
 * [x] CombineOutputs algorithm
 * [ ] Compositor / OrderedCompositor
 * [ ] DatasetCoordinatesMixin
 * [ ] ReprojectedSource
 * [ ] coordinate selection algorithms
 * [ ] signal algorithms
 * [ ] tests
 * [ ] notebook in podpac-examples

## Summary

 * The `outputs` trait is required for Nodes with multiple outputs. It contains the names of the outputs.
 * By default, all outputs are returned as a a single DataArray with an `'output'` dimension.
 * Use `output.sel(output=<output field>)` to select the data for a particular output.
 * Alternatively, the `output` attr can be used to specify a particular output to return on evaluation.

## DataSource with Multiple Outputs

```
import numpy as np
import traitlets as tl
from podpac import Coordinates
from podpac import Node
from podpac.data import Array
from podpac.algorithm import Algorithm, Arange, Min

nc = Coordinates([np.arange(5), np.arange(5)], dims=['lat', 'lon'])

class MyDataSource(Array):
    native_coordinates = nc
    source = np.random.random((5, 5, 2))
    outputs = ['a', 'b']

# node that provides all outputs
node = MyDataSource()
data = node.eval(node.native_coordinates)
print(data)
print(data.sel(output='a')

# node that provides a specific output, like normal
node = MyDataSource(output='a')
a = node.eval(nc)
print(a)
```

In the above, `a` and `data.sel(output='a')` will be the same.

## Custom Algorithm using a multi-output Node

Custom algorithms can be written that use several outputs from a multi-output node. The following algorithm node uses both outputs of the above datasource node, and returns a standard single output.

```
class MyAlgorithm(Algorithm):
    x = tl.Instance(Node)
    y = tl.Instance(MyDataSource)

    def algorithm(self, inputs):
        return inputs['x'] * inputs['y'].sel(output='a') * inputs['y'].sel(output='b')

node = MyAlgorithm(x=Arange(), y=MyDataSource())
result = node.eval(nc)
print(result)
```

## Standard Algorithm using a multi-output Node

Standard algorithms can use multi-output nodes as inputs by selecting a particular output. The following `Min` node uses one output of the above datasource node.

```
node = Min(source=MyDataSource(output='a'), dims='lat')
result = node.eval(nc)
print(result)
```

Note that the `Min` node can use multi-output sources (this won't be the case for all algorithms):

```
node = Min(source=MyDataSource(), dims='lat')
result = node.eval(nc)
print(result)
print(result.sel('output='a'))
```

## Algorithms with Multiple Outputs

The following algorithm uses single-output nodes but returns multiple outputs.

```
class MyAlgorithm2(Algorithm):
    x = tl.Instance(Node)
    y = tl.Instance(Node)
    outputs = ['sum', 'prod', 'diff']

    def algorithm(self, inputs):
        sum_ = inputs['x'] + inputs['y']
        prod = inputs['x'] * inputs['y']
        diff = inputs['x'] - inputs['y']
        a = np.array([sum_, prod, diff])
        a = np.moveaxis(a, 0, -1)
        return a

# node returning all outputs
node = MyMultiAlgorithm(
    x=Arange(),
    y=Array(source=np.random.random(nc.shape), native_coordinates=nc))
result = node.eval(nc)
print(result)
print(result.sel(output='sum'))

# node return a specific output
node = MyMultiAlgorithm(
    x=Arange(),
    y=Array(source=np.random.random(nc.shape), native_coordinates=nc),
    output='sum')

result = node.eval(nc)
print(result)
```


mlshapiro:
> # node that provides all outputs
> node = MyDataSource()
> output = node.eval(node.native_coordinates)
> print(output)
> print(output.sel(data='a')

Just to be clear - `output` is either a `UnitsDataArray` or a `Dataset` of `UnitsDataArray` components?

mpu-creare:
@mlshapiro  and I finally talked. We liked the overall design, just struggling with naming conventions. 

First a few questions using the current naming convention: 
* What happens when a Node doesn't have `outputs` defined? 
* Can you specify a list in `output` -- and will the order be maintained?

We don't really like the names `outputs` and `output` since this confuses how we've used `output` to refer to the results of `node.eval`. 

Here are some options:
* datafield
* channel
* datakey
* field
* key
* data_coord

WE can discuss on Monday.

mpu-creare:
Also, could you implement most of this functionality on the `node_eval` wrapper side? That way you won't have to do any additional work for specific nodes?

jmilloy:
> Just to be clear - output is always a UnitsDataArray with a 'data' dimension.

Yes, that's correct.

> just struggling with naming conventions.

Yes, I tried not to get bogged down by the names but I do agree. I like 'fields' and 'field'.

> What happens when a Node doesn't have outputs defined?

Evaluation will return a DataArray without a 'data' dimension, like it does currently. If you try to specify a `field` for a Node with no output fields defined, a TraitsError is raised.

> Can you specify a list in output -- and will the order be maintained?

No, you can either get all of the data fields, or one of them.

Is there a significant need for this? A node that returns one data field (without a data dimension) is useful because it can be used directly as an input in existing Algorithms and Compositors with no surprises, but a node that returns some might as well just return all because the user will have deal with it in a particular way anyways.

> how do I know which dimension corresponds to the coordinates vs. the data coordinate?

The `DatasetCoordinatesMixin` will need to be updated to specify the keys for the output data fields. (Possibly we should rename it to `DatasetFieldsMixin`.)

mpu-creare:
Based on our discussion let's go with:
* outputs as the name of the "fields"
* output as the name of the particular "field" that will be returned
* output as the dimension name in the `UnitDataArray`.

jmilloy:
```
def algorithm(self, inputs):
        return inputs['x'] * inputs['y'].sel(output='a') * inputs['y'].sel(output='b')
```

@mpu-creare I'm sorry to do this, but this example came up. I wanted to give you one more chance to decide that this is what you wanted, because `inputs['y'].sel(output='a')` throws me a bit. A simple "Looks fine!" and we just move on. Otherwise, compare to `inputs['y'].sel(field='a')` or similar.

podpac/core/algorithm/stats.py
```diff
@@ -45,6 +45,10 @@ class Reduce(Algorithm):
     _reduced_coordinates = tl.Instance(Coordinates, allow_none=True)
     _dims = tl.List(trait_type=str)
 
+    @tl.default("outputs")
```
jmilloy: @mpu-creare This simple update that sets the `outputs` using the `source.outputs` allows the reduce nodes (like Min and Median) to handle nodes with multiple outputs. Should I spend some more time on the other Algorithm nodes updating them as well?

 * I think the Convolution will be the same, so it's probably worth it
 * I think the coordinate selection nodes might be just as easy, as well
 * The Arithmetic node could be a bit more work

I guess the same thing applies to DataSource types and the Compositors. Should I be proactive updating those or do you want to do them as needed?
 * I think I should do OrderedCompositor (or do you want to require that OrderedCompositor sources are always single-output?)
 * I think the ReprojectedSource node will be easy, like the Reduce nodes here
 * We already discussed doing the DatasetCoordinatesMixin, which will cover
    - H5PY
    - Zarr
 * That leaves
    - PyDap
    - CSV
    - Rasterio
    - WCS

mpu-creare:
"Looks fine!"

I think it's worth adding this functionality to all the nodes. 
I specifically need it in an algorithm node... not sure it's possible in an Arithmetic node...

jmilloy:
Okay, I agree that it doesn't make very much sense for Arithmetic nodes.

What about Compositors? Which use-case do we need?
 - compositor sources can only be normal single-output nodes
 - compositor sources must all have the same outputs (and can  composite multiple outputs at once)
 - compositors sources can be anything, and the outputs of the compositor is the union of the outputs of its sources

mpu-creare:
I'm waffling between 2 and 3. 

* The first one is too restrictive -- people will want to composite RGB images together
* The second one seems like the right amount of restrictiveness -- compositors should be making similar data sources look the same
* The last one just scares me... a little... it basically allows compositors to ALSO composite along the output dimension...
    * You can image compositing two sources with no shared outputs... but maybe you want to do this to create an alpha channel for an RGB image
    * it also allows you to composite an RGB source with a RGBIR source (i.e. more than 3 channels)
   

I think go with 3.

jmilloy:
> I'm waffling between 2 and 3.

We could make a separate class `OrderedOutputCompositor`, so that 2 and 3 are actually different classes. Or a flag. I'll keep that in mind while implementing and see if it actually makes sense or not.

mpu-creare:
A separate class is not a bad idea... just try to anticipate what a naiive user would do.

jmilloy:
Another note. I think we'll need a simple convenience node that combines single-output sources into a multi-output node and names them.

```
r_node = MyDataSource()
b_node = MySecondDataSource()
g_node = MyThirdDataSource()
rgb_node = CombineOutputs(r=r_node, b=b_node, g=g_node)

# or perhaps
rgb_node = CombineOutputs(sources=[r_node, b_node, g_node], outputs=['r', 'g', 'b'])
```

Basically, the output compositor *almost* covers this case. Standard single-output datasources don't have an output dimension to composite, so the following would work:

```
r_node = MyDataSource(outputs=['r'])
b_node = MySecondDataSource(outputs=['b'])
g_node = MyThirdDataSource(outputs=['g'])
rgb_node = OrderedCompositor(sources=[r_node, b_node, g_node])
```

except that `MyDataSource(outputs=['r'])` will usually fail.

`CombineOutputs` is a very straightforward Algorithm node or Compositor node, depending on which spec we prefer (`keyword` args or `sources`).

mpu-creare:
`CombineOutputs` seems to assume that 'r_node' etc. all have the same CRS, and grid coverage?

jmilloy:
CRS - I don't think so, each source node will be evaluated at the requested coordinates, and if they are datasources they will be transformed to and from the native coordinates as needed.

grid coverage - No, they will just be evaluated at the requested coordinates, and return data with NaNs as appropriate.

```
r_node = MyDataSource()
b_node = MySecondDataSource()
g_node = MyThirdDataSource()
rgb_node = CombineOutputs(r=r_node, b=b_node, g=g_node)

r = r_node.eval(c)
g = r_node.eval(c)
b = r_node.eval(c)
rgb = rgb_node.eval(c)

xr.assert_equal(rgb['lat'], r['lat']) # etc
xr.assert_equal(rgb.sel(output='r'), r) # etc
```

Another way to put is that CombineOutputs is not a DataSource, so it does not have native_coordinates (and thus no native crs and no "grid coverage"). I guess CombineOutputs should be an Algorithm (because it shouldn't have shared_coordinates or source_coordinates either).

```
class CombineOutputs(Algorithm):
    @tl.default('outputs')
    def _default_outputs(self):
        return list(self._inputs.keys)

    def algorithm(self, inputs):
        data = np.array([inputs[key] for key in self.outputs])
        data = np.moveaxis(a, 0, -1)
        return data
```

This is a lot of detail, but that's because I wonder if I'm missing something.

mpu-creare:
You're right Jeff! I should have realized that... I was replying too quickly. 

I think `CombineOutputs` is a fantastic Node that would solve a lot of problem. 

Incidentally, if you get the `CombineOutputs(r=r_node, b=b_node, ...)` interface working, I would love to have a similar interface for `Arithmetic` so we don't have to name all the variables `A`, `B`, `C`, ... at the time of writing that Node, we didn't have the serialization/deserialization piece working well enough to make it work...

jmilloy:
I started this again over at #348 due to significant deviations from develop.

PR #313, opened 2019-09-27
Readonly source, native_coordinates (and dataset) traits

jmilloy:
* I set `source`, `native_coordinates` and `dataset` traits as read_only.
 * Traitlets does not allow read_only traits to be set at initialization, so I added a simple workaround in the DataSource `_first_init`.
 * I removed code handling source changes.
 * Note: `set_trait` still works for read_only traits in a pinch.
 * A general solution to making tagged node attrs read-only coming soon.

This is an issue we've been punting on for a while. I am unifying some of the "dataset" based datasources for the multiple outputs feature, and it is a good time to do this.

### DIFF
--- a/podpac/core/data/types.py
+++ b/podpac/core/data/types.py
@@ -74,7 +74,7 @@ class Array(DataSource):
     `native_coordinates` need to supplied by the user when instantiating this node.
     """
 
-    source = ArrayTrait()
+    source = ArrayTrait(read_only=True)
 
     @tl.validate("source")
     def _validate_source(self, d):
@@ -123,16 +123,17 @@ class PyDAP(DataSource):
         auth_session instead if you have security concerns.
     """
 
-    # required inputs
-    source = tl.Unicode(default_value="")
+    source = tl.Unicode(read_only=True)
+    dataset = tl.Instance("pydap.model.DatasetType", read_only=True)
+
+    # node attrs
     datakey = tl.Unicode().tag(attr=True)
 
-    # optional inputs and later defined traits
-    auth_session = tl.Instance(authentication.Session, allow_none=True)
+    # optional inputs
     auth_class = tl.Type(authentication.Session)
-    username = tl.Unicode(None, allow_none=True)
-    password = tl.Unicode(None, allow_none=True)
-    dataset = tl.Instance("pydap.model.DatasetType")
+    auth_session = tl.Instance(authentication.Session, allow_none=True)
+    username = tl.Unicode(default_value=None, allow_none=True)
+    password = tl.Unicode(default_value=None, allow_none=True)
 
     @tl.default("auth_session")
     def _auth_session_default(self):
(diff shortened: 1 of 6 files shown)


jmilloy:
@mpu-creare @mlshapiro I would love a quick peek at this if you can get a chance soon. I don't feel like we need a thorough review, but just to confirm that this does't break your expectations for how datasources should be created and used.

podpac/core/data/types.py
```diff
@@ -672,13 +637,15 @@ class H5PY(DatasetCoordinatedMixin, DataSource):
         Default is 'r'. The mode used to open the HDF5 file. Options are r, r+, w, w- or x, a (see h5py.File).
     """
 
-    source = tl.Unicode()
-    dataset = tl.Any(allow_none=True)
+    source = tl.Unicode(read_only=True)
+    dataset = tl.Any(read_only=True)
+    file_mode = tl.Unicode(default_value="r", read_only=True)
```
mpu-creare: Do you need a set_trait in the init for this as well?

mpu-creare:
A general solution would be good... I can see this getting out of hand.

jmilloy:
I think this is the best I can do:

```
clas Node(tl.HasTraits):
    ....
    def __init__(self, **kwargs):
        """ Do not overwrite me """
        # set "readonly" traits to read_only
        for name, trait in self.traits().items():
            if trait.metadata.get('readonly'):
                trait.read_only = True
        
        tkwargs = self._first_init(**kwargs)

        # set "readonly" traits using set_trait
        for name, trait in self.traits().items():
            if trait.metadata.get('readonly') and name in tkwargs:
                self.set_trait(name, tkwargs.pop(name))

        # Call traitlest constructor
        super(Node, self).__init__(**tkwargs)
        self.init()
```

Usage:

```
class DataSource(Node):
    source = tl.Any().tag(readonly=True)
    ....
```

jmilloy:
We could extend this to include node attrs:

```
class Node(tl.HasTraits):
    def __init__(self, **kwargs):
        """ Do not overwrite me """
        # set "readonly" and "attr" traits to read_only
        for name, trait in self.traits().items():
            if trait.metadata.get('readonly') or trait.metadata.get('attr'):
                trait.read_only = True
        
        tkwargs = self._first_init(**kwargs)

        # set "readonly" traits using set_trait
        for name, trait in self.traits().items():
            if (trait.metadata.get('readonly') or trait.metadata.get('attr')) and name in tkwargs:
                self.set_trait(name, tkwargs.pop(name))

        # Call traitlest constructor
        super(Node, self).__init__(**tkwargs)
        self.init()
```

Usage:

```
class MyDataSource(DataSource):
    source = tl.Unicode().tag(readonly=True)
    my_attr = tl.Int().tag(attr=True)
```

Which is an improvement over

```
class MyDataSource(DataSource):
    source = tl.Unicode().tag(readonly=True)
    my_attr1 = tl.Int().tag(attr=True, readonly=True)
    my_attr2 = tl.Int().tag(attr=True) # < ----------------------------- BUG!!!!
```

podpac/core/data/types.py
jmilloy: Hmmm. I'm surprised I made `file_mode` read_only, I don't think I meant to.

jmilloy:
Proposed alternative at #315

PR #314, opened 2019-09-27
Tests should run without warnings.

jmilloy:
On my machine, our tests pass with 700+ warnings.

In this PR, warnings are treated as errors by pytest. All warnings are fixed by globally filtering certain warnings, catching expected warnings, or fixing bugs in podpac.

Bugfixes:
 - (core/algorithm/stats.py) multi-indexes should be tuples, not lists
 - (core/coordinates/uniform_coordinates1d.py) `np.timedelta64` cannot be compared directly with 0
 - (core/coordinates/uniform_coordinates1d.py) uniform coordinates *do* need to check `start`, `stop`, and `step`
 - (core/data/interpolation.py) numpy arrays cannot be compared with `==`
 - (core/data/interpolator.py) `spatial_tolerance` should be defined as a trait for the base interpolator

### DIFF
--- a/podpac/core/algorithm/stats.py
+++ b/podpac/core/algorithm/stats.py
@@ -795,7 +795,7 @@ class Reduce2(Reduce):
 
         y = xr.full_like(output, np.nan)
         for x, xslices in xs:
-            yslc = [xslices[x.dims.index(dim)] for dim in self._reduced_coordinates.dims]
+            yslc = tuple(xslices[x.dims.index(dim)] for dim in self._reduced_coordinates.dims)
             y.data[yslc] = self.reduce(x)
         return y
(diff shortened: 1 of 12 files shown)


jmilloy:
@mlshapiro @mpu-creare Any idea what's going on with these builds? We are getting a segfault in some places and other strange errors that I don't see locally and that don't relate to the fairly minor changes in these commits.

jmilloy:
I don't understand, but I reverted the `error` filterwarning and that sorted a lot out.

jmilloy:
Okay, this is a big improvement, with 14 instances of the pyproj +init deprecation warning #316, which I actually decided to leave for now.

PR #315, opened 2019-09-27
general readonly tag

jmilloy:
Node traits tagged as 'readonly' or as 'attr' can be specified using keyword arguments during Node instatiation, but cannot be later modified. It uses the `read_only` flag under the hood but is a slightly more flexible.

```
class MyDataSource(DataSource)
    source = tl.Unicode().tag(readonly=True)
    my_attr = tl.Int().tag(attr=True)
```

```
>>> node = MyDataSource(source='mysource', my_attr=8)
>>> node.source = 'other'
...
traitlets.traitlets.TraitError: The "source" trait is read-only.
>>> node.my_attr = 6
traitlets.traitlets.TraitError: The "my_attr" trait is read-only.
```

### DIFF
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -139,7 +139,17 @@ class Node(tl.HasTraits):
 
     def __init__(self, **kwargs):
         """ Do not overwrite me """
+        # make tagged "readonly" and "attr" traits read_only
+        for name, trait in self.traits().items():
+            if trait.metadata.get("readonly") or trait.metadata.get("attr"):
+                trait.read_only = True
+
         tkwargs = self._first_init(**kwargs)
+
+        # set tagged "readonly" and "attr" traits using set_trait
+        for name, trait in self.traits().items():
+            if (trait.metadata.get("readonly") or trait.metadata.get("attr")) and name in tkwargs:
+                self.set_trait(name, tkwargs.pop(name))
 
         # Call traitlest constructor
         super(Node, self).__init__(**tkwargs)
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -139,17 +139,17 @@ class Node(tl.HasTraits):
 
     def __init__(self, **kwargs):
         """ Do not overwrite me """
-        # make tagged "readonly" and "attr" traits read_only
-        for name, trait in self.traits().items():
-            if trait.metadata.get("readonly") or trait.metadata.get("attr"):
-                trait.read_only = True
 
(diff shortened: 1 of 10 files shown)


jmilloy:
@mpu-creare  Try this on for size. Replaces #313.

podpac/core/node.py
```diff
@@ -139,8 +139,18 @@ def _validate_units(self, d):
...
+                trait.read_only = True
+
         tkwargs = self._first_init(**kwargs)
 
+        # set tagged "readonly" and "attr" traits using set_trait
+        for name, trait in self.traits().items():
+            if (trait.metadata.get("readonly") or trait.metadata.get("attr")) and name in tkwargs:
+                self.set_trait(name, tkwargs.pop(name))
+
         # Call traitlest constructor
```
mpu-creare: Simple implementation suggestion: 

```python
In [3]: class T(tl.HasTraits):
   ...:     a = tl.Int().tag(readonly=True)
   ...:     b = tl.Int()
   ...:     def __init__(self, **kwargs):
   ...:         super().__init__(**kwargs)
   ...:         for name, trait in self.traits().items():
   ...:             if (trait.metadata.get("readonly") and name in kwargs):
   ...:                 trait.read_only = True
   ...:

In [4]: t = T(a=1, b=2)

In [5]: t.a=3
---------------------------------------------------------------------------
TraitError                                Traceback (most recent call last)
<ipython-input-5-012fe921ffc2> in <module>
----> 1 t.a=3

C:\Anaconda3\envs\podpac\lib\site-packages\traitlets\traitlets.py in __set__(self, obj, value)
    581         """
    582         if self.read_only:
--> 583             raise TraitError('The "%s" trait is read-only.' % self.name)
    584         else:
    585             self.set(obj, value)

TraitError: The "a" trait is read-only.

In [6]: t.b=2
```

mpu-creare:
Looks good to me. I made a suggestion for a slightly simpler implementation, but you might have had a reason for what you did.

podpac/core/node.py
jmilloy: Ah, yeah, the issue is that the read_only flag gets set on the *class* trait, so with this implementation it doesn't work the second time:

```
>>> t1 = T(a=1)
>>> t2 = T(a=1)
TraitError: The "a" trait is read-only
```

mpu-creare: I figured there was a reason! Wow, that's crazy. A comment in the code might be helpful...

PR #323, opened 2019-10-30
ENH: Adding a generic algorithm that can execute arbitrary multi-line…

mpu-creare:
```python
import podpac
n1 = podpac.algorithm.Arange()
n2 = podpac.algorithm.SinCoords()
coords = podpac.Coordinates([podpac.clinspace(4800000, 4900000, 256), podpac.clinspace(-8500000, -8400000, 256)],
                     dims=['lat', 'lon'], crs='epsg:3857')
n = podpac.core.algorithm.algorithm.Generic(code='''import numpy as np
output = np.minimum(n1, n2)''', inputs={'n1':n1, 'n2': n2})
n.eval(coords)
```

Note, I've made this "secure", so you will need to set a setting when testing this: 
```python
podpac.settings.set_allow_python_eval_exec(True)
```

This doesn't use the usual 'settings' mechanism, and REQUIRES a file to exist in one of 3 locations (cwd, root podpac install, or ROOT_PATH as defined in the settings file.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,6 +28,7 @@ install:
 
   # install setup.py and dev extras
   - pip install .[devall]
+  - touch podpac/ALLOW_PYTHON_EVAL_EXEC
 
 # cache pip dependencies for faster builds
 cache: pip
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,6 +30,7 @@ install:
   - pip install .[devall]
   
   # Allow Python exec and eval functions for unit tests
+  - sudo mkdir /home/travis/.podpac
   - sudo touch /home/travis/.podpac/ALLOW_PYTHON_EVAL_EXEC
   
 # cache pip dependencies for faster builds
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,8 +30,8 @@ install:
   - pip install .[devall]
   
   # Allow Python exec and eval functions for unit tests
-  - sudo touch podpac/ALLOW_PYTHON_EVAL_EXEC
-
+  - sudo touch /home/travis/.podpac/ALLOW_PYTHON_EVAL_EXEC
+  
 # cache pip dependencies for faster builds
 cache: pip
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -28,7 +28,9 @@ install:
 
   # install setup.py and dev extras
   - pip install .[devall]
-  - touch podpac/ALLOW_PYTHON_EVAL_EXEC
+  
+  # Allow Python exec and eval functions for unit tests
+  - sudo touch podpac/ALLOW_PYTHON_EVAL_EXEC
 
 # cache pip dependencies for faster builds
 cache: pip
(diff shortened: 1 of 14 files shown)


jmilloy:
The PermissionError is raised at instantiation, and I wonder if it should be raised at evaluation. Perhaps there should be a warning an instantation.

I came across this because I just wanted to test the serialization, not evaluate, so I didn't think I should need to actually change the setting. Similarly, might I want to create the node, serialize it, and then use that to execute somewhere else, without ever evaluating it on my local machine and without changing the setting. It would also be fine to create the node, then change the setting, and then evaluate.

What do you think?

mpu-creare:
You make a good point. I just didn't want people to get blind-sided by this, so I figured fail early. 

Let's do warning on instantiation, and PermissionError on eval. Feel free to fix if you get to it first.

jmilloy:
Would you rather use keyword arguments `Generic(a=a_node, b=b_node)` instead of a dictionary `Generic(inputs={'a': a_node, 'b': b_node})`? I think I might be able to do that.

mpu-creare:
Yes! That would be great! I suppose that if you catch it on init you could put it into an inputs dictionary internally... nicer interface for users. I should have thought of that.

jmilloy:
Yeah, that's what I'm thinking, though it's a bit tricky because we want to pass by some keyword arguments to traitlets.

mpu-creare:
I wonder if that's too dangerous then... what if I user happens to overwrite, say cache_ctrl, just because they don't know it's in the base class... or I suppose the tricky part comes from checking for cases like that... well, I guess if any kwargs is not in trait names then you're ok. Otherwise, throw an error? 

Sorry for the rambling.

jmilloy:
Let me show you what I have in a minute. Wrapped up in some other changes, unfortunately.

mpu-creare:
Sure. Also happy to do a screenshare if that helps.

jmilloy:
```
class GenericInputs(Algorithm):
    """Base class for Algorithms that accept generic named inputs."""

    inputs = tl.Dict()

    @property
    def _inputs(self):
        return self.inputs

    def _first_init(self, **kwargs):
        trait_names = self.trait_names()
        input_keys = [key for key in kwargs if key not in trait_names]
        inputs = {key: kwargs.pop(key) for key in input_keys}
        return super(GenericInputs, self)._first_init(inputs=inputs, **kwargs)
```

They'll get a TraitsError if they try to use another trait as a node name.

The Arithmetic node and the Generic node both use this base class. So the Arithmetic is now freed up, too.

mpu-creare:
I would add: 
```python
...
input_keys = [key for key in kwargs if key not in trait_names 
                       and isinstance(kwargs[key], Node)]  # add this part
```
so that they don't accidentally miss-spell an existing trait (.e.g. `cach_ctlr='ram'` would get caught with the extra check)

The `TraitError` might be confusing, (.e.g. `cache_ctrl=Arange()` leads to a TraitError that it should be an instance of CacheCtrl instead of Node -- it doesn't tell the user that they need to use a different name for their inputs`. 

 so we should add a special note to the doc string... or catch the `TriatError` and add an additional message.

jmilloy:
Okay, I pushed these changes to develop.

PR #325, opened 2019-10-31
ENH: Adding node to apply masks.

mpu-creare:
Seems like there are a number of places where masking data is helpful. This implements a relatively simple `Mask` node to cover those cases. 

This was, in particular, needed for the DroughtMonitor application so that we can entirely avoid the generic and arithmetic algorithm nodes.


podpac/core/algorithm/algorithm.py
```diff
@@ -361,3 +361,72 @@ def algorithm(self, inputs):
...
+    """ Masks the `source` based on a boolean expression involving the `mask`.
+    
+    Attributes
+    ----------
+    source : podpac.Node
+        The source that will be masked
+    mask : podpac.Node
+        The data that will be used to compute the mask
+    masked_val : float, optional
+        Default value is np.nan. The value that will replace the maked items.
```
mlshapiro:
```suggestion
        Default value is np.nan. The value that will replace the masked items.
```

podpac/core/algorithm/algorithm.py
```diff
@@ -361,3 +361,72 @@ def algorithm(self, inputs):
     @property
     def _inputs(self):
         return self.inputs
+
+
+class Mask(Algorithm):
+    """ Masks the `source` based on a boolean expression involving the `mask`.
```
mlshapiro:
```suggestion
    """ Masks the `source` based on a boolean expression involving the `mask` (i.e. source[mask <bool_op> <bool_val> ] = <masked_val>). For a normal boolean mask input, default values for `bool_op`, `bool_val` and `masked_val` can be used.
```

podpac/core/algorithm/algorithm.py
```diff
@@ -361,3 +361,72 @@ def algorithm(self, inputs):
...
+    masked = Mask(source=a, mask=b,
+                  masked_val=np.nan, 
+                  bool_val=0, bool_op='>'
+                  in_place=True)
+    
+    """
+
+    source = NodeTrait()
+    mask = NodeTrait()
+    masked_val = tl.Float(np.nan).tag(attr=True, readonly=True)
```
mlshapiro: note `readonly` unneccessary

podpac/core/algorithm/algorithm.py
```diff
@@ -361,3 +361,72 @@ def algorithm(self, inputs):
...
+        Default value is 1. The value used to compare the mask when creating the boolean expression
+    bool_op : enum, optional
+        Default value is '=='. One of ['==', '<', '<=', '>', '>=']
+    in_place : bool, optional
+        Default is False. If True, the source array will be changed in placem which could affect the value of the source 
+        in other parts of the pipeline. 
+        
+    Examples
+    ----------
+    # Create a node that make the following substitution "a[b > 0] = np.nan"
```
mlshapiro: Add simple example for naive mask first

PR #329, opened 2019-11-04
Creates a settings context manager for easy temporary settings.

jmilloy:
Pretty simple, the settings can be used in a context manager to restore original settings after exiting a block.

```
with podpac.settings:
    podpac.settings["CACHE_OUTPUT_DEFAUL"] = False
    ...

# settings are restored
assert podpac.settings["CACHE_OUTPUT_DEFAULT"] == True
```

This can replace setup_method and teardown_method in tests, which is what I used it for here, but it seems useful in general. If you like it, I will propagate it to other places in the tests.

Note that just setting and restoring the settings in a test method (instead of setup and teardown methods) isn't a good idea because if the test fails, the settings won't be restored. Other alternatives I considered are a decorator `@podpac.settings.temporary_settings` that would be applied to specific test methods (a bit harder to implement) and a base test class with setup and teardown methods that always restore all of the settings (inefficient).

### DIFF
--- a/podpac/core/algorithm/generic.py
+++ b/podpac/core/algorithm/generic.py
@@ -33,6 +33,9 @@ class GenericInputs(Algorithm):
 
     def _first_init(self, **kwargs):
         trait_names = self.trait_names()
+        for key in kwargs:
+            if key in trait_names and isinstance(kwargs[key], Node):
+                raise RuntimeError("Trait '%s' is reserved and cannot be used as an Generic Algorithm input" % key)
         input_keys = [key for key in kwargs if key not in trait_names and isinstance(kwargs[key], Node)]
         inputs = {key: kwargs.pop(key) for key in input_keys}
         return super(GenericInputs, self)._first_init(inputs=inputs, **kwargs)
--- a/podpac/core/algorithm/test/test_generic.py
+++ b/podpac/core/algorithm/test/test_generic.py
@@ -17,6 +17,10 @@ class TestGenericInputs(object):
         assert "a" in node.inputs
         assert "b" in node.inputs
 
+    def test_reserved_name(self):
+        with pytest.raises(RuntimeError, match="Trait .* is reserved"):
+            GenericInputs(style=SinCoords())
+
     def test_serialization(self):
         node = GenericInputs(a=Arange(), b=SinCoords())
         d = node.definition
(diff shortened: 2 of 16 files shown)


podpac/core/algorithm/generic.py
```diff
@@ -33,6 +33,9 @@ def _inputs(self):
 
     def _first_init(self, **kwargs):
         trait_names = self.trait_names()
+        for key in kwargs:
+            if key in trait_names and isinstance(kwargs[key], Node):
+                raise RuntimeError("Trait '%s' is reserved and cannot be used as an Generic Algorithm input" % key)
```
jmilloy: Oops, this particular check is supposed to be in develop already. If we decide we don't like this idea I'll make sure to commit it and its to develop anyways.

podpac/core/settings.py
```diff
@@ -331,6 +331,13 @@ def set_allow_python_eval_exec(self, allow=False):
                     pass
             self["ALLOW_PYTHON_EVAL_EXEC"] = False
 
+    def __enter__(self):
+        self.original = {k: v for k, v in self.items()}
```
mpu-creare: Do you want to make original private?

podpac/core/algorithm/test/test_generic.py
```diff
@@ -28,80 +32,89 @@ def test_serialization(self):
...
-
-        podpac.settings.set_allow_python_eval_exec(False)
-        with pytest.warns(UserWarning, match="Insecure evaluation"):
+        with podpac.settings:
+            podpac.settings.set_allow_python_eval_exec(True)
             node = Generic(code="import numpy as np\noutput = np.minimum(a,b)", a=a, b=b)
 
-    def test_evaluate(self):
-        podpac.settings.set_allow_python_eval_exec(True)
+            podpac.settings.set_allow_python_eval_exec(False)
```
mpu-creare: This actually creates a file... so probably better to set the settings['ALLOW_PYTHON_EVAL_EXEC']=True directly, based on how you wrote the context manager.

mlshapiro:
this is a great idea, thanks @jmilloy

jmilloy:
@mpu-creare I re-requested review because there are a lot of small changes here. Using the settings context manager and also tracking down some remaining warnings during tests was fruitful - I found and fixed a few bugs that are included here.

podpac/core/algorithm/test/test_generic.py
```diff
@@ -28,80 +32,102 @@ def test_serialization(self):
...
-        for key in podpac.settings:
-            podpac.settings[key] = self.settings_orig[key]
+        with podpac.settings:
+            podpac.settings.set_allow_python_eval_exec(True)
+            with warnings.catch_warnings():
+                warnings.filterwarnings("error", "Insecure evaluation.*")
+                node = Arithmetic(A=sine_node, B=sine_node, eqn="2*abs(A) - B + {offset}", params={"offset": 1})
+
+            podpac.settings.set_allow_python_eval_exec(False)
+            with pytest.warns(UserWarning, match="Insecure evaluation"):
```
mpu-creare: does this need a .*? 
```suggestion
            with pytest.warns(UserWarning, match="Insecure evaluation.*"):
```

podpac/core/algorithm/test/test_generic.py
```diff
@@ -28,80 +32,102 @@ def test_serialization(self):
...
-        self.settings_orig = copy.deepcopy(podpac.settings)
+    def test_init(self):
+        sine_node = SinCoords()
 
-    def teardown_method(self):
-        for key in podpac.settings:
-            podpac.settings[key] = self.settings_orig[key]
+        with podpac.settings:
+            podpac.settings.set_allow_python_eval_exec(True)
+            with warnings.catch_warnings():
```
mpu-creare: if the secure eval is True, why would there be a warning?

podpac/core/algorithm/test/test_generic.py
```diff
@@ -28,80 +32,102 @@ def test_serialization(self):
...
-        coords = podpac.Coordinates([podpac.crange(-90, 90, 1.0), podpac.crange(-180, 180, 1.0)], dims=["lat", "lon"])
-        sine_node = SinCoords()
+            coords = podpac.Coordinates(
+                [podpac.crange(-90, 90, 1.0), podpac.crange(-180, 180, 1.0)], dims=["lat", "lon"]
+            )
+            sine_node = SinCoords()
 
-        with pytest.warns(UserWarning, match="Insecure evaluation"):
-            node = Arithmetic(A=sine_node, B=sine_node, eqn="2*abs(A) - B + {offset}", params={"offset": 1})
+            with pytest.warns(UserWarning, match="Insecure evaluation"):
```
mpu-creare: again, do you need the '.*'?

podpac/core/algorithm/test/test_generic.py
```diff
@@ -28,80 +32,102 @@ def test_serialization(self):
...
-    def test_init_(self):
+    def test_init(self):
         a = SinCoords()
         b = Arange()
 
-        podpac.settings.set_allow_python_eval_exec(True)
-        node = Generic(code="import numpy as np\noutput = np.minimum(a,b)", a=a, b=b)
+        with podpac.settings:
+            podpac.settings.set_allow_python_eval_exec(True)
+            with warnings.catch_warnings():
```
mpu-creare: IF true, why would there be a warning?

podpac/core/settings.py
```diff
@@ -331,6 +331,18 @@ def set_allow_python_eval_exec(self, allow=False):
                     pass
             self["ALLOW_PYTHON_EVAL_EXEC"] = False
 
+    def __enter__(self):
+        # save original settings
+        self._original = {k: v for k, v in self.items()}
+
+    def __exit__(self, type, value, traceback):
+        # restore eval/exec state (setting and file)
+        self.set_allow_python_eval_exec(self._original.get("ALLOW_PYTHON_EVAL_EXEC", False))
```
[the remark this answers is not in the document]
jmilloy: Okay, I did this because I didn't see it in the defaults or something like that. I think you are right.

PR #334, opened 2019-11-11
ENH: adds "UnitsDataArray.open" and "UnitsDataArray.create" class methods

mlshapiro:
- Adds `UnitsDataArray.open` class method
- Changes `create_data_array` to `create_dataarray` to be consistent with xarray naming convention in methods.
- Deprecated `create_dataarray` in favor of `UnitsDataArray.create`

This feature adds the ability to symmetrically go between UnitsDataArray -> netcdf -> UnitsDataArray.  The goal is move away from pickle in the lambda function interface and more generally, provide users a way to export/import binary outputs from/to PODPAC.

~~Both `creare_dataarray` and `open_dataarray` could probably be classmethods, like we have done in the Coordinates module. It may also be useful to `export` and `import` methods on `UnitsDataArray` that uses the `to_netcdf` and `open_datarray` in the background.~~
Done in e9abbc485ba15ffd36d578fa590f5a63d9c7e808

## Example

```python
# Eval with simple example
from podpac.data import Array
from podpac import Coordinates, UnitsDataArray
import numpy as np

# mock data
data = np.random.rand(5, 5)

# create native coordinates for data
lat = np.linspace(-10, 10, 5)
lon = np.linspace(-10, 10, 5)
native_coords = Coordinates([lat, lon], ['lat', 'lon'])

# create Array DataSource node
node= Array(source=data, native_coordinates=native_coords)

# evaluate
output = node.eval(node_array.native_coordinates)

# export
output_netcdf = output.to_netcdf()

# import
output2 = UnitsDataArray.open(output_netcdf)

# goal:
# output2 == output

```

### DIFF
--- a/podpac/__init__.py
+++ b/podpac/__init__.py
@@ -44,7 +44,7 @@ from podpac.core.coordinates import Coordinates, crange, clinspace
 from podpac.core.node import Node, NodeException
 import podpac.core.authentication as authentication
 from podpac.core.utils import NodeTrait
-from podpac.core.units import ureg as units
+from podpac.core.units import ureg as units, open_dataarray, UnitsDataArray
 
 # Organized submodules
 # These files are simply wrappers to create a curated namespace of podpac modules
--- a/podpac/__init__.py
+++ b/podpac/__init__.py
@@ -44,7 +44,7 @@ from podpac.core.coordinates import Coordinates, crange, clinspace
 from podpac.core.node import Node, NodeException
 import podpac.core.authentication as authentication
 from podpac.core.utils import NodeTrait
-from podpac.core.units import ureg as units, open_dataarray, UnitsDataArray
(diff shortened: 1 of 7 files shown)


mpu-creare:
Also, looks like the build is failing, which may explain the reduced coverage https://travis-ci.org/creare-com/podpac/builds/610568191

mlshapiro:
> I would prefer the functions to be a class method on UnitsDataArray

I would prefer this too. I'll make this change and review before merging

mlshapiro:
Updated PR context above

mpu-creare:
There was something nagging at the back of my mind. We did serialize/deserialize already in the cache: see here: https://github.com/creare-com/podpac/blob/develop/podpac/core/cache/file_cache_store.py#L147 

I'm guessing those lines will need some (minor) updates as well

PR #336, opened 2019-11-12
Feature/aws improvements

mlshapiro:
- Lambda node by default uses "invoke" instead of S3. Includes handler improvements to consolidate the "pipeline" data model that is passed among the various triggers.
- Data is passed back and forth via netcdf.
- Errors are returned via a LambdaException with stack trace (this is awesome).
- Lambda function can be restricted to only allow certain Nodes using `function_restrict_pipelines`. This uses an environment variable (`PODPAC_RESTRICT_PIPELINES`) in the lambda function to store the hashes of acceptable Nodes. The handler checks to see if the node hash is in the environment variable and raises a ValueError if not.

API Gateway and S3 triggers still needs testing. Handler may need to be updated for these as well.

In the future, I think we should deprecate the "eval_s3" and "eval_api" methods on the Lambda node since these are no longer relevant. We should keep the ability to add extra triggers for S3 and APIGateway since there are other use cases that need these. 

To Do:

- Parse the `response["LogResult"]` to get the "REPORT" which contains billing information
  - Perhaps we want to skip this, seems like it would be complicated to keep track of every node's billing information. Probably want to add a boto3 integration separately for this
- Test the APIGateway examples using the `requests` library. 
  - [ ] Make sure the WMS urls are getting parsed correctly.
  - [ ] Confirm that errors are getting passed back correctly
- Test the S3 trigger outside of podpac.
  - [ ] Make sure that the output file is put into the specified output directory


podpac/core/managers/aws.py
```diff
@@ -173,6 +181,13 @@ def _function_name_default(self):
 
         return settings["FUNCTION_NAME"]
 
+    @tl.default("function_triggers")
```
dsullivan-creare: This block seems to indicate I must set `function_eval_trigger` to 'APIGateway' in order to build the api gateway. More generally, it is not obvious to me how to specify that I want optional resources built (e.g. API Gateway).

dist/aws/handler.py
```diff
@@ -1,144 +1,215 @@
...
+
+        # return pipeline definition
+        return pipeline
+
+    elif trigger == "APIGateway":
+        print("Triggered from API Gateway")
+
+        pipeline = default_pipeline()
+        pipeline["url"] = event["queryStringParameters"]
+        pipeline["params"] = urllib.parse_qs(urllib.urlparse(pipeline["url"]).query)
```
dsullivan-creare: We actually don't need to do any URL parsing, because `event['queryStringParameters']` is a dict with all of the params already parsed.

dist/aws/handler.py
```diff
@@ -1,144 +1,215 @@
...
+                    pipeline["output"]["format_kwargs"]["return_base64"] = True
+
+        return pipeline
 
-    # TODO: handle "invoke" and "APIGateway" triggers explicitly
     else:
-        print("Not triggered by S3")
+        raise Exception("Unsupported trigger")
 
-        url = event["queryStringParameters"]
```
dsullivan-creare: It looks like it was in the porting of this block that we lost the isinstance check.

PR #338, opened 2019-11-13
FIX: add `to_image` method to UnitsDataArray and rename `get_image` to `to_image` in units module

mlshapiro:
This is a small change, but I think important.

I think its a little opaque to have a single `UnitsDataArray.to_format()` function with different formats and different corresponding args and kwargs. I think its better to write individual methods (i.e. `to_image()`) and then let the `to_format()` function just pick the best one based on the format. This is already mostly the case, but as a user I will look for the `to_image()` method if I'm trying to output the geotiff, not the `to()` method.

Also, `get_image` is not in keeping with the pandas/xarray standard of `to_format..()` for outputs. This again is not that important, but I want to try and keep this as consistent as possible.

### DIFF
--- a/podpac/core/pipeline/output.py
+++ b/podpac/core/pipeline/output.py
@@ -19,7 +19,6 @@ import numpy as np
 import traitlets as tl
 
 from podpac.core.node import Node
-from podpac.core.units import get_image
 
 
 class Output(tl.HasTraits):
@@ -210,4 +209,4 @@ class ImageOutput(Output):
 
     # TODO: docstring?
     def write(self, output, coordinates):
-        self.image = get_image(output, format=self.format, vmin=self.vmin, vmax=self.vmax, return_base64=True)
+        self.image = output.to_image(format=self.format, vmin=self.vmin, vmax=self.vmax, return_base64=True)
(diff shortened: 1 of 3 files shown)


podpac/core/pipeline/output.py
```diff
@@ -210,4 +209,4 @@ def __init__(self, node, name, format=None, mode=None, vmin=None, vmax=None):
 
     # TODO: docstring?
     def write(self, output, coordinates):
-        self.image = get_image(output, format=self.format, vmin=self.vmin, vmax=self.vmax, return_base64=True)
+        self.image = output.to_image(format=self.format, vmin=self.vmin, vmax=self.vmax, return_base64=True)
```
mpu-creare: This will all get deprecated in 2.0.

mpu-creare:
A few notes: 
* the `to` method is actually used for converting to different UNITS -- this matches the `pint` api
* the `to_format` method is a helper function that deals with string format types
* I agree that we should write a `to_geotiff` and `to_other_format` methods as they come up.

jmilloy:
Yep, another vote for separate methods.

mlshapiro:
> * the `to` method is actually used for converting to different UNITS -- this matches the `pint` api
> * the `to_format` method is a helper function that deals with string format types

I meant `to_format`, my fault. Comment corrected.

mlshapiro:
So leave this as `to_image` for now?

mlshapiro:
Great merging

Issue #238, opened 2019-04-12
Interpolation and intersection ignores ctype


jmilloy:
**Description**
The ctype ('left', 'right', 'midpoint', or 'point') for coordinates is not used anywhere.

**Steps to Reproduce**
```
import numpy as np
from podpac import Coordinates
from podpac.data import Array

data = np.random.random((100, 100))
nc = Coordinates([np.linspace(0, 10, 100), np.linspace(0, 10, 100)], dims=['lat', 'lon'], ctype='left')
node = Array(source=data, native_coordinates=nc)

c_left = Coordinates([np.linspace(1, 2, 20), np.linspace(1, 2, 20)], dims=['lat', 'lon'], ctype='left')
c_right = Coordinates([np.linspace(1, 2, 20), np.linspace(1, 2, 20)], dims=['lat', 'lon'], ctype='right')

o_left = node.eval(c_left)
o_right = node.eval(c_right)
```

**Expected Behavior**

`o_left` and `o_right` should be different, because they represent different locations in the data source.

**Observed Behavior**

`o_left` and `o_right` are the same.

**Additional Notes**

 * Possibly we can check `ctype` before intersecting or interpolating, the same way that we check `crs`. The `transform` method demonstrates that a similar `as_ctype` method could be easy to implement and integrate.
 * I'm not sure how the segment_lengths complicates this issue.

jmilloy:
@mpu-creare Have I thought this through correctly? And if so, what is the priority on this to you? Is it properly labeled a bug, or is it more of an enhancement to you? It goes right along with handling units and crs conversions -- does it belong the 1.0.0 release as a "feature complete" issue?

mpu-creare:
I'd label it as a bug... but it's probably more of a feature. You are correct, we haven't paid attention to it since we don't have a use case yet. We will very quickly begin having use-cases for `point` vs `midpoint`, but `left` and `right` will probably lag for quite a while.

jmilloy:
Maybe I can fix the "bug" aspect by adding a check when intersecting (maybe another when interpolating) than just raises an exception, and we can make a separate issue for the feature.

[...]

PR #344, opened 2019-11-27
FIX: Fixing time selection when precision is different

mpu-creare:
Consider: 
```
c1 = Coordinates(['2012-05-19'], ['time'])
c2 = Coordinates(['2012-05-19T12:00:00'], ['time'])
c12 = c1.intersect(c2)
c21 = c2.intersect(c1)
```

With this PR, I propose that c12 and c21 should contain an intersection.

### DIFF
--- a/podpac/core/coordinates/coordinates1d.py
+++ b/podpac/core/coordinates/coordinates1d.py
@@ -350,19 +350,31 @@ class Coordinates1d(BaseCoordinates):
             index or slice for the selected coordinates (only if return_indices=True)
         """
 
+        # empty case
+        if self.dtype is None:
+            return self._select_empty(return_indices)
+
         if isinstance(bounds, dict):
             bounds = bounds.get(self.name)
             if bounds is None:
                 return self._select_full(return_indices)
 
         bounds = make_coord_value(bounds[0]), make_coord_value(bounds[1])
+
+        # check type
+        if not isinstance(bounds[0], self.dtype):
+            raise TypeError(
+                "Input bounds do match the coordinates dtype (%s != %s)" % (type(self.bounds[0]), self.dtype)
+            )
+        if not isinstance(bounds[1], self.dtype):
+            raise TypeError(
+                "Input bounds do match the coordinates dtype (%s != %s)" % (type(self.bounds[1]), self.dtype)
+            )
+
         my_bounds = self.area_bounds.copy()
 
         # If the bounds are of instance datetime64, then the comparison should happen at the lowest precision
         if self.dtype == np.datetime64:
-            if not isinstance(bounds[0], np.datetime64):
-                raise TypeError("Input bounds should be of type np.datetime64 when selecting data from:", str(self))
-
             my_bounds, bounds = lower_precision_time_bounds(my_bounds, bounds, outer)
 
         # full
--- a/podpac/core/coordinates/coordinates1d.py
+++ b/podpac/core/coordinates/coordinates1d.py
@@ -356,13 +356,24 @@ class Coordinates1d(BaseCoordinates):
                 return self._select_full(return_indices)
 
         bounds = make_coord_value(bounds[0]), make_coord_value(bounds[1])
+        my_bounds = self.area_bounds.copy()
+
+        # If the bounds are of instance datetime64, then the comparison should happen at the lowest precision
+        if self.name == "time" and isinstance(my_bounds[0], np.datetime64) and isinstance(bounds[0], np.datetime64):
+            mine = np.timedelta64(1, my_bounds[0].dtype.name.replace("datetime64", "")[1:-1])
+            your = np.timedelta64(1, bounds[0].dtype.name.replace("datetime64", "")[1:-1])
+
+            if mine > your:
+                bounds = [b.astype(my_bounds[0].dtype) for b in bounds]
+            else:
+                my_bounds = [b.astype(bounds[0].dtype) for b in my_bounds]
 
         # full
-        if self.bounds[0] >= bounds[0] and self.bounds[1] <= bounds[1]:
+        if my_bounds[0] >= bounds[0] and my_bounds[1] <= bounds[1]:
             return self._select_full(return_indices)
 
         # none
-        if self.area_bounds[0] > bounds[1] or self.area_bounds[1] < bounds[0]:
+        if my_bounds[0] > bounds[1] or my_bounds[1] < bounds[0]:
             return self._select_empty(return_indices)
 
         # partial, implemented in child classes
--- a/podpac/core/coordinates/coordinates1d.py
+++ b/podpac/core/coordinates/coordinates1d.py
@@ -12,7 +12,7 @@ import traitlets as tl
 
 from podpac.core.utils import ArrayTrait
 from podpac.core.coordinates.utils import make_coord_value, make_coord_delta, make_coord_delta_array
-from podpac.core.coordinates.utils import add_coord, divide_delta
+from podpac.core.coordinates.utils import add_coord, divide_delta, lower_precision_time_bounds
 from podpac.core.coordinates.utils import Dimension, CoordinateType
 from podpac.core.coordinates.base_coordinates import BaseCoordinates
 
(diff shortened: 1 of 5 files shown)


mpu-creare:
Note, this change is required to have the SMAP EGI node work when requesting data for a single date.

jmilloy:
I want to look at this more closely later this afternoon. I think this makes sense. Leaving myself a note to double check - I think we may only want the lower precision in one direction, that is, `2019-01-01` would intersect `2019-01` but not `2019-01-01 12:00:00`.

podpac/core/coordinates/coordinates1d.py
```diff
@@ -356,13 +356,24 @@ def select(self, bounds, return_indices=False, outer=False):
                 return self._select_full(return_indices)
 
         bounds = make_coord_value(bounds[0]), make_coord_value(bounds[1])
+        my_bounds = self.area_bounds.copy()
+
+        # If the bounds are of instance datetime64, then the comparison should happen at the lowest precision
+        if self.name == "time" and isinstance(my_bounds[0], np.datetime64) and isinstance(bounds[0], np.datetime64):
```
jmilloy: I think we should check `if self.dtype == np.datetime64` instead of checking the name and the bounds types. We don't currently enforce any correspondence between the name and the dtype.

I do think it is a good idea to check the dtype of the input `bounds` to make sure they are the same as the coordinates dtype. I think you should move that to a separate check directly after `make_coord_value`.

podpac/core/coordinates/coordinates1d.py
```diff
@@ -356,13 +356,24 @@ def select(self, bounds, return_indices=False, outer=False):
...
+
+        # If the bounds are of instance datetime64, then the comparison should happen at the lowest precision
+        if self.name == "time" and isinstance(my_bounds[0], np.datetime64) and isinstance(bounds[0], np.datetime64):
+            mine = np.timedelta64(1, my_bounds[0].dtype.name.replace("datetime64", "")[1:-1])
+            your = np.timedelta64(1, bounds[0].dtype.name.replace("datetime64", "")[1:-1])
+
+            if mine > your:
+                bounds = [b.astype(my_bounds[0].dtype) for b in bounds]
+            else:
+                my_bounds = [b.astype(bounds[0].dtype) for b in my_bounds]
```
jmilloy: I think we should only do the else case.

If the coordinates are `[Oct 31, Nov 1, Nov 2]`
 - `select([Nov, Dec])` should return `[Nov 1, Nov 2]`
 - `select([Nov 1 10:00, Nov 1 12:00])` should return `[]` because Nov 1 isn't "within" 10:00 and 12:00 on Nov 1.

jmilloy: Can we write a utility function for this?

`my_bounds = lower_precision(self.area_bounds, other.bounds)`

podpac/core/coordinates/uniform_coordinates1d.py
```diff
@@ -384,12 +384,23 @@ def copy(self):
...
+        if self.name == "time" and isinstance(my_bounds[0], np.datetime64) and isinstance(bounds[0], np.datetime64):
+            mine = np.timedelta64(1, my_bounds[0].dtype.name.replace("datetime64", "")[1:-1])
+            your = np.timedelta64(1, bounds[0].dtype.name.replace("datetime64", "")[1:-1])
 
-        fmin = (lo - self.bounds[0]) / np.abs(self.step)
-        fmax = (hi - self.bounds[0]) / np.abs(self.step)
+            if mine > your:
+                bounds = [b.astype(my_bounds[0].dtype) for b in bounds]
+            else:
+                my_bounds = [b.astype(bounds[0].dtype) for b in my_bounds]
```
jmilloy: Can you help me understand why this is necessary? This is called by the `select` method you modified in the parent class, so the bounds should already be changed be the time we get here.

If not, factor this out and use the utility method I proposed above.

podpac/core/coordinates/coordinates1d.py
mpu-creare: Changed it to: 
```python
        if self.dtype == np.datetime64:
            if not isinstance(bounds[0], np.datetime64):
                raise TypeError("Input bounds should be of type np.datetime64 when selecting data from:", str(self))
```

podpac/core/coordinates/coordinates1d.py
jmilloy: I think it's helpful to also think about the primary use-case for this, which is interpolation, where the requested coordinates are `bounds` and the native coordinates are `my_bounds`. You can verify that in the DataSource eval:

```
        (
            self._requested_source_coordinates,
            self._requested_source_coordinates_index,
        ) = self.native_coordinates.intersect(coordinates, outer=True, return_indices=True)
```

When the requested coordinates are at higher resolution than the native_coordinates, you need convert `bounds` to the (lower) resolution of `my_bounds`. This is the `mine > your` case that I was arguing we should get rid of. So I guess I was wrong! More on that below.

When the native_coordinates are at higher resolution than the requested coordinates, then you need convert `my_bounds` to the (lower) resolution `bounds`. This is the `else` case. *Then the interpolation needs to average all of the native_coordinates.* This is similar to #238, which I believe is still unsolved (and in my opinion, is quite serious).

I argue that the `mine > your` case should *only* apply when `outer=True`. Here is the documentation on that:

```
        The default selection returns coordinates that are within the bounds::

            In [1]: c = ArrayCoordinates1d([0, 1, 2, 3], name='lat')

            In [2]: c.select([1.5, 2.5]).coordinates
            Out[2]: array([2.])

        The *outer* selection returns the minimal set of coordinates that contain the bounds::
        
            In [3]: c.select([1.5, 2.5], outer=True).coordinates
            Out[3]: array([1., 2., 3.])
```

I think that `2018-01-01` is not **within** `2018-01-01 12:00` and `2018-01-01 14:00`, which is the what select means when `outer=False`. But `2018-01-01` *is* in **the minimial set of coordinates that contain** `2018-01-01 12:00` and `2018-01-01 14:00`, which is what select means when `outer=True`.

mpu-creare: I can live with `mine > yours` when `outer == True`. 

I also agree that #238 being outstanding is quite serious, and some of this would be handled better with that in place.

jmilloy: I edited my comment and added a note about what I think is the SMAP case (the else case) and why the unresolved handling of higher resolution native coordinates may not be an issue there. If I heard you correctly, there is 1 higher resolution data point `2018-01-01T12:00:00.000` available for each day, so no reduction is necessary when requesting data for a particular day `2018-01-01`.

mpu-creare: you got it correct. So the else case was the SMAP case.

mpu-creare: Good idea, factored it out.

podpac/core/coordinates/uniform_coordinates1d.py
mpu-creare: The `bounds` get changed, that's fine, but `my_bounds` do not, so we need to do it again to `my_bounds`. We don't want to change `my_bounds` in-place in the object because other parts of the pipeline might still need the higher precision.

jmilloy: Ah, right, okay. Well, let's factor that out then, if we can.

mpu-creare:
Ok @jmilloy . I made those changes. Please merge when you're happy.

podpac/core/coordinates/coordinates1d.py
```diff
@@ -350,19 +350,31 @@ def select(self, bounds, return_indices=False, outer=False):
...
+        if not isinstance(bounds[0], self.dtype):
+            raise TypeError(
+                "Input bounds do match the coordinates dtype (%s != %s)" % (type(self.bounds[0]), self.dtype)
+            )
+        if not isinstance(bounds[1], self.dtype):
+            raise TypeError(
+                "Input bounds do match the coordinates dtype (%s != %s)" % (type(self.bounds[1]), self.dtype)
+            )
+
         my_bounds = self.area_bounds.copy()
```
mpu-creare: Will this throw errors when self.dtype == float and bounds == int?

jmilloy: I don't think so, because `make_coordinate_value` converts ints to floats.

mpu-creare: Awesome. I like this check -- it shows that you have an intimate knowledge of this class. :)

PR #348, opened 2019-12-17
Nodes with multiple outputs

jmilloy:
This is the replacement PR for #309


jmilloy:
@mlshapiro @mpu-creare Will you take a peek at commit https://github.com/creare-com/podpac/pull/348/commits/e97bbb91f602cc4ff8c6feaf8596c937ac987b82 and let me know what you think?

 * I split the `podpac.core.data.types` module into `reprojection`, `wcs`, `pydap`, `array_source`, and `file.
 * I standardized the `file.CSV`, `file.Dataset`, `file.Zarr`, and `file.H5PY` datasources to use the same base class.

Everything should work, tests are passing, etc, but do you have any naming or organizational preferences that I should heed at this time, or further standardization that I have overlooked?

I did this is here in preparation for adding multiple outputs support to these datasources, which will be a easier now that they are a bit more standardized.

mpu-creare:
Awesome. Thanks Jeff.  I would rename `wcs` --> `ogc` but that's it.

podpac/core/compositor.py
```diff
@@ -81,20 +82,51 @@ class Compositor(Node):
     source = tl.Unicode().tag(attr=True)
     sources = ArrayTrait(ndim=1)
     cache_native_coordinates = tl.Bool(True)
-
     interpolation = interpolation_trait(default_value=None)
+    strict_source_outputs = tl.Bool(False)
```
jmilloy: @mpu-creare Please let me know if you would like to use a different name for this. I also considered `strict_outputs_checking` and a number of others.

I also considered flipping the boolean and calling it `combine_outputs`, which will coincide with the similar `CombineOutputs` algorithm node.

We can also enable strict checking by default, if you would prefer.

podpac/core/compositor.py
```diff
@@ -81,20 +82,51 @@ class Compositor(Node):
...
+        return "_".join(str(source) for source in self.sources[:3])
 
     @tl.default("source_coordinates")
     def _source_coordinates_default(self):
         return self.get_source_coordinates()
 
+    @tl.validate("sources")
+    def _validate_sources(self, d):
+        self.outputs  # check for consistent outputs
+        return [copy.deepcopy(source) for source in d["value"]]
```
jmilloy: @mpu-creare @mlshapiro I think this is the desired behavior, but let me know if I'm mistaken. I haven't had a lot to do with interpolation. See `test_compositor::TestCompositor::test_iteroutputs_interpolation` for the case this covers.

podpac/core/compositor.py
```diff
@@ -229,35 +261,22 @@ def iteroutputs(self, coordinates):
             n_threads = 0
 
         if settings["MULTITHREADING"] and n_threads > 1:
```
jmilloy: @mpu-creare This is definitely a bug. When the `OrderedCompositor.composite` method breaks early (because there are no more NaN values to fill), the cleanup code after the yield loop never gets called. In this case, that includes setting the `_multi_threaded` attribute, closing the pool, and most importantly calling releasing the threads in the thread manager. The fix here evaluates the sources in parallel but waits for them all to finish before yielding results so that it can cleanup first.

The reason I've written all this is because there is a slight variation that we can consider. The following code still finishes evaluating all of the sources and cleans up the threads, but it also yields results as soon as they are ready. I believe this could be *marginally* more efficient, but I didn't think it was worth the unusual `GeneratorExit` idiom. I want to know if you think the following is preferable:

```
# evaluate nodes asynchronously using thread pool
pool = thread_manager.get_thread_pool(processes=n_threads)
results = [pool.apply_async(src.eval, [coordinates]) for src in src_subset]
pool.close()
try:
    for result in results:
        yield result.get()
except GeneratorExit:
    # the consuming generator was exited early
    # we still need to wait for the remaining threads and then release them from the thread manager
    for result in results:
        result.get()
thread_manager.release_n_threads(n_threads)
```

Note that it is correct to put the `thread_manager.release_n_threads(n_threads)` call *outside* the GeneratorExit because then it will be called whether the consuming loop breaks early or not.

podpac/core/compositor.py
```diff
@@ -229,35 +261,22 @@ def iteroutputs(self, coordinates):
...
-
-            # This prevents any more tasks from being submitted to the pool, and will close the workers one done
+            outputs = pool.map(lambda src: src.eval(coordinates), src_subset)
             pool.close()
-
-            # Release these number of threads back to the thread pool
             thread_manager.release_n_threads(n_threads)
-            self._multi_threaded = True
+            for output in outputs:
+                yield output
```
jmilloy: @mpu-creare This is definitely a bug. When the `OrderedCompositor.composite` method breaks early (because there are no more NaN values to fill), the cleanup code after the yield loop never gets called. In this case, that includes setting the `_multi_threaded` attribute, closing the pool, and most importantly calling releasing the threads in the thread manager. The fix here evaluates the sources in parallel but waits for them all to finish before yielding results so that it can cleanup first.

The reason I've written all this is because there is a slight variation that we can consider. The following code still finishes evaluating all of the sources and cleans up the threads, but it also yields results as soon as they are ready. I believe this could be *marginally* more efficient, but I didn't think it was worth the unusual `GeneratorExit` idiom. I want to know if you think the following is preferable:

```
# evaluate nodes asynchronously using thread pool
pool = thread_manager.get_thread_pool(processes=n_threads)
results = [pool.apply_async(src.eval, [coordinates]) for src in src_subset]
pool.close()
try:
    for result in results:
        yield result.get()
except GeneratorExit:
    # the consuming generator was exited early
    # we still need to wait for the remaining threads and then release them from the thread manager
    for result in results:
        result.get()
thread_manager.release_n_threads(n_threads)
```

Note that it is correct to put the `thread_manager.release_n_threads(n_threads)` call *outside* the GeneratorExit clause because then it will be called whether the consuming loop breaks early or not.

podpac/core/test/test_compositor.py
```diff
@@ -1,54 +1,328 @@
...
+            node.shared_coordinates()
+
+    @pytest.mark.xfail
+    def test_select_sources(self):
+        # with source coordinates defined
+        node = Compositor(sources=[ARRAY_LAT, ARRAY_LON, ARRAY_TIME], source_coordinates=COORDS)
+
+        subset = node.select_sources(COORDS)
+        subset = node.select_sources(COORDS[5:10, 5:10])
+        subset = node.select_sources(podpac.Coordinates([-1, -1, -1], dims=["lat", "lon", "time"]))
```
jmilloy: See #351

jmilloy:
@mpu-creare I believe this is ready for review.
 - tests are included
 - unfortunately, this includes a number of other bugs, maintenance fixes, and test updates, but I've tried to highlight these in comments to you where appropriate
 - no notebooks or examples are included

jmilloy:
- coverage increased to 92% (once merged)

podpac/core/node.py
```diff
@@ -103,13 +102,27 @@ class Node(tl.HasTraits):
...
     cache_ctrl = tl.Instance(CacheCtrl, allow_none=True)
     style = tl.Instance(Style)
 
+    @tl.default("outputs")
+    def _default_outputs(self):
+        return None
+
+    @tl.validate("output")
+    def _validate_output(self, d):
+        if self.outputs is None:
```
mpu-creare: Strange that you don't need `and d['value'] is not None`.... but fine.

podpac/core/node.py
```diff
@@ -103,13 +102,27 @@ class Node(tl.HasTraits):
      * ``_output``: the output of the most recent call to eval
```
mpu-creare: The doc string seems out of date? Should we add more to explain the attributes?

podpac/core/compositor.py
mpu-creare: Got it. Nice fix. Thanks!

podpac/core/algorithm/signal.py
```diff
@@ -156,11 +156,15 @@ def eval(self, coordinates, output=None):
...
+        if "output" not in source.dims:
+            result = scipy.signal.convolve(source, full_kernel, mode="same", method=method)
+        else:
+            # source with multiple outputs
+            result = np.array([scipy.signal.convolve(src, full_kernel, mode="same", method=method) for src in source])
         result = result[exp_slice]
 
         if output is None:
-            output = self.create_output_array(coordinates, data=result)
+            output = self.create_output_array(coordinates)
```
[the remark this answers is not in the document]
jmilloy: I don't know. Thanks.

podpac/core/algorithm/signal.py
```diff
@@ -156,11 +156,15 @@ def eval(self, coordinates, output=None):
         else:
             method = "auto"
 
-        result = scipy.signal.convolve(source, full_kernel, mode="same", method=method)
+        if "output" not in source.dims:
+            result = scipy.signal.convolve(source, full_kernel, mode="same", method=method)
+        else:
+            # source with multiple outputs
+            result = np.array([scipy.signal.convolve(src, full_kernel, mode="same", method=method) for src in source])
```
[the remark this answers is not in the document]
jmilloy: Would you like me to try to support that in this algorithm? Or should we wait until we need it? Sometimes it's probably just easier to define a custom Algorithm.

podpac/core/algorithm/stats.py
```diff
@@ -247,6 +242,89 @@ def reduce_chunked(self, xs, output):
         raise NotImplementedError
 
 
+class ReduceOrthogonal(Reduce):
```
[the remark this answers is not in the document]
jmilloy: Well, I just renamed `Reduce2` and moved it up.

podpac/core/algorithm/signal.py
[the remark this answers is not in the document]
mpu-creare: We'll support it when we need it.

podpac/core/compositor.py
```diff
@@ -81,20 +83,51 @@ class Compositor(Node):
...
-        return "_".join(source)
+        return "_".join(str(source) for source in self.sources[:3])
 
     @tl.default("source_coordinates")
     def _source_coordinates_default(self):
         return self.get_source_coordinates()
 
+    @tl.validate("sources")
+    def _validate_sources(self, d):
+        self.outputs  # check for consistent outputs
```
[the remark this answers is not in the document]
jmilloy: Yeah, I factored out the output validation so that it could be called here, but decided this was cleaner.

podpac/core/compositor.py
```diff
@@ -229,35 +262,20 @@ def iteroutputs(self, coordinates):
...
-            # Evaluate nodes in parallel/asynchronously
-            results = [pool.apply_async(f, [src]) for src in src_subset]
-
-            # Yield results as they are being requested, blocking when the thread is not finished
-            for src, res in zip(src_subset, results):
-                yield res.get()
-                # src._output = None # free up memory
-
-            # This prevents any more tasks from being submitted to the pool, and will close the workers one done
+            outputs = pool.map(lambda src: src.eval(coordinates), src_subset)
```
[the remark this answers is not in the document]
jmilloy: - pool.map and pool.map_async runs a single function on a list of inputs, whereas pool.apply and pool.apply_async run a single function on a single input.
 - pool.map and pool.apply are blocking and return the actual value(s), whereas pool.map_async and pool.apply_async immediately return *result objects* and you have to call `.get()` to block and get the value(s).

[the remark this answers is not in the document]
jmilloy: See [this comment](https://github.com/creare-com/podpac/pull/348/commits/bb6beeea5a9c449b4075e9b36ff263c2493a5990#r361759119) explaining this change and including an alternative that still uses apply_async.

podpac/core/node.py
jmilloy: I think that the validator isn't called when the default value is supplied, or something like that, but if someone supplied `=None` intentionally, you are probably right.

podpac/core/compositor.py
mpu-creare: I think stick with what you have.

PR #350, opened 2019-12-24
FIX #346: remove alt_units in favor of using CRS strings for everything.

### LINKED ISSUE #346: "You will likely lose important projection information" when using using alt_units
**Description**

The `set_vunits` coordinates utility function triggers a pyproj warning about losing information with some versions of pyproj. The Coordinates `alt_units` requires this function.

**Steps to Reproduce**

Using python 3.8 and pyproj 2.4.2:

```
from podpac.core.coordinates.utils import set_vunits
set_vunits("+proj=merc +vunits=m", "ft")
```

You can also run the tests in `test_coordinates_utils.py::test_set_vunits` and many others.

**Observed Behavior**

```
UserWarning: You will likely lose important projection information when converting to a PROJ string from another format.
```

This occurs in both locations that `set_vunits` calls `to_proj4()`.

**Expected Behavior**

The `+vunits` should be set without any warnings.

**Additional Notes**

I think the problem is that the way we handle the `alt_units` parameter uses proj4 strings internally, which are lossy. We should either, in order of preference:
 - fix `set_vunits` to avoid losing projection information, if possible
 - catch the UserWarnings in `set_vunits` and replace them with our own warning that using `alt_units` may be lossy. We should also point to an alternative.
 - remove the `alt_units` parameter. The user would have to use the `crs` string to specify altitude units.

mlshapiro:
Removes `alt_unit` keyword in favor of sticking to PROJ strings and pyproj processing.

```python
alt = ArrayCoordinates1d([0, 1, 2], name="alt")
c = Coordinates([alt], crs="+proj=merc +vunits=ft")

# transform coordinates
c.transform("+proj=merc +vunits=m")
```

Unfortunately this makes it a little more strict to define CRS when using alt units.
We should decide if this is *too* strict for common users.

This will now throw an error since WGS84 does not have vertical units:

```python
alt = ArrayCoordinates1d([0, 1, 2], name="alt")
Coordinates([alt], crs="EPSG:4326")
```

There is no simple way to say "use CRS EPSG:4326" *and* use meters for altitude units.


jmilloy:
I'm in favor of this.

@mpu-creare Do you know of a data source that we care about that uses WGS84 *and* has altitude? Or can we expect anyone using altitude to not be using WGS84?

mpu-creare:
We really don't have ANY datasets with Alt units that I'm aware of... does GFS have anything with altitude?

mpu-creare:
I also approve of this... the more we can push to Proj4 the better.

PR #352, opened 2020-01-03
Adds validation, tests, and docstring for Compositor source_coordinates

jmilloy:


podpac/core/compositor.py
```diff
@@ -189,8 +201,8 @@ def select_sources(self, coordinates):
 
             except:  # Likely non-monotonic coordinates
                 _, I = self.source_coordinates.intersect(coordinates, outer=False, return_indices=True)
-
-            src_subset = self.sources[I]
+            i = I[0]
```
mpu-creare: Does this only select a single source then? It could potentially be multiple sources that satisfy the constraint...

jmilloy: No. `intersect` returns a *list* of indexes, one for each dimension of the coordinates. Since the coordinates are validatedto have a single dimension, I is always a singleton list. The reason `self.sources[I]` worked before was because numpy silently/implicitly applies the singleton element of `I[0]` when indexing `self.sources`. If `source_coordinates` has more than one dimension (which is now prevented), the `I` will not be a singleton list and `self.sources[I]` fails. The `I[0]` is not strictly necessary, but it is correct and more explicit.

jmilloy: You can also confirm in the tests that I include an example that selects both sources, only the first source, only the second, and no sources.

mpu-creare: Awesome. That solves it. Had I spent time actually looking harder, i would have figured it out. Thanks Jeff.

PR #364, opened 2020-02-04
Feature/geotiff export

mpu-creare:
This adds geotiff export support to podpac. This is in support of the WCS OGC endpoint. 

Right now the rotated coordinates are probably not working, but with this update 90% of the WCS endpoint should work. That piece is untested.

### DIFF
--- a/podpac/core/coordinates/coordinates.py
+++ b/podpac/core/coordinates/coordinates.py
@@ -32,6 +32,11 @@ from podpac.core.coordinates.stacked_coordinates import StackedCoordinates
 from podpac.core.coordinates.dependent_coordinates import DependentCoordinates
 from podpac.core.coordinates.rotated_coordinates import RotatedCoordinates
 
+# Optional dependencies
+from lazy_import import lazy_module, lazy_class
+
+rasterio = lazy_module("rasterio")
+
 
 class Coordinates(tl.HasTraits):
     """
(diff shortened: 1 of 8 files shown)


mpu-creare:
Okay. Tests are passing. Let me know if you have any comments @jmilloy, I'd like to merge today.

podpac/core/coordinates/coordinates.py
```diff
@@ -442,6 +451,38 @@ def from_url(cls, url):
...
+        """ Creates Coordinates from GDAL Geotransform. 
+        
+        """
+        try:
+            rcoords = RotatedCoordinates.from_geotransform(geotransform)
+            if rcoords.theta == 0:
+                raise Exception  # More appropriate to use uniform coordinates
+            coords = Coordinates([rcoords], dims=["lat,lon"], crs=crs)
+            return coords
+        except:
```
jmilloy: Can you make this more specific? This is a big red flag for me because obscured exceptions are really hard to figure out.

If this is meant to catch the `rcoords.theta == 0`, then can we just do that check? If you need to catch other exceptions, too, I'd rather see something like this:

```
try:
    rcoords = RotatedCoordinates.from_geotransform(geotransform)
except <SpecificException>:
    _logger.debug(<some description of the expected/handled error>)
    return self._from_geotransform_nonrotated(geotransform)

if rcoords.theta == 0:
    _logger.debug("Rasterio source dataset does not have Rotated Coordinates")
    return self._from_geotransform_nonrotated(geotransform)

return Coordinates([rcoords], dims=['lat,lon'], crs=crs)
```

podpac/core/coordinates/coordinates.py
```diff
@@ -692,6 +733,19 @@ def shape(self):
...
+        if len(self.dims) == len(self.udims):  # shortcut in case no stacking
+            return self.shape
+
+        ushape = []
+        for ud in self.udims:
+            for i, d in enumerate(self.dims):
+                if ud in d:
+                    ushape.append(self.shape[i])
+                    break
+        return tuple(ushape)
```
jmilloy: I think you can just do

```
@property
def ushape(self):
    return tuple(self[dim].size for dim in self.udims)
```

podpac/core/coordinates/coordinates.py
```diff
@@ -801,6 +855,55 @@ def hash(self):
         json_d = json.dumps(self.full_definition, separators=(",", ":"), cls=podpac.core.utils.JSONEncoder)
         return hash_alg(json_d.encode("utf-8")).hexdigest()
 
+    @property
+    def geotransform(self):
+        """ :tuple: GDAL geotransform. """
+        # Make sure we only have 1 time and alt dimension
+        if "time" in self.udims and self.ushape[self.udims.index("time")] > 1:
```
jmilloy: `if "time" in self.udims and self['time'].size > 1`

podpac/core/coordinates/coordinates.py
```diff
@@ -801,6 +855,55 @@ def hash(self):
...
+    def geotransform(self):
+        """ :tuple: GDAL geotransform. """
+        # Make sure we only have 1 time and alt dimension
+        if "time" in self.udims and self.ushape[self.udims.index("time")] > 1:
+            raise TypeError(
+                'Only 2-D coordinates have a GDAL transform. This array has a "time" dimension of {} > 1'.format(
+                    self.ushape[self.udims.index("time")]
+                )
+            )
+        if "alt" in self.udims and self.ushape[self.udims.index("alt")] > 1:
```
jmilloy: if "alt" in self.udims and self['alt'].size > 1

podpac/core/coordinates/coordinates.py
```diff
@@ -801,6 +855,55 @@ def hash(self):
...
+            transform = rasterio.transform.Affine.translation(
+                self[first].start - self[first].step / 2, self[second].start - self[second].step / 2
+            ) * rasterio.transform.Affine.scale(self[first].step, self[second].step)
+            transform = transform.to_gdal()
+        elif (
+            "lat,lon" in self.dims
+            or "lon,lat" in self.dims
+            and isinstance(self._coords.get("lat,lon", self._coords.get("lon,lat")), RotatedCoordinates)
+        ):
+            transform = self._coords.get("lat,lon", self._coords.get("lon,lat")).geotransform
```
jmilloy: You can ignore me on this one, but I scratched my head for a few moments. It would much easier to read (and actually fewer lines) as two cases. This still works, right?

```
        elif "lat,lon" in self.dims and isinstance(self._coords["lat,lon"], RotatedCoordinates):
            transform = self._coords["lat,lon"].geotransform
        elif "lon,lat" in self.dims and isinstance(self._coords["lon,lat"], RotatedCoordinates):
            transform = self._coords['lon,lat'].geotransform
```

podpac/core/coordinates/coordinates.py
mpu-creare: AH! I was looking for something like that! THanks!

podpac/core/coordinates/coordinates.py
mpu-creare: I didn't quite take your suggestion but yes, I was being too cute (read obscure) with my exception raising.

podpac/core/coordinates/coordinates.py
mpu-creare: much better.

mpu-creare:
@jmilloy  Looks like the tests are passing. Please merge if you're happy. Otherwise I'll merge it when I have a moment.

PR #361, opened 2020-02-04
Feature/billing

mlshapiro:
Enable budgeting for podpac resources through the `aws` module.

Budgets can be made for function + associated resources or for all podpac created resources listed under an AWS account. Budget tracking/filtering is done through special tags created at resource generation: 

{`_podpac_resource`: `true`}
{`_podpac_resource_hash`: `hash`}

I anticipate corner cases for resources that were not created at the same time. This will need to get thoroughly tested before we make any promises about budgeting capabilities. For now this more of potential notification than a guaranteed notification.

To use the feature, create the settings:

```
settings["AWS_BUDGET_AMOUNT"] = 10.0 # budget for all podpac AWS resources, in USD
settings["AWS_BUDGET_EMAIL"] = "user@email.com" # notification e-mail for budget alarms
```

Also documented here: https://github.com/creare-com/podpac-examples/blob/develop/notebooks/developer/aws-lambda-tutorial.ipynb

Then use the `aws.Lambda` node as before.

See https://github.com/creare-com/podpac-examples/blob/develop/notebooks/developer/test/test-lambda.ipynb for an example of how to budget for all podpac resources created through the `aws` module.

PR #365, opened 2020-02-05
WIP: Release/1.3.0

mpu-creare:
This is waiting on #364 and ~~#361~~ . But starting so that I can make the release notes.


mpu-creare:
A number of these bug fixed really need some unittests!

PR #369, opened 2020-02-07
Feature/smap auth tag

mlshapiro:
Passing EarthData credentials into the Lambda functions changed in v1.3.

This PR is more of a discussion - do we want to tag EarthData credentials as attrs?

### DIFF
--- a/podpac/core/authentication.py
+++ b/podpac/core/authentication.py
@@ -159,13 +159,9 @@ class EarthDataSession(Session):
         print("Updating login information for: ", self.hostname)
 
         if username is None:
-            username = input("Username: ")
-
-        settings["username@" + self.hostname] = username
+            self.username = input("Username: ")
 
         if password is None:
-            password = getpass.getpass()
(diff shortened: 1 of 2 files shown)


jmilloy:
Certainly in principle they should not be Node attrs. Do we need an additional tag that flags a few attributes as lambda metadata? The aws manager should handle sending it.

mpu-creare:
So, the real issue is being able to pass username/password credentials to a completely generic lambda function. Let's discuss at today's meeting.

mlshapiro:
I like the idea of having attributes tagged specifically for awslambda nodes (maybe better to have auth attributes like `.tag(auth=True)`, but it seems like the right way to deal with credentials is via settings.

We have the capability to customize for each AWS lambda node via `eval_settings`. I think the missing piece here is actually in the authentication handling - connecting that to settings. I'll think about how we want to automate this for friday.

mpu-creare:
We decided you have to pass the auth information via `eval_settings`. So, just need to document.

Issue #297, opened 2019-07-18
Harmonize __repr__ for nodes


mpu-creare:
**Description**
Right now some of our nodes have a nice __repr__, but really we should have a default one

jmilloy:
See also #341

jmilloy:
**repr**

One line with class name, node type if needed, and key info.

```
>>> repr(sm)
<SMAP product='SPL4SMAU'>
>>> repr(node)
<Rasterio source='s3://bucket/path/to/file'>
```

**str**

One line with class name, node type if needed, key info, and a list of node attributes.

```
>>> sm
<SMAP product='SPL4SMAU' (attrs: base_url, interpolation, layer_key, output, product, units, version)>
>>> node
<Rasterio source='s3://bucket/path/to/file' (attrs: band, crs, interpolation, output, outputs, source, units)>
```

jmilloy:
Note you can easily see the details using `json_pretty`:

```
>>> print(sm.json_pretty)                                                                                                        
{
    "SMAP_SPL4SMAU": {
        "plugin": "__main__",
        "node": "SMAP",
        "attrs": {
            "base_url": "https://n5eil02u.ecs.nsidc.org/opendap/SMAP",
            "interpolation": {
                "method": "nearest",
                "params": {
                    "time_tolerance": "2,h"
                }
            },
            "layer_key": "{rdk}sm_surface_analysis",
            "version": 4
        }
    }
}
>>> print(node.json_pretty)
{
    "Rasterio": {
        "node": "core.data.rasterio_source.Rasterio",
        "attrs": {
            "band": 1,
            "source": "s3://bucket/path/to/file"
        }
    }
}
```

But it is a bit obtuse. I wonder if we want something that will print output more like:

>>> node.attrs()
podpac.datalib.SMAP
	base_url: https://n5eil02u.ecs.nsidc.org/opendap/SMAP
	interpolation: {'method': 'nearest', 'params': {'time_tolerance': numpy.timedelta64(2,'h')}}
	layer_key: {rdk}sm_surface_analysis
	product: SPL4SMAU
	units: None
	version: 4
>>> 
podpac.core.data.rasterio_source.Rasterio'>
	band: 1
	crs: None
	interpolation: nearest
	output: None
	outputs: None
	source: s3://bucket/path/to/file
	units: None
```

mpu-creare:
fixed by #371

PR #371, opened 2020-02-11
Node attr improvements

### LINKED ISSUE #359: Improved handling of computed Dataset source, Compositor sources, and other Node attrs
**Description**

Computed attrs should not be included when serializing a node. Furthermore, including them can lead to deadlock if the computed values attempt to use the cache, and the current workarounds are quite opaque. Some examples of computed attrs are `Dataset.source` and `Compositor.sources` in SMAP and MODIS datalib; `Compositor.outputs`; and `Rasterio.nan_vals`.

 * `Datasource.source`, `Datasource.interpolator`, `Compositor.sources`, and `Algorithm.inputs` should be `Node.attrs` like everything else, without requiring special handling.
 * Like `native_coordinates`/`get_native_coordinates()`, we should provide `get_*` machinery for `Datasource.source`, `Compositor.sources`, and other attrs that are sometimes provided explicitly and sometimes computed from other attrs.
 * These `get_*` methods should *remove* the `'attr'` tag from the traits when called, so that the corresponding traits are not inclued in the base_definition.
    - note that `base_definition` can be referenced before the `get_*` method or default has been called, so it should *also* check `trait_is_defined`. I believe this will offer the desired behaviour.
    - `@tl.default` methods may also want to remove the `'attr'` tag, because default values also do not need to be included in the definition, though in some case, it may be better to include them for generating consistent cache keys
 * In some cases, `Datasource.native_coordinates` are *not* computed should be *included* in the base_definition. These changes will make it easier to do that clearly, too.

jmilloy:
Closes #359 #297 #378

 * Use traits tagged `attr=True` for all node attrs
 * Use properties for computed attributes (instead of a trait with `tl.default`)
 * Adds `@cached_property` decorator to easily avoid recomputing properties where appropriate
 * `native_coordinates` is a DataSource property (instead of a trait with `tl.default`)
    * Still use `get_native_coordinates` in child classes to compute coordinates (native coordinates cannot *also* be provided at runtime).
    * Use `cache_native_coordinates` to cache the native coordinates using the `cache_ctrl`. (Default is False.)
    * The compositor uses `set_native_coordinates` to pass shared coordinates into its sources so that they don't need to call `get_native_coordinates`.
 * Adds `DiskCacheMixin`, `NoCacheMixin`, and `S3Mixin` to factor out some common Node idioms
 * Includes `RequestSessionMixin` from #376 for better EarthData credentials handling.
 * file datasources are overhauled and standardized
 * Node `__repr__` and `__str__` are standardized and easily extensible

The datalibs will need careful attention due to these changes:
 - [x] drought_monitor
 - [x] egi
 - [x] gfs
 - [x] intake
 - [x] smap
 - [x] smap_egi
 - [x] terraintiles


jmilloy:
Personal notekeeping. I still need to revisit the `cached_property` and deploy it as neede throughout podpac and the datalibs. I also plan to review the datalib `tl.defaults`.

podpac/core/utils.py
```diff
@@ -329,3 +329,41 @@ def _get_from_url(url):
...
+        elif self.has_cache(key):
+            value = self.get_cache(key)
+            setattr(self, key, value)
+        else:
+            value = fn(self)
+            setattr(self, key, value)
+            self.put_cache(value, key)
+        return value
+
+    return wrapper
```
jmilloy: @mpu-creare I added two cached property decorators:
 * `cached_property`: store the property as a private attribute
 * `cached_node_property`: store the property as a private attribute *and* using the node `cache_ctrl`

What do you think? I used the `cached_property` everywhere I thought it made sense throughout podpac core. Most of those can be replaced with `cached_node_property` (but a special few cannot). So the question is: do you think I should use the podpac cache version `@cached_node_property` everywhere I can, instead? Another way to put this is: do you see `@cached_property` or `@cached_node_property` as the "common" case?

I will rename them according to your answer.

jmilloy: Basically, I have an *internal* use for the one that doesn't use the podpac cache, but if you only want to provide users with the one that *does* use the podpac cache, I want to reduce this to one decorator and handle the other cases a special way.

But if you think we do want both, I'll leave it (though maybe you can suggest a better pair of names).

mpu-creare: I like the two decorators. You could probably also do it with an input to the decorator, but I really don't mind two. How about: `cached_property` and `cached_property_persistent` as the former only works for a session, while the latter *could* work across sessions depending on the user's cache preferences.

jmilloy:
@mpu-creare I'm trying to finish this out by replacing instances of `@tl.default` where a property should have been used. I feel good about everything except SMAP. Can you tell me which of these really do need to be customizable at object instantiation? (The others will be switched to properties.)

 - `smap.SMAPSource.rootdatakey`
 - `smap.SMAPSource.layerkey`
 - `smap.SMAPProperties.source`
 - `smap.SMAPProperties.layerkey`
 - `smap.SMAPDateFolder.base_url`
 - `smap.SMAPDateFolder.version`
 - `smap.SMAPDateFolder.layerkey`
 - `smap.SMAPDateFolder.is_source_coordinates_complete`
 - `smap.SMAP.layerkey`

Alternatively, should any of the `rootdatakey`, `layerkey`, `base_url`, on `version` traits be be tagged as `attr=True`? Seems likely... at least for base_url.

Similarly for smap egi, I see that `short_name`, `lat_key`, `lon_key`, `data_key`, and `version` come from `EGI`. Do they sometimes need to be customizable at object instantiation, or can they just be overwritten as properties in the `smap_egi.SMAP` class?

jmilloy:
@mpu-creare What do I need to do to use the AirMOSS datalib (what credentials do I need)? Is there documentation about that somewhere?

podpac/datalib/egi.py
```diff
@@ -229,8 +227,6 @@ def eval(self, coordinates, output=None):
                 " which case EGI is returning no data."
             )
             raise e
-        # Force update on native_coordinates (in case of multiple evals)
-        self.set_trait("native_coordinates", self.get_native_coordinates())
```
jmilloy: @mpu-creare The `native_coordinates` are now a property, so I just removed this `set_trait` call. Is that okay?

Originally I made the native_coordinates a cached_property, but this `set_trait` suggests to me that we shouldn't be saving anything between evals, so I changed it to just a normal property.

jmilloy:
@mpu-creare I've got the SMAP nodes working. Took some work. I'm pushing very soon.

mpu-creare:
Nice, thanks Jeff.

jmilloy:
@mpu-creare This is finally ready. It's huge. I'm not sure how much you will feasibly be able to review. Let me know if you want to talk about it or get a walk through in any way. The description has a decent overview.

podpac/core/data/array_source.py
```diff
@@ -43,30 +45,23 @@ class Array(DataSource):
...
                                      podpac.clinspace(1, 32, 32, 'lat'),
                                      podpac.clinspace(1, 34, 34, 'lon')])
-    >>> node = podpac.data.Array(source=data, native_coordinates=coords, outputs=['R', 'G', 'B'])
+    >>> node = podpac.data.Array(data=data, native_coordinates=coords, outputs=['R', 'G', 'B'])
     >>> output = node.eval(coords)
     """
 
-    source = ArrayTrait().tag(readonly=True)
-    native_coordinates = tl.Instance(Coordinates, allow_none=False).tag(attr=True)
+    data = ArrayTrait().tag(attr=True)
```
mpu-creare, line 52: I personally liked `source` because we use that everywhere. I agree `data` makes more sense here as `source` usually describes a `Node` or a `str`, but I just know it's going to trip me up.

podpac/core/data/datasource.py
```diff
@@ -475,75 +487,20 @@ def get_native_coordinates(self):
...
-                    rep += "\n\t\t%s: %s" % (c.name, c)
-                elif isinstance(c, StackedCoordinates):
-                    for _c in c:
-                        rep += "\n\t\t%s[%s]: %s" % (c.name, _c.name, _c)
-
-                # rep += '{}: {}'.format(c.name, c)
-        rep += "\n\tinterpolation: {}".format(self.interpolation)
-
-        return rep
+        if not self.trait_is_defined("_native_coordinates"):
```
mpu-creare, line 505: WOW that's a lot cleaner. Nice. NIIICEE.

podpac/core/utils.py
```diff
@@ -330,4 +326,72 @@ def _get_from_url(url):
...
+
+    Keyword Arguments
+    -----------------
+    use_cache_ctrl : bool
+        If True, the property is cached using the Node cache_ctrl. If False, the property is only cached as a private
+        attribute. Default False.
+
+    Notes
+    -----
+    The property should not depend on any other attributes that are not tagged with ``attr=True``.
```
mpu-creare, line 345: This double-negative is confusing. In fact, I'm confused. What?

podpac/datalib/egi.py
mpu-creare: I'm not sure @mlshapiro can you advise?

podpac/datalib/modis.py
```diff
@@ -0,0 +1,90 @@
...
+            return None
+
+        prefix = "/".join(prefix)
+        objs = self.s3.ls(prefix)
+        return [obj.replace(prefix + "/", "") for obj in objs if "_scenes.txt" not in obj]
+
+
+if __name__ == "__main__":
+
+    source = MODISSource(
```
mpu-creare, line 80: Is this functional?

podpac/datalib/nasaCMR.py
```diff
@@ -143,40 +147,13 @@ def _get_all_granule_pages(auth_session, url, entry_map, max_paging_depth=100000
...
         page_url = url + "&page_num=%d" % (i + 1)
-        page_entries = json.loads(_get_from_url(auth_session, page_url))["feed"]["entry"]
+        page_entries = json.loads(_get_from_url(page_url, session))["feed"]["entry"]
         if not page_entries:
             break
         entries.extend(list(map(entry_map, page_entries)))
     return entries
-
-
-def _get_from_url(auth_session, url):
```
mpu-creare, line 158: Thanks for moving this to utils

podpac/__init__.py
```diff
@@ -43,7 +43,7 @@ def makedirs(name, mode=511, exist_ok=False):
 from podpac.core.coordinates import Coordinates, crange, clinspace
 from podpac.core.node import Node, NodeException
 import podpac.core.authentication as authentication
-from podpac.core.utils import NodeTrait
```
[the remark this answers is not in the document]
jmilloy: Me, too. This merge request ought to fix that. The NodeTrait is now a `tl.Instance` instead of a `tl.ForwardDeclaredInstance`, and it imports the `Node` class as needed.

podpac/core/authentication.py
```diff
@@ -1,171 +1,192 @@
...
+        except ValueError as e:
+            if self.auth_required:
+                raise e
+            else:
+                _log.warning("No auth provided for session")
+
+        return s
+
+
+class S3Mixin(tl.HasTraits):
```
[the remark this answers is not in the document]
jmilloy: I'm not sure, I only see one. I moved it from node.py to authentication in a recent commit in this branch...?

podpac/core/data/datasource.py
jmilloy: This went through a silly number of iterations.

podpac/datalib/modis.py
jmilloy: Yes, the MODISSource works. The full product (compositor) does not.

podpac/datalib/nasaCMR.py
jmilloy: @mlshapiro  thanks.

podpac/core/utils.py
jmilloy: Is this better?

```
Notes
-----
Podpac caching using the cache_ctrl will be unreliable if the property depends on any non-tagged traits.
The property should only use node attrs (traits tagged with ``attr=True``).
```

podpac/core/data/array_source.py
jmilloy: It think it's weird to have different child classes with different types for the "same" trait. Or, to use the same trait name for different traits. But I see where you're coming from.

jmilloy:
@mpu-creare Okay, I addressed your comments, and tests are passing on my machine on python 2.7 and 3.5 now (as well as python 3.8).

mpu-creare:
Really nice work Jeff and Marc. This was a huge effort. I'm very impressed.

PR #376, opened 2020-03-13
Feature/earthdata auth

mlshapiro:
This PR refactors authentication handling to be a mixin. Generally, this results in much less code and hopefully simplifies the settings handling.

Classes that implement the `RequestsSessionMixin` will now have the option to require auth (`auth_required` trait). If auth is required, the class can set the credentials directly to the settings using the `set_credentials` method.

For example:

```python
from podpac import authentication
from podpac import Node

class NewNode(authentication.RequestsSessionMixin, Node):
    source="source"
    hostname="myhost.org"

node = NewNode()

# node attributes
node.hostname == "myhost.org"
type(node.session) == requests.Session

# username
# node.username is a property pulled from the settings
# if it is not defined, it will throw a ValueError telling the user to define using
# `set_credentials`
node.username -> ValueError

# this stores credentials on the settings
node.set_credentials(username="username", password="password")
node.username == settings["username@myhost.org"]
node.password== settings["password@myhost.org"]

node.session.auth = (username, password)
```

Overview of changes:

- The base class is the `podpac.core.authentication.RequestsSessionMixin`.
- Moves `EarthDataSession` to the `podpac.datalib.smap` module and renames to `SMAPSessionMixin`. This class is not used anywhere else, and is fairly specific to SMAP handling, so it should be in the datalib.
- Updates many instances of SMAP/NasaCMR `auth_session` to be `session` from mixin.
- Removes duplicated `_get_from_url` in favor of `podpac.core.utils._get_from_url`.
- Updates PyDAP to use `RequestsSessionMixin`
- Updates tests for authentication and PyDAP. Adds smap test for authentication.


jmilloy:
@mlshapiro I haven't looked through this really at all. On Friday I factored out the S3 auth into a mixin. I wonder if you like that idea. If the Earth Data auth might be used in different Nodes, a single mixin that adds the necessary traits and uses the correct settings by default might work for that, too.

The mixin looks like this:

```
class S3Mixin(tl.HasTraits):
    """ Mixin to add S3 credentials and access to a Node. """

    anon = tl.Bool(False)
    access_key_id = tl.Unicode(allow_none=True)
    secret_access_key = tl.Unicode(allow_none=True)
    region_name = tl.Unicode(allow_none=True)
    s3fs_kwargs = tl.Dict()

    @tl.default("access_key_id")
    def _get_access_key_id(self):
        return settings["AWS_ACCESS_KEY_ID"]

    @tl.default("secret_access_key")
    def _get_secret_access_key(self):
        return settings["AWS_SECRET_ACCESS_KEY"]

    @tl.default("region_name")
    def _get_region_name(self):
        return settings["AWS_REGION_NAME"]

    @cached_property
    def s3(self):
        if self.anon:
            return s3fs.S3FileSystem(anon=True)
        else:
            return s3fs.S3FileSystem(
                key=self.access_key_id,
                secret=self.secret_access_key,
                region_name=self.region_name,
                client_kwargs=self.s3fs_kwargs,
            )
```

In your case I think you could put the username and password traits and the `observe` methods into the mixin.

mlshapiro:
@jmilloy this is a great idea, i will take a look at your changes and modify this PR before any reviews.

mlshapiro:
@jmilloy what does `@cached_property` do? Does that mean the credentials can be in node cache? Does this have the potential to expose credentials if the cache is on S3?

jmilloy:
`@cached_property` only caches to a local private attribute in the object.

`@cached_property(use_cache_ctrl)` will use the podpac cache.

mlshapiro:
> `@cached_property` only caches to a local private attribute in the object.
> 
> `@cached_property(use_cache_ctrl)` will use the podpac cache.

Great thank you!

mlshapiro:
@jmilloy this is ready, but I will wait until your updates are merged from #370

doc/source/earthdata.md
```diff
@@ -36,29 +39,33 @@ a security issue for you, do not use this method.
 To store your credentials use the following code in an interactive Python session: 
 ```python
 from podpac.core.authentication import EarthDataSession
-eds = EarthDataSession()
-eds.update_login()
+
+EarthDataSession(username="<earthdata-username>", password="<earthdata-password>", store_credentials=True)
```
jmilloy: It looks like this is out of date now. Did I miss something, or did you?

mlshapiro: whoops. forgot to update docs. will do this now

jmilloy:
I merged this into #371. A bit of an effort. So we can merge this as soon as the docs are ready.

mlshapiro:
> 
> 
> I merged this into #371. A bit of an effort. So we can merge this as soon as the docs are ready.

Complete in afa17d390bbf6fce797892313c8e2699bdc57a7f

PR #380, opened 2020-04-02
Feature/deprecate py35 and below

mlshapiro:
Can we do this? Tests are already showing we need to

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,7 +21,7 @@ install_requires = [
     "pint>=0.8",
     "scipy>=1.0",
     "traitlets>=4.3",
-    "xarray>=0.10",
+    "xarray>=0.15",
     "requests>=2.18",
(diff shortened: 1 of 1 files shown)


jmilloy:
I think we can deprecate python 3.5 even if we can't deprecate python 2. When I asked about this in the fall Matt said we needed to keep python 2 support.

(I had to allow pyproj 2.2 in order to run the setup.py on my machine with python 2. I made that change this morning.)

mlshapiro:
ah ok, so in that case, we can say python >= 3.6, or py 2.7.  Too bad we can't deprecate py 2. If thats the case, xarray deprecated python 2 in 0.12, so we'll have to keep it at 0.11.

I'll close this PR and see if I can get the tests to work with setup.py as is

mpu-creare:
Late to the party, but agree with everything.

PR #385, opened 2020-04-07
Skip crs Validation when possible

jmilloy:
CRS validation with pyproj is very slow, so this is an optimization. For example, when `Coordinates` is subselecting itself in `__getitem__`, it does not need to revalidate the crs.

@mpu-creare @mlshapiro any qualms?

This reduces the runtime for the tests on my machine from 80s to 50s.

### DIFF
--- a/podpac/core/algorithm/stats.py
+++ b/podpac/core/algorithm/stats.py
@@ -875,6 +875,11 @@ class GroupReduce(UnaryAlgorithm):
         native_time_mask = np.in1d(N, E)
 
         # use requested spatial coordinates and filtered available times
+        import pdb
+
+        pdb.set_trace()  # breakpoint e7271746 //
+
+        # WHAT?
         coords = Coordinates(
             time=avail_time.data[native_time_mask],
             lat=requested_coordinates["lat"],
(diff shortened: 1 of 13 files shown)


podpac/core/algorithm/stats.py
```diff
@@ -875,6 +875,11 @@ def _get_source_coordinates(self, requested_coordinates):
         native_time_mask = np.in1d(N, E)
 
         # use requested spatial coordinates and filtered available times
+        import pdb
+
+        pdb.set_trace()  # breakpoint e7271746 //
+
+        # WHAT?
```
mpu-creare, line 882: I'm guessing  you'd want to remove this breakpoint

podpac/core/coordinates/coordinates1d.py
```diff
@@ -76,7 +76,7 @@ def __init__(self, name=None, ctype=None, segment_lengths=None):
 
             self.set_trait("segment_lengths", segment_lengths)
 
-        super(Coordinates1d, self).__init__()
+        # super(Coordinates1d, self).__init__()
```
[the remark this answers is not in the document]
jmilloy: It was a just in case. I took it out as part of profiling, need to revert that. Thanks.

podpac/core/algorithm/stats.py
jmilloy: Thanks. This code is never called, and you can tell because it uses the **old** Coordinates spec.

Issue #368, opened 2020-02-07
The cache lock should be specific to a Node type. 


mpu-creare:
**Description**
Right now two different nodes use the SAME cache lock. This leads to problems if, serially, both nodes need to check the cache. This can happen of Node1 checks has_cache, needs to compute it's sources for the source definition, which then needs to check the cache. 

This happened with the follow Code before 6e8c1f0a0a753b994dd5d4c75d0b041e089d04c0

**Steps to Reproduce**
```python
from podpac.datalib.terraintiles import TerrainTiles
from podpac import Coordinates, clinspace

# create terrain tiles node
node = TerrainTiles(tile_format='geotiff', zoom=8)

# create coordinates to get tiles
c = Coordinates([clinspace(40, 43, 1000), clinspace(-76, -72, 1000)], dims=['lat', 'lon'])

# evaluate node
o = node.eval(c)  # Locks when trying to get self.nan_vals because the cache is checked a second time
```

**Expected Behavior**
The reason for the lock in the first place is to ensure that two of the same Node don't try to simultaneously put something or retrieve something from the cache. This should NOT affect different Nodes. 

**Additional Notes**
Fix could be to implement a dictionary of locks within the thread manager that depends on the node definition, with a default global lock as a fallback... or perhaps the lock needs to happen deeper down in the caching machinery.

mpu-creare:
@jmilloy  FYI... I see you've run into this too.

mpu-creare:
Also related to #359

jmilloy:
> Fix could be to implement a dictionary of locks within the thread manager that depends on the node definition

As a first reaction, I don't think this will work because the problem occurs when the node definition property tries to access the cache, so then we wouldn't be able to use the node definition to create the cache lock.

> This can happen of Node1 checks has_cache, needs to compute it's sources for the source definition, which then needs to check the cache.

Right. I think that *nothing* that is computed should be in the source definition, which fixes this issue, and is what #359 is about. 

Unfortunately, even with #359, a user *could* inadvertently do it wrong and cause themselves a head-ache. I think we could add machinery that raises an exception when a node definition tries to access the cache. It might be a bit complicated, but I think it would be isolated to the cache submodule.

mpu-creare:
> As a first reaction, I don't think this will work because the problem occurs when the node definition property tries to access the cache, so then we wouldn't be able to use the node definition to create the cache lock.

Derp! That makes sense. And yes, I think #359 still makes this a lot better. 

[...]

PR #386, opened 2020-04-08
Feature/cosmos stations

mpu-creare:
Adding a datalib node for reading COSMOS station data.


podpac/core/node.py
```diff
@@ -577,8 +577,8 @@ def has_cache(self, key, coordinates=None):
         if self.cache_ctrl is None:
             return False
 
-        with thread_manager.cache_lock:
-            return self.cache_ctrl.has(self, key, coordinates=coordinates)
+        # with thread_manager.cache_lock:
```
mpu-creare, line 580: This keeps biting me with cached properties. It's pretty hard not to hit this lock in a single process. 

I think removing it is also pretty safe, since if multiple threads are reading there's not issue. The only foreseeable problem is if another Node if writing to the cache one one Thread, and a second Node reads it incompletely (reading faster than writing). But this is remote enough that I'm not going to worry about that case right now. 

The right thing to do is to apply the lock in the cache module.

podpac/core/algorithm/stats.py
```diff
@@ -916,14 +892,25 @@ def eval(self, coordinates, output=None):
...
+        if output is None:
+            coords = podpac.coordinates.merge_dims(
+                [coordinates.drop("time"), Coordinates([out.coords["time"]], ["time"])]
+            )
+            coords = coords.transpose(*out.dims)
+            output = self.create_output_array(coords, data=out.data)
+        else:
+            output.data[:] = out.data[:]
+
+        ## map
```
mpu-creare, line 905: @jmilloy : I completely changed how this node work... I figured that was okay because I suddenly had an actual use-case and there were no unit tests. :)

podpac/core/node.py
jmilloy: Okay. That's too bad. #371 was supposed to help with this, so I'm curious what problems you ran into.

Most of the times I had trouble here, it was because a node attr had a default method that tried to access the cache, and without the lock you get an infinite recursion instead, so removing it didn't help.

> I think removing it is also pretty safe, since if multiple threads are reading there's not issue.

That's true. We should use the correct synchronization tools to allow multiple simultaneous readers.

> The right thing to do is to apply the lock in the cache module.

Yeah, I guess this is also true. It would be more clear what needs to wait for what at that level, too.

jmilloy: Summary: I think we should 
 * remove the lock here in the node (as you have done here)
 * reinstate the `trait_is_default` check (see below)
 * make a new issue for adding appropriate synchronization into the cache module (including #368, simultaneous readers, etc)

---

We discussed this over in #368, and I was reminded that one of the things that I did during #371 was to omit node attrs that used the default value from the definition. In addition to looking cleaner to a user, this had the *primary* benefit of never calling `tl.default` methods when caching, which prevented most of the deadlocks that I was encountering.

(Note that it is sufficient to check `trait_is_defined` in order to avoid calling the `tl.default` methods when creating the definition, but that breaks the stability of the node definition and therefore also the node hash.)

The catch is that towards the end of #371, I reverted the `trait_is_default` check. Traitlets handles `List` and `Dict` trait defaults poorly, so it required a bit of hack in a few places. I forgot the primary motivation (deadlocking), and "cleaner node definitions" as a user-exeperience-optization just didn't justify the hacks.


https://github.com/creare-com/podpac/commit/35fe059e16d76dd8ffa12b9b5b297bc0f07076f7 is the commit that removed `trait_is_default` (but a lot of other things), and you can see an example of the "hack" to get traitlets container defaults to work as expected in coord_select.py. I think it would take about an hour to reinstate.

Does that make sense, and so do you agree I should add the `trait_is_default` check back in?

podpac/core/algorithm/stats.py
jmilloy: How do you feel about still leaving it without tests?

mpu-creare: Not great. I'm planning to add some tests, before finishing this PR.

mpu-creare: If you have bandwidth, that would be great if you could write some tests. I'm also trying to finish up the parallel manager branch's tests.

podpac/core/node.py
jmilloy: Yeah, the compositor `outputs` default really is always the problem. I'm surprised it would hit the thread lock, but doesn't go into an infinite loop without the lock. I'll just have to try it out when I get a chance.

Good point about the defaults changing between versions.

podpac/core/algorithm/stats.py
jmilloy: Yeah, I'll check in with you before I starting writing tests so we don't duplicate effort, but I might be able to get to it before you.

podpac/core/node.py
jmilloy: Okay. The cosmos stations node currently with or without the thread lock because you have set `outputs=None` for the compositor.

Without setting `outputs=None`, you get either deadlock (with the thread lock) or infinite recursion. I think this is a key point that needs to be fixed, because it is not clear, and it makes the compositor hard to use. I think that the outputs default needs to be None for the Compositor class with an explicit mixin or flag to enable automatic multiple outputs detection. Let me throw you a quick proposal for that.

mpu-creare: Yikes, I missed the infinite recursion... is there a sane way we can break out of it? I mean, if we managed to set up an infinite recursion, i suspect others will end up doing the same thing. Is it tied to caching? Could we detect a loop in the cache and then just recalculate when a loop is detected?

jmilloy: @mpu-creare Please let me know about #387 or #388. I think removing the thread lock here is fine, in either case.

jmilloy:
@mpu-creare You'll need to clear your COSMOSStations cache after pulling commit https://github.com/creare-com/podpac/pull/386/commits/c1738807df4d93bc7c4a1c783e9b5f7fdef95c43

PR #387, opened 2020-04-10
Bugfix: compositor auto_outputs flag

jmilloy:
Currently Compositors automatic detect `outputs` from the `sources`, which causes deadlock or infinite recursion in fairly normal usage, including SMAP and COSMOS. The problem is quite hard to diagnose, and the workaround is to explicitly set `outputs = None` in these compositor nodes.

The proposed solution is to add a `auto_outputs` flag to Compositors, which defaults to False.

The workaround can be removed from SMAP and COSMOS, and compositor child classes that want to autodetect multiple outputs from sources can enable this feature as follows:

```
class MyCompositor(podpac.core.OrderedCompositor):
    auto_outputs = True
```

*Note: either this PR or #388 should be merged, but not both.*


jmilloy:
We're going with this one, will update with a test before merging.

PR #388, opened 2020-04-10
BUGFIX: Extracts automatic outputs detection from compositor sources …

jmilloy:
Currently Compositors automatic detect `outputs` from the `sources`, which causes deadlock or infinite recursion in fairly normal usage, including SMAP and COSMOS. The problem is quite hard to diagnose, and the workaround is to explicitly set `outputs = None` in these compositor nodes.

The proposed solution is to add factor out the automatic outputs detection into a mixin.

The workaround can be removed from SMAP and COSMOS, and then compositor child classes that want to autodetect multiple outputs from sources can enable this feature as follows:

```
class MyCompositor(podpac.compositor.AutoOutputs, podpac.compositor.OrderedCompositor):
    ...
```

*Note: either this PR or #388 should be merged, but not both.*

### DIFF
--- a/doc/source/api.rst
+++ b/doc/source/api.rst
@@ -180,6 +180,7 @@ Stitch multiple data sources together
     podpac.compositor.OrderedCompositor
     podpac.compositor.UniformTileCompositor
     podpac.compositor.UniformTileMixin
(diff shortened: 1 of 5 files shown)


jmilloy:
We chose #388.

PR #389, opened 2020-04-15
node definition loops and caching

jmilloy:
* raise a `NodeDefinitionError` when a circular node definition is detected
 * raise a `NodeDefinitionError` when the node definition is accessed before the node traits are fully initialized
 * catch `NodeDefinitionError` in node caching methods and add a more informative message, including the `key`
 * cache put methods overwrite/update the cache by default
 * `LoadFileMixin` caches the `dataset` independent of the other node attrs (such as `interpolation` or `data_key`).
 * `Node.__init__` calls the `tl.HasTraits.__init__` *within* the `hold_trait_notifications` context in order to avoid calling some validation methods before other traits have even ben initialized.
 * CacheCtrl is cleaned up, with bugfixes and tests
 * small bugfixes in CSV and WCS nodes

The circular node definition handling prevents the cache deadlock. Then I had to change `LoadFileMixin` dataset caching in order to avoid hitting those exceptions, and it is better now. As a result, we do not have to remove the lock in `has_cache` if we don't want to.

Note also that I did not return `False` in `has_cache` when the `NodeDefinitionError` occurs, as we discussed. If that error is reached, the code trying to use the cache will *never* succeed. It is not like it is going to succeed in another state. The user should just not be caching that object, or needs to find another way/place to cache it. (Same for `put_cache`, etc).

Note also that I did not have to omit default traits from the node definition, so I did not yet add the podpac version to the definition.

The meat of this PR is in commit a425106.

### DIFF
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -580,7 +580,7 @@ class Node(tl.HasTraits):
         with thread_manager.cache_lock:
             return self.cache_ctrl.has(self, key, coordinates=coordinates)
 
-    def rem_cache(self, key, coordinates=None, mode=None):
+    def rem_cache(self, key, coordinates=None, mode="all"):
         """
         Clear cached data for this node.
 
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -78,6 +78,12 @@ class NodeException(Exception):
     pass
 
 
+class NodeDefinitionError(NodeException):
+    """ Raised node definition errors, such as when the definition is circular or is not yet unavailable. """
+
+    pass
+
+
 @common_doc(COMMON_DOC)
 class Node(tl.HasTraits):
     """The base class for all Nodes, which defines the common interface for everything.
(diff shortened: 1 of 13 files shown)


podpac/core/node.py
```diff
@@ -388,60 +401,72 @@ def definition(self):
...
+            definition = OrderedDict(zip(refs, definitions))
+            json.dumps(definition, cls=JSONEncoder)
+            return definition
 
-        # finalize, verify serializable, and return
-        definition = OrderedDict(zip(refs, definitions))
-        json.dumps(definition, cls=JSONEncoder)
-        return definition
+        finally:
+            self._definition_guard = False
```
mpu-creare, line 469: I didn't know you could have a 
```
try: 
    ...
    return 
finally:
    etc
``` 
and have etc actually run. Neat!

podpac/core/node.py
```diff
@@ -388,60 +401,72 @@ def definition(self):
             Dictionary-formatted node definition.
         """
 
+        if getattr(self, "_definition_guard", False):
+            raise NodeDefinitionError("node definition has a circular dependency")
+
+        if not getattr(self, "_traits_initialized_guard", False):
+            raise NodeDefinitionError("node is not yet fully initialized")
```
mpu-creare, line 408: Elegant solution. nicely done.

PR #391, opened 2020-04-16
ENH: Add podpac version to node definition.

### LINKED ISSUE #390: Add podpac version to node definition / hash
**Description**
Nodes may change between versions. Including the podpac version in the hash would avoid silent cache-related errors across upgrades.

The podpac version could be included in the definition as a node attr. It should be added explicitly when loading a JSON definition, since users are likely to omit it. Another option would be to include the version only when computing the hash.

**Additional context**
Avoiding computing default values for traits when accessing the cache would be helpful in some contexts. E.g. see #389. We have implemented, reverted, and discussed omitting default traits from the node definition in order to do this. There are drawbacks to that (less affordance for users, hacks required with traitlets containers such as List and Dict), but in addition, node defaults may change between versions. This would fix that issue.

jmilloy:
* Adds `"podpac_version"` to the node definition.
 * Warns about mismatching version when deserializing.
 * Omit hash (and style) from hash.

One question: since `__eq__` uses the hash, two nodes can be equal (as in `==`) even if their styles are different. I just want to check if that is what you want, or should equality depend on the json instead (which still includes the style)?

closes #390

### DIFF
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -392,7 +392,7 @@ class Node(tl.HasTraits):
 
         return d
 
-    @property
+    @cached_property
     def definition(self):
         """
         Full node definition.
@@ -487,7 +487,8 @@ class Node(tl.HasTraits):
     def hash(self):
         """ hash for this node, used in caching and to determine equality. """
 
-        d = self.definition
+        # deepcopy so that the cached definition property is not modified by the deletes below
+        d = deepcopy(self.definition)
 
         # omit version
         if "podpac_version" in d:
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -9,6 +9,7 @@ import functools
 import json
 import inspect
 import importlib
+import warnings
 from collections import OrderedDict
 from copy import deepcopy
 from hashlib import md5 as hash_alg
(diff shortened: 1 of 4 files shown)


mpu-creare:
I'm fine with that.

podpac/core/node.py
```diff
@@ -385,12 +387,12 @@ def _base_definition(self):
...
 
         # style
-        if self.style != Style() and self.style.definition:
+        if self.style.definition:
             d["style"] = self.style.definition
 
         return d
 
-    @cached_property
+    @property
```
mpu-creare, line 395: Why no longer a cached_property?

jmilloy: We use a private attribute in cached_property, so if you modify the dictionary that you get back, then the next time you access the `definition`, you will get the modified version!

That's a problem *whenever* you use private attributes behind properties, which is a common idiom, so I'm not exactly worried about in the general case. But in the hash property, I delete the `'podpac_version'` key and all of the `'style'` keys from the definition. So I had to either downgrade the cached_property to a property or deepcopy the definition before deleting keys. Either seems fine, really.

(Since the hash and the json are still cached_properties, it seems like we won't hit the definition by itself all that often. But since the hash is also a cached_property, deep copying the definition once to produce it wouldn't be so bad either.)

jmilloy: Switching to deepcopy

Issue #372, opened 2020-02-18
Algorithm Nodes should NOT cache outputs by default


mpu-creare:
**Description**
This is a big gotcha for new users. When developing a new algorithm node, if they (1) evaluate the results, (2) then change the code, and (3) evaluate the results again, nothing will change because the output is retrieved from the cache. 

**Steps to Reproduce**
```python
class MyAlg(Algorithm):
     data = NodeTrait()
     def algorithm(self, inputs):
         return inputs['data']
my_alg = MyAlg()
o = my_alg.eval(coords)

class MyAlg(Algorithm):
     data = NodeTrait()
     def algorithm(self, inputs):
         return inputs['data'] * 0 
my_alg = MyAlg()
o2 = my_alg.eval(coords)

assert np.all(o2.data != o.data)
```

This throws an assertion error .

**Expected Behavior**
I would expect o2 to be all zeros. 

**Observed Behavior**
o2 == o, since it comes from the cache. 

**Additional Notes**
There are multiple ways we could solve this problem. I think the best way is to make `cache_update` True by default for Algorithm and Compositor Nodes, and then in the documentation note how to optimize a pipeline that uses the same Algorithm/Compositor node in multiple places.

jmilloy:
I think we are going to set `cache_outputs = tl.Bool(default_value=False)` for Algorithm nodes, maybe for other nodes as well. The caching is an optimization that some user's will want to turn on.

PR #394, opened 2020-04-17
Feature/only cache data source outputs default

mpu-creare:
New users often get confused when developing new algorithms -- they make changes but get the same result. This is because the data is fetched from the cache (non-unique pipeline definition). Based on #372 we decided to change the default behavior for most nodes.

### DIFF
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -97,8 +97,9 @@ class Node(tl.HasTraits):
         (CACHE_NODE_OUTPUT_DEFAULT for general Nodes, and CACHE_DATASOURCE_OUTPUT_DEFAULT  for DataSource nodes). 
         If True, outputs will be cached and retrieved from cache. If False, outputs will not be cached OR retrieved from cache (even if 
         they exist in cache). 
-    cache_update: bool
-        Default is True. Should the node's cached output be updated from the source data?
+    force_eval: bool
+        Default is False. Should the node's cached output be updated from the source data? If True it ignores the cache
+        when computing outputs but puts results into the cache (thereby updating the cache)
     cache_ctrl: :class:`podpac.core.cache.cache.CacheCtrl`
         Class that controls caching. If not provided, uses default based on settings.
     dtype : type
@@ -129,7 +130,7 @@ class Node(tl.HasTraits):
 
     dtype = tl.Any(default_value=float)
     cache_output = tl.Bool()
-    cache_update = tl.Bool(False)
+    force_eval = tl.Bool(False)
     cache_ctrl = tl.Instance(CacheCtrl, allow_none=True)
 
     # list of attribute names, used by __repr__ and __str__ to display minimal info about the node
@@ -972,7 +973,7 @@ def node_eval(fn):
         key = cache_key
         cache_coordinates = coordinates.transpose(*sorted(coordinates.dims))  # order agnostic caching
 
-        if not self.cache_update and self.cache_output and self.has_cache(key, cache_coordinates):
+        if not self.force_eval and self.cache_output and self.has_cache(key, cache_coordinates):
             data = self.get_cache(key, cache_coordinates)
             if output is not None:
                 order = [dim for dim in output.dims if dim not in data.dims] + list(data.dims)
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -93,7 +93,10 @@ class Node(tl.HasTraits):
     Attributes
     ----------
     cache_output: bool
-        Should the node's output be cached? If not provided or None, uses default based on settings.
+        Should the node's output be cached? If not provided or None, uses default based on settings 
+        (CACHE_NODE_OUTPUT_DEFAULT for general Nodes, and CACHE_DATASOURCE_OUTPUT_DEFAULT  for DataSource nodes). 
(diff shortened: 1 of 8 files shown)


podpac/core/node.py
```diff
@@ -969,7 +972,7 @@ def wrapper(self, coordinates, output=None):
         key = cache_key
         cache_coordinates = coordinates.transpose(*sorted(coordinates.dims))  # order agnostic caching
 
-        if not self.cache_update and self.has_cache(key, cache_coordinates):
+        if not self.cache_update and self.cache_output and self.has_cache(key, cache_coordinates):
```
mpu-creare, line 975: @jmilloy What do you think about this addition. Make sense? I adjusted the documentation to reflect this. 

Basically, if a user realizes they have cache_output = True and then turn it off, the would STILL get cached results WITHOUT this addition.

jmilloy: Makes sense.

Another option would be to observe `cache_output` and remove the output from the cache when it is changed to False. Without that, a user could turn cache_output off, then turn it back on and get the old results again.

BTW, I think `cache_update` is confusing. I think you should consider renaming it something like `force_eval` or `force_reevaluate` or `force_ignore_cached_output`.

jmilloy: Do you want to set cache_dataset in LoadFileMixin to False by default, also?

Either way, we should fix the has_cache check there in the same way:

```diff
diff --git a/podpac/core/data/file_source.py b/podpac/core/data/file_source.py
index 2749da1..2f38932 100644
--- a/podpac/core/data/file_source.py
+++ b/podpac/core/data/file_source.py
@@ -88,7 +88,7 @@ class LoadFileMixin(S3Mixin):
     @cached_property
     def dataset(self):
         # use the _dataset_caching_node "stub" here because the only node attr we care about is the source
-        if self._dataset_caching_node.has_cache(key="dataset"):
+        if self.cache_dataset and self._dataset_caching_node.has_cache(key="dataset"):
             data = self._dataset_caching_node.get_cache(key="dataset")
             with BytesIO(data) as f:
                 return self._open(BytesIO(data), cache=False)
```

mpu-creare: why is dataset even using the has_cache? don't have you have to set `use_cache_ctrl` on the `cached_property` decorator to use the cache?

mpu-creare: > BTW, I think `cache_update` is confusing. I think you should consider renaming it something like `force_eval` or `force_reevaluate` or `force_ignore_cached_output`.

I agree completely. I keep having a hard time finding this setting. The idea was that all the cache-related attributes would start with `cache`... but I usually end uplooking for `force_eval`.

jmilloy: This one is a special case because it has to cache the binary representation instead of the object itself. So it does the has_cache and put_cache itself. I think it would fail to cache the object in some cases if you use `use_cache_ctrl=True` directly.

Actually, it is now *also* special now because I am using a stub node (with only the source attribute but no data_key or outputs) to do the caching of the dataset.

jmilloy: `cache_force_eval`?

mpu-creare: > Another option would be to observe `cache_output` and remove the output from the cache when it is changed to False. Without that, a user could turn cache_output off, then turn it back on and get the old results again.

i think in this case the user should set `update_cache` (or `force_eval`). I wouldn't want to remove something from the cache without the user's permission.

jmilloy: > i think in this case the user should set update_cache (or force_eval). I wouldn't want to remove something from the cache without the user's permission.

okay, makes sense

jmilloy: then again, by that reasoning, you could remove `cache_output` altogether, and just use `force_eval` from the get-go...

mpu-creare: > Another option would be to observe `cache_output` and remove the output from the cache when it is changed to False. Without that, a user could turn cache_output off, then turn it back on and get the old results again.

i think in this case the user should set `update_cache` (or `force_eval`). I wouldn't want to remove something from the cache without the user's permission. 




> 
> 
> > why is dataset even using the has_cache?
> 
> This one is a special case because it has to cache the binary representation instead of the object itself. So it does the has_cache and put_cache itself. I think it would fail to cache the object in some cases if you use `use_cache_ctrl=True` directly.
> 
> Actually, it is now _also_ special now because I am using a stub node (with only the source attribute but no data_key or outputs) to do the caching of the dataset.

Got it. I missed that the `put_cache` was in the `open` method. In that case yes, I think it should be `False` by default. In fact, I believe I've run into this problem where I get the cached dataset instead of the source dataset (which has changed during the session and I'm opening it again to check it!). I ended up just restarting my kernel -- thought it was Zarr caching issue...

Thanks for pointing these out @jmilloy

mpu-creare: > 
> 
> then again, by that reasoning, you could remove `cache_output` altogether, and just use `force_eval` from the get-go...

Let's see, right now `force_eval=True` will ignore the cache, but NOT update the output unless  `cache_output=True`. I think that's useful in case you want to do an experiment without losing what you have in the cache... subtle though... I'll have to think about that.

PR #395, opened 2020-04-17
Move segments from coordinates to datasource

jmilloy:
### Summary

 * coordinates `ctype` and `segment_lengths` are replaced by DataSource `boundary`
 * coordinates `area_bounds` is removed
 * rasterio reprojection bounds (and WCS request bounds) are calculated from the coordinates `start`, `stop`, and `step`
 * coordinates `geotransform` supports uniform ArrayCoordinates1d

### Datasource coordinates boundary

The Datasource `boundary` is a dictionary of coordinate boundaries (one for each dimension). Point coordinates are omitted or None. The trait is validated and tested, but it is unused at this point and raises NotImplementedError in some cases. Note that all of these are coordinate offsets.

```
# simple centered boxes/ranges
boundary={'lat': 0.5, 'lon': 0.5}
boundary={'time': "1,D"}

# simple non-centered boxes/ranges
boundary={'lat': [0, 0.5], 'lon': [0, 0.5]}
boundary={'time': ["0,D", "1,D"]}

# polygons 
boundary={'lat': [0.0, -0.5, 0.0, 0.5], 'lon': [-0.5, 0.0, 0.5, 0.0]}

# array of boundaries, specified for each coordinate
boundary={"lat": [[-0.1, 0.4], [-0.2, 0.3], [-0.3, 0.2]], "lon": [[-0.1, 0.4], [-0.2, 0.3]]}
```

### get_area_bounds

Area bounds are not currently used anywhere, but the logic to compute area bounds from coordinates and a boundary has been updated and may be useful. See `get_area_bounds` and associated tests. It supports all of the above boundary types.


mpu-creare:
#396 touches segments / ctypes in a few minor places. I think we should merge that FIRST, as this one probably has a lot of changes / updates.

jmilloy:
I admit that I only have a cursory understanding, but I think that this is revealing to me the extent to which we are mis-using rasterio reprojection to do interpolation and getting lucky that our defaults usually work.

Basically, we infer the rasterio transform bounds from the coordinates *within* the coordinates object via the inferred segment_lengths. If the segment lengths cannot be inferred, or if the user supplies their own segment_lengths, the rasterio "interpolation" will be incorrect or raise an exception. (This is why you get an exception when you request coordinates with only one value in either dimension: we cannot infer the segment_lengths, so we silently fall back on the bounds which make an invalid geotransform because the high and low bonud is the same!).

I think I can make it work by inferring the area_bounds in the rasterio interpolate method, which will make everything a lot clearer. It will be clear what's actually going on, and I will be able to raise the right exceptions at that time.

Note that nothing else uses the soon-to-be-removed area_bounds. I believe that rasterio is unable to handle anything but centered rectangular coordinates, and that non-default coordinate boundaries (e.g. ctype left or non-uniform segment_lengths) must use a different interpolator.

mpu-creare:
See if the new `Coordinates.geotransform` will serve the needed purpose in the rasterio interpolator. 

Perhaps we can return `False` for the `rasterio` interpolator in cases where `area_bounds` cannot be inferred.

jmilloy:
> Perhaps we can return False for the rasterio interpolator in cases where area_bounds cannot be inferred.

Yes, the can_interpolate method is doing a better job that I realized.

jmilloy:
> See if the new Coordinates.geotransform will serve the needed purpose in the rasterio interpolator.

I haven't looked closely, but it is giving be the exception when I don't think it should.

```
self.coordinates.geotransform
*** TypeError: Only 2-D coordinates that are uniform or rotated have a GDAL transform. These coordinates Coordinates (EPSG:4326)
	lat: ArrayCoordinates1d(lat): Bounds[1.0, 3.0], N[3], ctype['midpoint']
	lon: ArrayCoordinates1d(lon): Bounds[1.0, 3.0], N[3], ctype['midpoint'] do not.
```

jmilloy:
Ahh, you check `isinstance(self._coord['lat'], UniformCoordinats1d)` when you could check `self._coord[`lat`].is_uniform`.

podpac/core/interpolation/interpolators.py
```diff
@@ -270,25 +270,16 @@ def interpolate(self, udims, source_coordinates, source_data, eval_coordinates,
...
+            east = c["lon"].bounds[1] + c["lon"].step / 2.0
+            south = c["lat"].bounds[0] - c["lat"].step / 2.0
+            north = c["lat"].bounds[1] + c["lat"].step / 2.0
+            cols = c["lon"].size
+            rows = c["lat"].size
             return transform.from_bounds(west, south, east, north, cols, rows)
 
         with rasterio.Env():
+            # src_transform = transform.Affine.from_gdal(*source_coordinates.geotransform)
             src_transform = get_rasterio_transform(source_coordinates)
```
jmilloy, line 283: @mpu-creare This is working, but can you show me how to use the `geotransform` here? Is it possible?

podpac/core/coordinates/coordinates.py
```diff
@@ -879,9 +863,9 @@ def geotransform(self):
         # Do the uniform coordinates case
         if (
             "lat" in self.dims
-            and isinstance(self._coords["lat"], UniformCoordinates1d)
             and "lon" in self.dims
-            and isinstance(self._coords["lon"], UniformCoordinates1d)
+            and self._coords["lat"].is_uniform
+            and self._coords["lon"].is_uniform
```
jmilloy, line 868: @mpu-creare I added start, step, and stop to ArrayCoordinates1d so now this works for array coordinates, too.

podpac/core/coordinates/coordinates1d.py
```diff
@@ -293,6 +194,56 @@ def copy(self):
...
+            L, H = self.argbounds
+            lo_offset = min(make_coord_delta_array(boundary[L]))
+            hi_offset = max(make_coord_delta_array(boundary[H]))
+
+        lo, hi = self.bounds
+        lo = add_coord(lo, lo_offset)
+        hi = add_coord(hi, hi_offset)
+
+        return lo, hi
+
```
jmilloy, line 245: @mpu-creare Not used, but this is how the boundaries are meant to be interpreted. I think it's pretty clean.

podpac/core/coordinates/test/test_array_coordinates1d.py
```diff
@@ -742,41 +486,66 @@ def test_index(self):
...
+    def test_get_area_bounds_overlapping(self):
+        values = np.array([0.0, 1.0, 4.0, 6.0])
+        c = ArrayCoordinates1d(values)
 
-        # inferred segment_lengths
-        c = ArrayCoordinates1d([1, 2, 4, 7], ctype="left")
-        c2 = c[1]
-        assert c2.segment_lengths == 2.0 or np.array_equal(c2.segment_lengths, [2.0])
+        area_bounds = c.get_area_bounds([[-0.1, 0.1], [-10.0, 10.0], [-0.1, 0.1], [-0.1, 0.1]])
+        assert_equal(area_bounds, [-11.0, 11.0])
```
jmilloy, line 548: @mpu-creare And the tests for get_area_bounds, showing `boundary` in use. Again, I think this is pretty clean.

podpac/core/data/datasource.py
```diff
@@ -437,6 +465,9 @@ def eval(self, coordinates, output=None):
                     + "request Coordinates coordinate reference system ({})".format(coordinates.crs)
                 )
 
+        # get indexed boundary
+        self._requested_source_boundary = self.get_boundary(self._requested_source_coordinates_index)
+
```
jmilloy, line 470: This call (and the get_boundary method) is only necessary for arrays of non-uniform boundaries. And since we aren't actually using the boundary anywhere in the interpolation, it isn't necessary at all. But I didn't want to forget about it. Otherwise, the datasource eval method is fundamentally unchanged.

podpac/core/interpolation/interpolators.py
```diff
@@ -270,25 +270,16 @@ def interpolate(self, udims, source_coordinates, source_data, eval_coordinates,
...
-            west, east = c["lon"].area_bounds
-            south, north = c["lat"].area_bounds
-            cols, rows = (c["lon"].size, c["lat"].size)
-            # print (east, west, south, north)
+            west = c["lon"].bounds[0] - c["lon"].step / 2.0
+            east = c["lon"].bounds[1] + c["lon"].step / 2.0
+            south = c["lat"].bounds[0] - c["lat"].step / 2.0
+            north = c["lat"].bounds[1] + c["lat"].step / 2.0
+            cols = c["lon"].size
+            rows = c["lat"].size
```
jmilloy, line 278: @mpu-creare This is what we would replace `c.geotransform` if possible, but I couldn't figure out how exactly.

jmilloy:
@mpu-creare This is all set. Note that the interpolation tests are completed unchanged, I just broke them up into three files. Let me know if you to change anything in the boundary spec (it is similar but I think better to what we discussed in #395).

mpu-creare:
> 
> 
> Ahh, you check `isinstance(self._coord['lat'], UniformCoordinats1d)` when you could check `self._coord[`lat`].is_uniform`.

OH! That's good to know! I might have to fix that in #396 as well.

podpac/core/compositor/tile_compositor.py
```diff
@@ -89,7 +89,7 @@ def tile_width(self):
 
 
 @common_doc(COMMON_DATA_DOC)
-class UniformTileMixin(podpac.Node):
+class UniformTileMixin(tl.HasTraits):
```
mpu-creare, line 92: What is the rule for `Mixins`? Is there any harm inheriting from `Node`? Why not inherit from `object` then? (probably need the `traitlets` metaclass functionality I'd reckon. )

podpac/core/coordinates/coordinates1d.py
```diff
@@ -293,6 +193,56 @@ def copy(self):
...
 
+    def get_area_bounds(self, boundary):
+        """
+        Get low and high coordinate area bounds.
+
+        Arguments
+        ---------
+        boundary : float, timedelta, array, None
+            Boundary offsets in this dimension.
+            * For a centered uniform boundary, use a single positive float or timedelta offset
```
mpu-creare, line 204: should this be step / 2 ?

mpu-creare: otherwise in the offsets below you have to divide boundary by 2

podpac/core/coordinates/test/test_array_coordinates1d.py
```diff
@@ -742,41 +486,66 @@ def test_index(self):
...
-        c2 = c[[]]
-        assert_equal(c2.segment_lengths, [])
+        # segment
+        area_bounds = c.get_area_bounds([-0.2, 0.7])
+        assert_equal(area_bounds, [-0.2, 6.7])
 
-        # uniform segment_lengths
-        c = ArrayCoordinates1d([1, 2, 4, 5], segment_lengths=0.5)
+        # polygon (i.e. there would be corresponding offets for another dimension)
+        area_bounds = c.get_area_bounds([-0.2, -0.5, 0.7, 0.5])
```
mpu-creare, line 508: I don't quite understand this one...

podpac/core/coordinates/test/test_array_coordinates1d.py
mpu-creare: That's pretty good. I'm impressed you covered the overlapping case.

podpac/core/data/datasource.py
```diff
@@ -369,12 +399,11 @@ def eval(self, coordinates, output=None):
...
             coordinates = coordinates.transform(self.coordinates.crs)
 
-        # intersect the coordinates with requested coordinates
-        # to get coordinates within requested coordinates bounds
+        # intersect the coordinates with requested coordinates to get coordinates within requested coordinates bounds
         # TODO: support coordinate_index_type parameter to define other index types
-        (self._requested_source_coordinates, self._requested_source_coordinates_index) = self.coordinates.intersect(
-            coordinates, outer=True, return_indices=True
-        )
+        (rsc, rsci) = self.coordinates.intersect(coordinates, outer=True, return_indices=True)
```
mpu-creare, line 404: Does this need to know the boundary data?

podpac/core/interpolation/interpolators.py
mpu-creare: This is fine. As a one-liner, it would have been 
```python
return rasterio.Affine.from_gdal(*c.geotransform)
```
But this is pretty clear in case of future debugging...

podpac/core/coordinates/coordinates1d.py
mpu-creare: I get it down. I'll make the documentation a little clearer.

podpac/core/compositor/tile_compositor.py
jmilloy: I don't know exactly. I think you're right, object is probably better. I'm not sure if the traits will work without subclassing tl.HasTraits here, but it might because it is always mixed with a different class that does inherit from tl.HasTraits.

jmilloy: I don't remember why I changed it.

podpac/core/coordinates/test/test_array_coordinates1d.py
jmilloy: I'm not sure if we will ever use this one, but it is very general.

For a centered square, you could use `boundary = {'lat': [-1.0, -1.0, 1.0, 1.0], 'lon': [-1.0, 1.0, -1.0, 1.0]}`. So for lat, you get boundary offsets `[-1.0, -1.0, 1.0, 1.0]`, which appear a bit meaningless at that point.

For area bounds, all you care about is the min and max offset, and currently Coordinates1d `get_area_bounds` does that conversion (from boundary offsets to area bounds offsets). I guess it could be more clear if full Coordinates `get_area_bounds` method did the conversion, e.g:

```
class Coordinates:
    def get_area_bounds(self, boundary):
        offsets = {}
        for dim, value in boundary.items():
            if np.ndarray(vaue).ndim == 0:
               value = make_coord_delta(value)
               offsets[dim] = [-value, value]
            elif np.ndarray(value).ndim == 1:
                value = make_coord_delta_array(value)
                offsets[dim] = [min(value), max(value)]
            else:
                value = np.array([make_coord_delta_array(elem) for elem in value])
                offsets[dim] = np.array(np.min(value, axis=1), np.max(value, axis=1)).T
                
        return {dim: self[dim].get_area_bounds(offsets.get(dim)) for dim in self.udims}
```

And then Coordinates1d.get_area_bounds would expect offsets, and be a bit clearer:

```
class Coordinates1d:
    def get_area_bounds(self, offsets):
        # point coordinates
        if offsets is None:
            return self.bounds

        # empty coordinates
        if self.size == 0:
            return self.bounds

        if np.array(offsets).ndim == 1:
            # uniform boundary
            lo_offset, hi_offset = offsets
        else:
            # non-uniform boundary
            L, H = self.argbounds
            lo_offset = offsets[L][0]
            hi_offset = offsets[H][1]

        lo, hi = self.bounds
        lo = add_coord(lo, lo_offset)
        hi = add_coord(hi, hi_offset)

        return lo, hi
```

That is a little bit less efficient, but maybe worth it.

jmilloy: I'm not sure when/where we will need the area_bounds though.

podpac/core/data/datasource.py
jmilloy: I passed the boundary through at one point, but it does not seem to need it. Maybe as an optimization.

jmilloy:
> `return rasterio.Affine.from_gdal(*c.geotransform)`

I think you can see that I tried that and left it commented it out. It is not the same and doesn't work.

jmilloy:
Okay fixed that typo in the tests.

mpu-creare:
> 
> 
> > `return rasterio.Affine.from_gdal(*c.geotransform)`
> 
> I think you can see that I tried that and left it commented it out. It is not the same and doesn't work.

Yeah, I'll need to look into that because it should work. They are doing the same thing, or should be. This means one of the two are wrong.

podpac/core/compositor/tile_compositor.py
mpu-creare: I just did a quick test. Looks like inheriting from `object` works well. 

```python
import traitlets as tl
class Mix(object):
    l = tl.List()

class T(tl.HasTraits):
    u = tl.Unicode()

class G(Mix, T):
 pass

g = G()
g.l
[]
g.trait_names()
['l', 'u']
```

I don't think I'll change it... we could do a PR to change all Mixin's to `object`, but I'd rather understand it better before doing something like that. This was just for interest.

podpac/core/coordinates/test/test_array_coordinates1d.py
mpu-creare: Ah, this is great for future. Nice. Good spec.

mpu-creare:
> 
> 
> > > `return rasterio.Affine.from_gdal(*c.geotransform)`
> > 
> > 
> > I think you can see that I tried that and left it commented it out. It is not the same and doesn't work.
> 
> Yeah, I'll need to look into that because it should work. They are doing the same thing, or should be. This means one of the two are wrong.

Yeah, I see the issue. The `coordinates.geotransform ` "naturally" handles the order of the dimensions (`is_descending`). For example, these give the same results:
```python
coords_src = Coordinates([clinspace(10, 0, 3), clinspace(0, 10, 5)], dims=["lat", "lon"])
rasterio.Affine.from_gdal(*coords_src.geotransform)
Affine(2.5, 0.0, -1.25,
       0.0, -5.0, 12.5)
coords_src = Coordinates([clinspace(0, 10, 3), clinspace(0, 10, 5)], dims=["lat", "lon"])
get_rasterio_transform(coords_src)
Affine(2.5, 0.0, -1.25,
       0.0, -5.0, 12.5)
```
But this does not give the correct answer (since step is < 0 and is not accounted for in `get_rasterio_transform`
```python
coords_src = Coordinates([clinspace(10, 0, 3), clinspace(0, 10, 5)], dims=["lat", "lon"])
get_rasterio_transform(coords_src)
Affine(2.5, 0.0, -1.25,
       0.0, -1.6666666666666667, 7.5)
```

So there is a small bug there... I'll fix it.

jmilloy:
> > > > `return rasterio.Affine.from_gdal(*c.geotransform)`
> > > 
> > > 
> > > I think you can see that I tried that and left it commented it out. It is not the same and doesn't work.
> > 
> > 
> > Yeah, I'll need to look into that because it should work. They are doing the same thing, or should be. This means one of the two are wrong.
> 
> Yeah, I see the issue. The `coordinates.geotransform ` "naturally" handles the order of the dimensions (`is_descending`). For example, these give the same results:
> 
> ```python
> coords_src = Coordinates([clinspace(10, 0, 3), clinspace(0, 10, 5)], dims=["lat", "lon"])
> rasterio.Affine.from_gdal(*coords_src.geotransform)
> Affine(2.5, 0.0, -1.25,
>        0.0, -5.0, 12.5)
> coords_src = Coordinates([clinspace(0, 10, 3), clinspace(0, 10, 5)], dims=["lat", "lon"])
> get_rasterio_transform(coords_src)
> Affine(2.5, 0.0, -1.25,
>        0.0, -5.0, 12.5)
> ```
> 
> But this does not give the correct answer (since step is < 0 and is not accounted for in `get_rasterio_transform`
> 
> ```python
> coords_src = Coordinates([clinspace(10, 0, 3), clinspace(0, 10, 5)], dims=["lat", "lon"])
> get_rasterio_transform(coords_src)
> Affine(2.5, 0.0, -1.25,
>        0.0, -1.6666666666666667, 7.5)
> ```
> 
> So there is a small bug there... I'll fix it.
 
Aha! Thanks.

jmilloy:
Well, instead of fixing the bug in get_rasterio_transform, I would prefer to use geotransform here if that works. I feel like when I tried it, the tests didn't pass... maybe even threw an exception, but I could be wrong.

mpu-creare:
> Well, instead of fixing the bug in get_rasterio_transform, I would prefer to use geotransform here if that works. I feel like when I tried it, the tests didn't pass... maybe even threw an exception, but I could be wrong.

Yep, that's the plan! I think tests are now passing...

PR #396, opened 2020-04-18
Feature/uniform coords transform to uniform

mpu-creare:
This simplified any transformed coordinates to independent / array / uniform coordinates. 

This only one missing is stacked array to stacked uniform (a pretty rare case). 

@jmilloy please have a quick look in case I missed something. It's a little convoluted so I want to document that code a little better, but beyond that I think it's pretty good.


jmilloy:
I will review this afternoon.

podpac/core/coordinates/dependent_coordinates.py
```diff
@@ -513,48 +513,26 @@ def _transform(self, transformer):
...
+                continue
+
+            i = self.dims.index(dim)
+            slc = slc_start.copy()
+            slc[i] = slice(None)
+            if np.any(np.abs(coords[i][tuple(slc)] - coords[i]) >= TOL):  # check indepedent
+                return self
+            coords[i] = ArrayCoordinates1d(coords[i][tuple(slc)].squeeze(), name=dim).simplify()
+
+        return coords
```
jmilloy, line 535: I think the simplify method could be useful in general. What do you think?

(Looking at this now, maybe I shouldn't skip time. I'm not sure why I did that.)

Are you interested in something similar in the `Coordinates` class? I haven't looked at that very closely yet.

podpac/core/coordinates/array_coordinates1d.py
```diff
@@ -233,6 +233,29 @@ def copy(self):
...
+
+        TOL = 1e-7
+        dcrds = np.diff(self.coordinates)
+        if np.all(np.abs(dcrds - dcrds[0]) < np.abs(TOL * dcrds[0])):
+            return UniformCoordinates1d(
+                self.coordinates[0], self.coordinates[-1], self.coordinates[1] - self.coordinates[0], **self.properties
+            )
+
+        return self
+
```
jmilloy, line 258: @mpu-creare This is basically `make_coords` from before. Two things:

1. We will be able to use `self.start, self.stop, self.step` after merging #395 
2. We should be able to use `is_uniform` for the check, but that doesn't currently have a tolerance. I think we should add a tolerance to the `is_uniform` check for array coordinates. What do you think?
2b. I bet this check doesn't work for time coordinates.

podpac/core/coordinates/dependent_coordinates.py
jmilloy: @mpu-creare Okay I think it may never happen but I pushed an update to handle dependent time coordinates as well. The other option would be just to *not* simplify if there are time coordinates.

mpu-creare: Yeah, I agree. I considered factoring it out, but I wasn't sure where to put it, and i was rushing as usual.

podpac/core/coordinates/array_coordinates1d.py
mpu-creare: I agree with a tolerance on 2a. 
I wouldn't be surprised if it doesn't work for time -- future optimization. :)

jmilloy: ArrayCoordinates1d is_uniform uses allclose, so it already has a tolerance. I used allclose in DependentCoordinates simplify, also.

mpu-creare: The only thing to be careful of there is conversion between float32 and float64. Files often are saved at float32, and then loaded into memory as float32. I'm not sure `all_close` can detect between those two cases...

jmilloy: Okay, I added `atol=1e-7`.

jmilloy: Should I also use `rtol=0`?

jmilloy: > Should I also use rtol=0?

The answer is yes.

jmilloy:
Can you help me find two crs that that transform spatially uniform -> dependent but that I can add vunits to?

For example, `+proj=merc +vunits=ft` -> `+proj=utm +vunits=m` works, but the transformation does not require dependent lat, lon coordinates. Alternatively, `EPSG:4326` -> `EPSG:32629` requires dependent lat, lon coordinates, but `EPSG:4326` uses `+proj=latlon` and you can't seem to add vunits to that.

podpac/core/coordinates/coordinates.py
```diff
@@ -1324,66 +1324,175 @@ def transform(self, crs=None):
...
+                [
+                    [
+                        self["alt"].coordinates,
+                        self["alt"].coordinates * 0 + self["lat"].coordinates.mean(),
+                        self["alt"].coordinates * 0 + self["lon"].coordinates.mean(),
+                    ]
+                ],
+                ["alt_lon_lat"],
+                crs=self.crs,
+            )
```
jmilloy, line 1486: What happens if you just do the altitude transform by itself, e.g.

```
            temp_coords = Coordinates([[self["alt"].coordinates]], ["alt_lon_lat"], crs=self.crs)
            t_alt = temp_coords.transform(crs)["alt"]
```

jmilloy:
Pending tests this is ready to merge.

PR #397, opened 2020-04-20
Simplify coordinates transform by letting altitude coordinates themselves.

jmilloy:
Tests still pass. Am I missing a case if I do this?

I like that there isn't an early return case.

This takes advantage of the fact that the `Coordinates1d._transform` method ignores 'lat' and 'lon' coordinates, assuming they have already been transformed. That was something I did in order to make that `for c in cs` clause seamless when I first implemented it.

### DIFF
--- a/podpac/core/coordinates/coordinates.py
+++ b/podpac/core/coordinates/coordinates.py
@@ -1379,6 +1379,8 @@ class Coordinates(tl.HasTraits):
                 cs.insert(i, c)
 
         # transform
+        # note that unstacked lat and lon coordinates are ignored in the _transform method
+        # so the simplified transform above will be preserved
         ts = []
         for c in cs:
             tc = c._transform(transformer)
--- a/podpac/core/coordinates/coordinates.py
+++ b/podpac/core/coordinates/coordinates.py
@@ -1355,30 +1355,28 @@ class Coordinates(tl.HasTraits):
             if tc:
                 cs[self.dims.index("lat")] = tc[0]
                 cs[self.dims.index("lon")] = tc[1]
-                if "alt" in self.dims:
-                    cs[self.dims.index("alt")] = tc[2]
-                return Coordinates(cs, crs=crs, validate_crs=False)
 
             # otherwise convert lat-lon to dependent coordinates
-            ilat = self.dims.index("lat")
-            ilon = self.dims.index("lon")
-            if ilat == ilon - 1:
-                c1, c2 = self["lat"], self["lon"]
-            elif ilon == ilat - 1:
-                c1, c2 = self["lon"], self["lat"]
+            else:
+                ilat = self.dims.index("lat")
+                ilon = self.dims.index("lon")
+                if ilat == ilon - 1:
+                    c1, c2 = self["lat"], self["lon"]
+                elif ilon == ilat - 1:
+                    c1, c2 = self["lon"], self["lat"]
 
-            c = DependentCoordinates(
-                np.meshgrid(c1.coordinates, c2.coordinates, indexing="ij"),
-                dims=[c1.name, c2.name],
-                ctypes=[c1.ctype, c2.ctype],
-                segment_lengths=[c1.segment_lengths, c2.segment_lengths],
-            )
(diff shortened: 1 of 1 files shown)


mpu-creare:
What happens if you have a coordinate system that is essentially a curve through `lat`, `lon`, `alt`? I think you're assuming that the coordinate systems will always be orthogonal, which is not true in general.

mpu-creare:
I realize we don't have a usecase for this... but I'm anticipating one in an upcoming project.

jmilloy:
Okay, that is the case I was thinking about. I am only aware of this point of defining the altitude units orthogonally to the spatial crs using vunits, so I wasn't sure if such a case actually existed.

mpu-creare:
In fairness, I'm not sure pyproj/proj4 can handle such a case...

jmilloy:
Yeah, I was going to asking if there is an existing crs on spatialreference.org or at least a way to create such a crs in proj or wkt. If there isn't, then no reason to support it... but maybe you have another idea?

jmilloy:
Another factor is that I think the code as is (without this change) *only* handles this case if the coordinates can be simplified. There are two non-simplified cases: we handle the lat-lon dependent and alt orthogonal case, but we don't handle the lat-lon-alt all dependent case. We would need to distinguish between the two cases, check that *all three* dimensions are present, and add it to the newly created dependent coordinates.

I'm strongly opposed to adding features that we don't use, *especially* if they are only partially implemented anyways. So I'm probing this pretty hard.

jmilloy:
Okay, thinking about the three cases carefully. The cases are **lat, lon, alt** (all indpendent), **lat-lon, alt** (lat-lot dependent, alt independent), and **lat-lon-alt** (all dependent). If I have this right, the **lat+lon+alt** case is not handled correctly (italics) in either version! 

### Currently

 * **lat, lon, alt**
     - Coordinates.simplify detects that lat, lon, and alt are indepedent and transforms them

 * **lat+lon, alt**
     * Coordinates.simplify detects that lat, lon, and alt are not all independent
     * lat and lon are converted to DependentCoordinates
     * in the final loop: lat+lon are transformed together, alt is transformed indpendently

 * **lat+lon+alt**
     * Coordinates.simplify detects that lat, lon, and alt are not all independent
     * lat and lon are converted to DependentCoordinates
     * *in the final loop: lat+lon are transformed together, alt is transformed independently*

### Proposal (this branch)

 * **lat, lon, alt**
     * Coordinates.simplify detects that lat and lon are indepedent and transforms them
     * in the final loop: alt is transformed indepedently

* **lat+lon, alt**
     * Coordinates.simplify detects that lat and lon are not indpendent
     * lat and lon are converted to DependentCoordinates
     * in the final loop: lat+lon are transformed together, alt is transformed indpendently

* **lat+lon+alt**
     * Coordinates.simplify detects that lat and lon are not indepedent
     * lat and lon are converted to DependentCoordinates
     * *in the final loop: lat+lon are transformed together, alt is transformed indpendently*

---

The solution would be to put the 'alt' into the DependentCoordinates if necessary so that it can be transformed along with the lat and lon. I don't think that the DependentCordinates simplify transform correctly handles the **lat+lon, alt** case (more on that below), so we would have to do some work to detect the different cases so that we didn't put it in unnecessarily. Plus, that the efficient way. The other (inefficient) way would be to just always put the alt in and fix DependentCoordinates simplify to handle that case.

So, I think that the DependentCoordinates simplify_transform only simplifies if *all* of the spatial dimensions are independent. I bet that `test_transform_uniform_to_dependent_to_uniform` won't works if altitude is included. I tried to check that real quick but the to_dependent part of that didn't even seem to work correctly. So we actually do need to fix that, and then maybe it's worth keeping the dependent lat+lon+alt handling because it would actually work...

mpu-creare:
> I'm strongly opposed to adding features that we don't use, _especially_ if they are only partially implemented anyways. So I'm probing this pretty hard.

That's fair. It's also probably a very rare case. please disregard this case. 

Thanks for taking this on and being careful about it.

PR #401, opened 2020-04-21
Release/2.0.0

mpu-creare:
2.0.0 release.


podpac/core/coordinates/array_coordinates1d.py
```diff
@@ -88,7 +88,7 @@ def __init__(self, coordinates, name=None):
                 if self._is_uniform:
```
mpu-creare, line 88: @jmilloy 
on line 77, is that supposed to be a "divided by" instead of a * (c[1] - c[0]) ? Because right now smaller extents will have smaller deltas and bigger extents will have bigger deltas.

mpu-creare: I believe it should be 
```
 deltas = (self.coordinates[1:] - self.coordinates[:-1]).astype(float) / (
                self.coordinates[1] - self.coordinates[0]
            ).astype(float) * self.coordinates.size
```
which would scale the deltas to [0, 1], roughly.

jmilloy: The deltas is only used to determine first if it's sorted and second if it's uniform. So it can be whatever works best... it makes sense to me that scaling to [0-1] would be more numerically stable, but I think it could also be `* np.sign(self.coordinates[1] - self.coordinates[0])` so that it doesn't scale at all.

jmilloy: It could also be without any scaling or even sign manipulation:

```
deltas = (self.coordinates[1:] - self.coordinates[:-1]).astype(float)
if not (np.all(deltas <= 0) or np.all(delta >= 0))
```

I would either use this or the np.sign version.

mpu-creare: Got it. I didn't realize the sign was the issue. Thanks!

PR #414, opened 2020-07-08
FIX: Adding issubset function for Dependent Coordinates

mpu-creare:
It was missing and breaks the Sentinel datalib (and possibly others).

@jmilloy Would you mind quickly reviewing this -- I want to merge it in for the newest release.

### DIFF
--- a/podpac/core/coordinates/dependent_coordinates.py
+++ b/podpac/core/coordinates/dependent_coordinates.py
@@ -8,6 +8,7 @@ import traitlets as tl
 import lazy_import
 from six import string_types
 import numbers
+import logging
 
 from podpac.core.settings import settings
 from podpac.core.utils import ArrayTrait, TupleTrait
@@ -18,6 +19,8 @@ from podpac.core.coordinates.coordinates1d import Coordinates1d
 from podpac.core.coordinates.array_coordinates1d import ArrayCoordinates1d
 from podpac.core.coordinates.stacked_coordinates import StackedCoordinates
 from podpac.core.coordinates.cfunctions import clinspace
+
+_logger = logging.getLogger(__file__)
 
 
 class DependentCoordinates(BaseCoordinates):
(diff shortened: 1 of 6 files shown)


jmilloy:
Looking at this now.

podpac/core/coordinates/dependent_coordinates.py
```diff
@@ -492,6 +495,39 @@ def transpose(self, *dims, **kwargs):
             properties["dims"] = dims
             return DependentCoordinates(coordinates, **properties)
 
+    def issubset(self, other):
```
mpu-creare, line 498: I believe this is NEVER called in practice, since `Coordinates.issubset` loops through `Coordinates.values()` which gets you to the `ArrayCoordinatesNd.issubset` function instead...

jmilloy: I haven't looked closely at that yet, but that doesn't really make sense to me.

jmilloy: Meaning that if that is true, then I think there is a bug in `Coordinates.issubset`.

jmilloy:
This is what I would have expected, and passes all of your tests.

```
def issubset(self, other):
        mine = zip(*[self[dim].coordinates.ravel() for dim in self.dims])
        other = zip(*[other[dim].coordinates.ravel() for dim in self.dims])
        return set(mine).issubset(other)
```

I don't really follow what you mean by `From this point forward, it gets expensive. We're essentially doing NN interpolation here.`

I think there is some more work to do here.
 - `DependentCoordinates.issubset` needs to handle `other` of type `Coordinates`
 - I don't think that `StackedCoordinates.issubset` is correct.
 - I'm not sure that `Coordinates.issubset` handles `DependentCoordinates` correctly.

jmilloy:
I can look in more detail (basically write the tests that I think are failing) in the morning, and probably fix them before lunch.

podpac/core/coordinates/dependent_coordinates.py
mpu-creare: Yeah I'm wrong. It is called.

mpu-creare:
> 
> 
> I can look in more detail (basically write the tests that I think are failing) in the morning, and probably fix them before lunch.

Please do. As you can see I really don't know what I'm doing. 

```python
mine = zip(*[self[dim].coordinates.ravel() for dim in self.dims])
other = zip(*[other[dim].coordinates.ravel() for dim in self.dims])
return set(mine).issubset(other)
```
is a much cleaner implementation of what I was trying to do near the end of the function. But it could be very expensive computationally for cases with lots of coordinates. You are essentially finding the subset of overlapping data points which you want to pull out of the dataset if the result is True. If False, hopefully it will shortcut very rapidly. So, I chose to do that last, and catch any of the easier-to-find falses before that. Although arguably the 1-D comparisons (not pairwise) might be just as expensive...

jmilloy:
I don't think you save any time by checking dimensions separately in the False case, and obviously it takes a lot longer in the True case. Each one of these:

```
    if not all([self[dim].issubset(other) for dim in self.dims]):
            return False
```

takes about the same amount of time as one of these:

```
mine = zip(*[self[dim].coordinates.ravel() for dim in self.dims])
other = zip(*[other[dim].coordinates.ravel() for dim in self.dims])
mine.issubset(theirs)
```

You can do a million coordinates with 3 dimensions in less than a second.

mpu-creare:
Ok. I'll fix it up.

mpu-creare:
Or... are you working on this right now?

mpu-creare:
@jmilloy Is this ready to go?

jmilloy:
Yes, you could re-review, but I think it is good to go.

PR #425, opened 2020-09-28
FIX: Exception when bounds are lower precision than coordinates

mpu-creare:
Problem: 
```python
        # Confusing time intersection
        ct = Coordinates([["2012-05-19T12:00:00", "2012-05-19T13:00:00", "2012-05-20T14:00:00"]], ["time"])
        cti = Coordinates([["2012-05-18", "2012-05-19"]], ["time"])
        ct2 = ct.intersect(cti, outer=True)
```
This throws in error because `lt` is empty, since `np.datetime64('2020-01-01T00') < np.datetime64('2020-01-01') == False`

As a solution I change the bounds to be the same precision, so `np.datetime('2020-01-01')` --> `np.datetime('2020-01-01T23')`

### DIFF
--- a/podpac/core/coordinates/array_coordinates1d.py
+++ b/podpac/core/coordinates/array_coordinates1d.py
@@ -13,7 +13,7 @@ import traitlets as tl
 from collections import OrderedDict
 
 from podpac.core.utils import ArrayTrait
-from podpac.core.coordinates.utils import make_coord_array
+from podpac.core.coordinates.utils import make_coord_array, higher_precision_time_bounds
 from podpac.core.coordinates.coordinates1d import Coordinates1d
 
 
@@ -281,6 +281,9 @@ class ArrayCoordinates1d(Coordinates1d):
     # ------------------------------------------------------------------------------------------------------------------
 
     def _select(self, bounds, return_indices, outer):
+        if self.dtype == np.datetime64:
+            _, bounds = higher_precision_time_bounds(self.bounds, bounds, outer)
+
         if not outer:
             gt = self.coordinates >= bounds[0]
             lt = self.coordinates <= bounds[1]
--- a/podpac/core/coordinates/utils.py
+++ b/podpac/core/coordinates/utils.py
@@ -539,6 +539,46 @@ def lower_precision_time_bounds(my_bounds, other_bounds, outer):
     return my_bounds, other_bounds
 
 
+def higher_precision_time_bounds(my_bounds, other_bounds, outer):
+    """
+    When given two bounds of np.datetime64, this function will convert both bounds to the higher-precision (in terms of 
+    time unit) numpy datetime4 object if outer==True, otherwise only my_bounds will be converted.
+    
+    Parameters
+    -----------
+    my_bounds : List(np.datetime64)
+        The bounds of the coordinates of the dataset
+    other_bounds : List(np.datetime64)
+        The bounds used for the selection
+    outer : bool
+        When the other_bounds are higher precision than the input_bounds, only convert these IF outer=True
+        
+    Returns
+    --------
+    my_bounds : List(np.datetime64)
+        The bounds of the coordinates of the dataset at the new precision
+    other_bounds : List(np.datetime64)
+        The bounds used for the selection at the new precision, if outer == True, otherwise return original coordinates    
+        
+    Notes
+    ------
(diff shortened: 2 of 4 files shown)


mpu-creare:
@jmilloy I'm asking you to have a quick look since you are working on coordinates right now and you implemented this in the first place AND this can get pretty tricky in general.

podpac/core/coordinates/utils.py
```diff
@@ -539,6 +539,46 @@ def lower_precision_time_bounds(my_bounds, other_bounds, outer):
...
+    """
+    if not isinstance(other_bounds[0], np.datetime64) or not isinstance(other_bounds[1], np.datetime64):
+        raise TypeError("Input bounds should be of type np.datetime64 when selecting data from:", str(my_bounds))
+
+    if not isinstance(my_bounds[0], np.datetime64) or not isinstance(my_bounds[1], np.datetime64):
+        raise TypeError("Native bounds should be of type np.datetime64 when selecting data using:", str(other_bounds))
+
+    if my_bounds[0].dtype > other_bounds[0].dtype and outer:
+        sign = [0, 1]
+        other_bounds = [(b + s).astype(my_bounds[0].dtype) - 1 for s, b in zip(sign, other_bounds)]
```
mpu-creare, line 575: Not completely sure this line is as it should be... in particular perhaps it should be 
```suggestion
        other_bounds = [(b + s).astype(my_bounds[0].dtype) - s for s, b in zip(sign, other_bounds)]
```

**note** `-1` change to `s`

mpu-creare:
@jmilloy mind having a quick look at this? Trying to get it into the next release.

jmilloy:
Looking at this now. I never saw the original notifications about this.

jmilloy:
@mpu-creare Where else did we have something like this? I feel like you committed a datetime precision fix once before...

mpu-creare:
There have been lots of time-related bugs. I think in the past it had something to do with edges, or edgecases... probably in crange or clinspace. That's my memory.

jmilloy:
Oh, I see. `lower_precision_bounds` already existed, just higher_precision_bounds is new.

mpu-creare:
Oh right! Derp, forgot about that...

podpac/core/coordinates/utils.py
jmilloy: Yeah, it definitely *seems* like it should be  ` - s` instead of `- 1`.

I found this easier to think about unpacked:

```
other_bounds = [
    (other_bounds[0] + 0).astype(my_bounds[0].dtype) - 0,
    (other_bounds[1] + 1).astype(my_bounds[0].dtype) - 1
]
```

or probably just:

```
other_bounds = [
    other_bounds[0].astype(my_bounds[0].dtype),
    (other_bounds[1] + 1).astype(my_bounds[0].dtype) - 1
]
```

 Can you explain what the `+1 -1` does? We should add a comment about that.

podpac/core/coordinates/array_coordinates1d.py
```diff
@@ -281,6 +281,9 @@ def _get_definition(self, full=True):
     # ------------------------------------------------------------------------------------------------------------------
 
     def _select(self, bounds, return_indices, outer):
```
jmilloy, line 283: @mpu-creare Why is it that we use `lower_precision_time_bounds` for `UniformCoordinates1d._select` but `higher_precision_coordinates` for `ArrayCoordinates1d.select`? Seems like it should be the same, and we should be able to move that up into the `Coordinates1d.select` method.

https://github.com/creare-com/podpac/blob/514dc5da3975fa1c6051f0da1691cf5ba4972bf8/podpac/core/coordinates/uniform_coordinates1d.py#L434

jmilloy: > Seems like it should be the same, and we should be able to move that up into the top of the Coordinates1d.select method.

Oh, we do use `lower_precision_time_bounds` in `Coordinates1d.select`. It is probably redundant in `UniformCoordinates1d._select`. The ArrayCoordinates1d has to use the higher precision coordinates  for comparison to the actual coordinates. I get it now.

podpac/core/coordinates/utils.py
jmilloy: Okay, I get it the `+1` and `-1`. I'll add a comment.

jmilloy (approved):
I haven't thought about the non-outer test that you added, but otherwise this looks good.

mpu-creare:
Thanks for adding the comments and reviewing! Much appreciated.

PR #429, opened 2020-10-23
Clean up DataSource debug attributes

jmilloy:
## Bugfix

The bug that prompted this is that the selected coordinates were not actually being used in `_get_data`.

```
    (rsc, rsci) = self.coordinates.intersect(coordinates, outer=True, return_indices=True)

    self._requested_source_coordinates = rsc
    self._requested_source_coordinates_index = rsci

    ...

    (rsc, rsci) = _selector(rsc, rsci, coordinates)

    ...

    self._requested_source_data = self._get_data(
        self._requested_source_coordinates, self._requested_source_coordinates_index
    )
```

## Background

I think at one point, the `DataSource._evaluated_coordinates`, `Datasource._requested_source_coordinates`, etc were used in other methods. We've been trying to get rid of them in various maintenance tasks for a while. I think with the recent refactors, that is finally the case!

I've kept them as "debug" attributes, but they are longer used.

### DIFF
--- a/podpac/core/data/datasource.py
+++ b/podpac/core/data/datasource.py
@@ -163,11 +163,13 @@ class DataSource(Node):
     # privates
     _coordinates = tl.Instance(Coordinates, allow_none=True, default_value=None, read_only=True)
 
-    _original_requested_coordinates = tl.Instance(Coordinates, allow_none=True)
-    _requested_source_coordinates = tl.Instance(Coordinates)
-    _requested_source_coordinates_index = tl.Tuple()
-    _requested_source_data = tl.Instance(UnitsDataArray)
-    _evaluated_coordinates = tl.Instance(Coordinates)
+    if settings["DEBUG"]:
+        _original_requested_coordinates = tl.Instance(Coordinates, allow_none=True)
+        _requested_source_coordinates = tl.Instance(Coordinates)
+        _requested_source_coordinates_index = tl.Tuple()
+        _requested_source_boundary = tl.Dict()
+        _requested_source_data = tl.Instance(UnitsDataArray)
+        _evaluated_coordinates = tl.Instance(Coordinates)
 
     @tl.validate("boundary")
     def _validate_boundary(self, d):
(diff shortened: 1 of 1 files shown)


jmilloy:
@mpu-creare Am I right about that bug? I guess it's something we overlooked that  in _selector implementation / recent refactor? 

There are a few similar attributes in the `Interpolate` node, too, that I should clean up, too.

jmilloy:
Another note, the bug can be fixed like this:

```
        if _selector is not None:
            (rsc, rsci) = _selector(rsc, rsci, coordinates)
            self._requested_source_coordinates = rsc
            self._requested_source_coordinates_index = rsci
```

But I just think the messiness of these debug attributes caused the bug, so I would prefer this cleaned up version if it will work.

podpac/core/data/datasource.py
```diff
@@ -338,27 +340,30 @@ def _eval(self, coordinates, output=None, _selector=None):
         ]
         coordinates = coordinates.drop(extra)
 
-        # store input coordinates to evaluated coordinates
-        self._evaluated_coordinates = deepcopy(coordinates)
```
mpu-creare, line 342: I believe that this deepcopy was important...

jmilloy: Yeah, the coordinates are changed after this because the crs is transformed. But I think the only time the deepcopied self._evaluated_coordinates is used is for it's dims, which don't change. I'll make that more clear.

jmilloy: I'm not sure it's better, but I clarified that in the subsequent commit.

podpac/core/data/datasource.py
```diff
@@ -368,42 +376,45 @@ def _eval(self, coordinates, output=None, _selector=None):
...
+                    new_rsci.append(slice(np.max(index), np.max(index) + 1))
 
-            self._requested_source_coordinates_index = tuple(new_rsci)
+            rsci = tuple(new_rsci)
 
         # get data from data source
-        self._requested_source_data = self._get_data(
-            self._requested_source_coordinates, self._requested_source_coordinates_index
-        )
+        rsd = self._get_data(rsc, rsci)
```
mpu-creare, line 398: This used to take no inputs and use `self._requested_source_coordinates_index` etc. internally. I was never able to follow this code so I refactored it, and that's how I introduced the bug. Thanks for catching it +1

jmilloy: Well, thanks for finally fixing it so that we don't those internal variables.

mpu-creare (approved):
This is great thanks! Merging.

mpu-creare:
> 
> 
> @mpu-creare Am I right about that bug? I guess it's something we overlooked that in _selector implementation / recent refactor?
> 
> There are a few similar attributes in the `Interpolate` node, too, that I should clean up, too.

Don't worry about that for now. I'm deep inside that node right now and lots will change.

Issue #340, opened 2019-11-15
ENH/BUG: Raise warning when interpolation params are ignored


mlshapiro:
**Description**

Currently, some interpolation params will be ignored if the Interpolator doesn't support them. 

**Describe the solution you'd like**

Raise a warning if interpolation params are defined, but the interpolator being used doesn't use them.

**Additional context**

This will be dependent on keeping the Interpolators up to date with the params they support.

jmilloy:
@mlshapiro Can you give an example of an unsupported parameter? Are these parameters that we intend to support in the future?

jmilloy:
I think this happens when the spatial tolerance is ignored when dependent coordinates become necessary due to projection transform.

mlshapiro:
Yes, like `spatial_tolerance` and `time_tolerance` - they are included in classes like Rasterio and ScipyGrid, but they don't do anything. And they also may be used by Rasterio only by certain interpolation methods.  

I'm not sure a warning is necessary for this scenario, but I think that some though should go into how we handle parameters/dimensions that are *ignored* when doing an interpolation.

jmilloy:
Matt pointed out that it can be a surprise when the coordinates are converted to DependentCoordinates internally and then a different interpolation method is used which ignores your spatial tolerance.

mpu-creare:
And now that I found the notebook again, I'll point it out here: https://github.com/creare-com/podpac-examples/blob/master/notebooks/examples/drought-monitor/03-data-access-harmonization-processing.ipynb

PR #430, opened 2020-10-23
Feature/enhance interpolation

mpu-creare:
This is a large overhaul / enhancement to the underlying interpolation machinery. This PR also includes #424.

### Enhancements
* Adding `Selectors`
  * Source data will now be sub-selected *before* being retrieved. This is supported for "nearest", "linear", "bilinear" and "cubic" interpolation methods
* Can now do linear interpolation (both time and space) over a single (i.e. 1D) dimensions
* Improved the `NearestNeighborInterpolator` so that:
  * By default interpolation only happens inside the source bounds (set `respect_bounds=False` to overwrite this)
  * Users can choose to `remove_nan=True`, which will remove nans from the source before interpolating
  * **ALL** (?) combinations of stacked/unstacked coordinates work
* Stacked sources can now be evaluated based on one or more of the stacked dimension. I.e. evaluating "lat_time" using "time" works. 
* Added the `Xarray` interpolator, based on the `xarray` `interp` method (which wraps `scipy`)

### Fixes
* Unused interpolation parameters are no longer silently ignored, but logged instead. 
* Interpolation now no longer silently falls back to nearest interpolation in some cases


mpu-creare:
@jxm FYI

mpu-creare:
@jmilloy Could you please fix the WCS tests on this branch? 

I want to merge this in soonish because it's getting a little big.

jmilloy:
Should be done soon, fighting with traitlets.

jmilloy:
I pushed some fixes, but I think I'm down to a bug in the selector.

This is after line 346 in the datasource `_eval`:

```
>>> _selector
<bound method InterpolationManager.select_coordinates of InterpolationManager
	('default',): nearest, ['NearestNeighbor', 'XarrayInterpolator', 'Rasterio', 'ScipyPoint', 'ScipyGrid'], {}>
>>> rsc
Coordinates (EPSG:4326)
	lon: ArrayCoordinates1d(lon): Bounds[-132.9023, -53.60509999999999], N[3]
	lat: ArrayCoordinates1d(lat): Bounds[23.6293, 53.7588], N[2]
>>> rsci
(array([[ 0],
       [10],
       [99]]), array([[ 0, 99]]))
>>> self.coordinates[rsci]
Coordinates (EPSG:4326)
	lon: ArrayCoordinates1d(lon): Bounds[-132.9023, -53.60509999999999], N[3], Shape(3, 1)
	lat: ArrayCoordinates1d(lat): Bounds[23.6293, 53.7588], N[2], Shape(1, 2)
```

Notice that the `rsci` is not flat, and that `self.coordinates[rsci] != rsc`, which is a problem.

jmilloy:
I should have mentioned, it fails when it tries to convert the index to slices. We could flatten the index there and probably fix this issue. But I think that the selector should return flat index elements here, and *definitely* `self.coordinates[rsci]` should always match `rsc`.

jmilloy:
I pushed changes but there are still a bunch of tests failing. Let me know if you think any of them are mine to fix. The new indexes (from `Selector.merge_indices`) causes problems for the coordinates.

mpu-creare:
Thanks @jmilloy, I'll have a look.

mpu-creare:
@jmilloy I believe this is now (FINALLY!) ready. Found a lot more edge cases once I fleshed out the unit tests / coverage. Please review the code when you have time. It would be prudent not to rush it. I tried to be thorough but it's possible I missed a case here or there.

jmilloy:
@mpu-creare A bunch of tests still aren't passing for me. What do you think is going on?

[podpac.txt](https://github.com/creare-com/podpac/files/5561926/podpac.txt)

mpu-creare:
I changed the default CRS in settings to something that would have an altitude dimension defined... by default. That could be part of it. I'll have to dig on the rest. Will do that tonight.

mpu-creare:
@jmilloy there should now be a few more failing tests. :) 

Looking at your errors, I'm guessing it's a combination of the default CRS in the settings file, xarray version, rasterio version, and pyproj version.

mpu-creare:
I suppose it helps if I actually post my versions:
```bash
numpy==1.19.3
pyproj==3.0.0.post1
rasterio==1.1.8
scipy==1.5.3
traitlets==5.0.5
traittypes==0.2.1
xarray==0.16.1
zarr==2.5.0
```

jmilloy:
Well, I'm definitely behind you on everything, but I'm not behind anything in our setup.py requirements.

```
numpy==1.19.0
pyproj==2.6.1.post1
rasterio==1.1.5
scipy==1.5.1
traitlets==4.3.3
traittypes==0.2.1
xarray==0.15.1
zarr==2.4.0
```

I guess I'll make a new environment that similar to yours and test that first.

mpu-creare:
I created a fresh environment just the other day... we may have to update setup.py. I'm thinking of doing a 3.0 release this round.

jmilloy:
Yeah, it's mostly the CRS in the settings. I guess we need set the DEFAULT_CRS, or specify it, in certain tests.

podpac/core/interpolation/interpolator.py
```diff
@@ -198,6 +198,8 @@ class Interpolator(tl.HasTraits):
     methods_supported = tl.List(tl.Unicode())
     dims_supported = tl.List(tl.Unicode())
     spatial_tolerance = tl.Float(allow_none=True, default_value=np.inf)
+    time_tolerance = tl.Union([tl.Unicode(), tl.Instance(np.timedelta64, allow_none=True)])
```
jmilloy, line 201: @mpu-creare is the `time_tolerance` and `alt_tolerance` okay here? Without these, some of the tests throw warnings (and will throw exceptions in traitlets 5). There is another way I can fix it, though, if they really shouldn't be here in the base class.

podpac/core/data/test/test_datasource.py
```diff
@@ -330,26 +327,13 @@ def test_evaluate_missing_dims(self):
             node.eval(Coordinates([1], dims=["time"]))
 
     def test_evaluate_crs_transform(self):
```
jmilloy, line 329: @mpu-creare I took most of this out so that it was just testing the datasource part. I'll make sure to add this test back into the interpolation tests somewhere if it covers a case that isn't cover there already.

podpac/core/interpolation/interpolator.py
mpu-creare: These should not be in the base class -- not all the interpolators support those parameters... in fact spatial tolerance shouldn't be there either. Otherwise these parameters might be silently ignored, #340 . 

So yes, right now the way parameters for interpolators are being set will break in traitlets 5, I didn't consider that. If you can fix it, that would be great... perhaps we should do something similar as we did with the nodes and tag the interpolation parameters... feel free to push this fix off to another PR.

podpac/core/data/test/test_datasource.py
mpu-creare: Great! Thanks for actually looking carefully at the whole test suite.

podpac/core/interpolation/interpolator.py
jmilloy: I think we need to fix it now. The problem is the super call in the mase Interpolator `__init__`. It's easy to just extract the tolerance arguments in the child class `__init__` methods that need them. But, `spatial_tolerance` and `alt_tolerance` are both used in the base Interpolator `_loop_helper`, so it's kind of confusing or even buggy, separate from the traitlets issue. They are also all documented in the COMMON_INTERPOLATOR_DOCS, though that is less important.

jmilloy:
The `Interpolator` _loop_helper isn't really covered by the tests at all. Can we fix that before merging this?

podpac/core/interpolation/xarray_interpolator.py
```diff
@@ -0,0 +1,123 @@
...
+                )
+                continue
+            if not source_coordinates.is_stacked(d) and eval_coordinates.is_stacked(d):
+                new_dim = [dd for dd in eval_coordinates.dims if d in dd][0]
+                coords[d] = xr.DataArray(
+                    eval_coordinates[d].coordinates, dims=[new_dim], coords=[eval_coordinates.xcoords[new_dim]]
+                )
+
+            elif source_coordinates.is_stacked(d) and not eval_coordinates.is_stacked(d):
+                raise InterpolatorException("Xarray interpolator cannot handle multi-index (source is points).")
```
jmilloy, line 100: What's going on here? The Xarray Interpolator can handle stacked source to stacked eval, but not stacked source to unstacked eval? Why is it raised as an exception instead checked in `can_interpolate`?

podpac/core/interpolation/xarray_interpolator.py
```diff
@@ -0,0 +1,123 @@
...
+        kwargs = self.kwargs.copy()
+        kwargs.update({"fill_value": self.fill_value})
+
+        coords["kwargs"] = kwargs
+
+        if self.fill_nan:
+            for d in source_coordinates.dims:
+                if not np.any(np.isnan(source_data)):
+                    break
+                source_data = source_data.interpolate_na(method=self.method, dim=d)
```
jmilloy, line 114: It doesn't look like this is really tested. You have some fill_nan=True cases, but all of them break out, so we don't actually have a test that would fail if the nans weren't filled.

jmilloy (changes requested):
There is so much here, it's hard to really review.
 * I cleaned up some tests and warnings, and from that we have the tolerance traits to sort out.
 * I also looked at the coverage. Nearest neighbor interpolation looks well tested, which is important because you worked a lot on that. The XarrayInterpolator has the fill_nan case.
 * I also looked at the XarrayInterpolator specifically, since it is new, and just had a few questions.

mpu-creare:
Thanks for the review @jmilloy  -- working on it.

jmilloy:
FWIW, this is how I was going to solve the tolerance traits. I still think this is the best way to handle traits that are only defined in child classes. Note it is only necessary because the Interpolator `__init__` calls `super().__init__(**kwargs)` itself.

```
class NearestNeighbor(Interpolator):

    ...

    def __init__(self, spatial_tolerance=None, time_tolerance=None, alt_tolerance=None, **kwargs):
        if spatial_tolerance is not None:
            self.set_trait('spatial_tolerance', spatial_tolerance)
        if time_tolerance is not None:
            self.set_trait('time_tolerance', time_tolerance)
        if alt_tolerance is not None:
            self.set_trait('alt_tolerance', alt_tolerance)

        super(NearestNeighbor, self).__init__(**kwargs)
```

podpac/core/interpolation/xarray_interpolator.py
mpu-creare: Yep, that makes no sense. It's actually already handled in the `can_interpolate` function. This shouldn't be there. Removed (in a forthcoming commit).

podpac/core/interpolation/xarray_interpolator.py
mpu-creare: I now have an explicit test -- thanks very much for finding this missed test.

podpac/core/interpolation/interpolator.py
mpu-creare: I greatly simplified `_looper_helper`, which was a holdover from Phase I. Most of that functionality now lives in `interpolation_manager`.

mpu-creare:
Merging... will deal with any outstanding issues in the release branch during testing.

PR #431, opened 2020-10-28
WCS + soilgrids

jmilloy:


jmilloy:
@mpu-creare Should I write some tests using a WCS server? It will be easy to do, I just don't know if we want to put any tests in the suite that require a network.

mpu-creare:
I don't want unit tests that require network access. If you could add a few "integration" tests in there (that aren't run every commit) that would be great -- just so we can run at manually every now and again. For the unit tests, see if there's anything that's easy to test based on the coverage report? But don't spend too much time trying to mock a WCS server.

jmilloy:
Added some WCS tests for coverage.

podpac/core/data/ogc.py
```diff
@@ -4,373 +4,264 @@
...
-                UniformCoordinates1d(left, right, size=size[0], name="lon"),
-            ]
-        )
+            return super()._eval(coordinates, output=output, _selector=selector)
 
-    @common_doc(COMMON_DATA_DOC)
-    def get_coordinates(self):
-        """{get_coordinates}
+        # for uniform stacked, unstack to use the requested coordinates (the WCS server will interpolate)
+        if "lat_lon" in coordinates.dims and coordinates["lat"].is_uniform and coordinates["lon"].is_uniform:
```
mpu-creare, line 145: what about `"lon_lat"`, or `"lon_time_lat"`, or ... 

in my branch I defined the `"is_stacked"` helper: 
```python
if coordinates.is_stacked('lat') and coordinates.is_stacked('lon') and coordinates["lat"].is_uniform and coordinates["lon"].is_uniform```

File: podpac/core/data/ogc.py [hunk 2]
```diff
@@ -4,373 +4,264 @@
...
-                c = self._requested_coordinates[dim]
-                if c.size == 1:
-                    cs.append(ArrayCoordinates1d(c.coordinates[0], name=dim))
-                elif isinstance(c, UniformCoordinates1d):
-                    cs.append(UniformCoordinates1d(c.bounds[0], c.bounds[1], abs(c.step), name=dim))
-                else:
-                    # TODO: generalize/fix this
-                    # WCS calls require a regular grid, could (otherwise we have to do multiple WCS calls)
-                    cs.append(UniformCoordinates1d(c.bounds[0], c.bounds[1], size=c.size, name=dim))
+            udata = super()._eval(coordinates, output=None, _selector=selector)
```
mpu-creare, line 151: The remote server **may** not support the user's crs. Probably need to transform the coordinates to support crs's before this step.

mpu-creare: Also, did you drop unsupported coordinates before this step?

podpac/core/data/ogc.py
```diff
@@ -4,373 +4,264 @@
...
+        # check for errors
+        xml = bs4.BeautifulSoup(content, "lxml")
+        error = xml.find("serviceexception")
+        if error:
+            raise WCSError(error.text)
+
+        # get data using rasterio
+        with rasterio.MemoryFile() as mf:
+            mf.write(content)
+            dataset = mf.open(driver="GTiff")
```
mpu-creare, line 249: Does the driver need to change based on the `self.format`

podpac/core/data/test/test_wcs.py
```diff
@@ -1,196 +1,145 @@
...
-
-        # requests exceptions
-        self.mock_requests(data_status_code=400)
-        with pytest.raises(Exception):
-            output = node.eval(time_coordinates)
-        with pytest.raises(Exception):
-            output = node.eval(time_coordinates)
+        self.node1.coordinates
+
+    def test_eval_grid(self):
```
mpu-creare, line 111: Could you add a test with a different CRS (one not supported by the remote server).

podpac/datalib/soilgrids.py
```diff
@@ -0,0 +1,119 @@
...
+
+    format = "geotiff_byte"
+    max_size = 16384
+    _repr_keys = ["layer"]
+
+
+class SoilGridsWRB(SoilGridsBase):
+    """ SoilGrids: WRB classes and probabilities (WCS) """
+
+    source = "https://maps.isric.org/mapserv?map=/map/wrb.map"
```
mpu-creare, line 22: In terms of an implementation for a datasource, that is gloriously simple. Love it.

mpu-creare (approved):
You'll see a couple of comments, none are dealbreakers, would be good to address. I'm approving so you can merge at your discretion. 

I'm so excited to see this implementation. It's much better than the old way. Nice work.

podpac/core/data/ogc.py
```diff
@@ -4,373 +4,264 @@
...
-        Coordinates
+        # coordinates
+        w, s, e, n = metadata.boundingBoxWGS84
+        low = metadata.grid.lowlimits
+        high = metadata.grid.highlimits
+        xsize = int(high[0]) - int(low[0])
+        ysize = int(high[1]) - int(low[1])
+
+        coords = []
+        coords.append(UniformCoordinates1d(s, n, size=ysize, name="lat"))
```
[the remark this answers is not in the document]
jmilloy: Well, the coordinates end up being the same regardless, so we're just talking about variable names here. Would it be more sensible to as `w, n, e, s = metadata.boundingBoxWGS84`?

podpac/core/data/ogc.py
```diff
@@ -4,373 +4,264 @@
...
+        ):
 
-            r = http.request("GET", self.capabilities_url)
-            capabilities = r.data
-            if r.status != 200:
-                raise Exception("Could not get capabilities from WCS server:" + self.capabilities_url)
-        else:
-            raise Exception("Do not have a URL request library to get WCS data.")
+            def selector(rsc, rsci, coordinates):
+                return coordinates, tuple(slice(None) for dim in coordinates)
```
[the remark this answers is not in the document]
jmilloy: Yeah, that was great. At first I had subclassed WCS from `Node` and rewritten most of the `DataSource._eval` but without selecting coordinates, but this allowed me to re-use the `DataSource._eval` completely, and handle both grid coordinates directly and non-gridded coordinates with interpolation in the same node. A great result.

podpac/core/data/ogc.py
jmilloy: Maybe, I would need an example. Or do you want to throw a `NotImplemetedError("Not yet supported")` if the format is not `geotiff` or `geotiff_byte`, for now?

podpac/core/data/ogc.py
jmilloy: Yeah, thanks. Fixed now.

podpac/core/data/ogc.py
jmilloy: The coordinates are transformed by the DataSource._eval, **but** it is better to transform them here in order to select the correct case!

jmilloy: Dims are dropped later, also.

podpac/core/data/ogc.py
```diff
@@ -4,373 +4,264 @@
...
-    layer_name = tl.Unicode().tag(attr=True)
-    version = tl.Unicode(default_value=WCS_DEFAULT_VERSION).tag(attr=True)
-    crs = tl.Unicode(default_value=WCS_DEFAULT_CRS).tag(attr=True)
+    layer = tl.Unicode().tag(attr=True)
+    version = tl.Unicode(default_value="1.0.0").tag(attr=True)
 
-    # list of attribute names, used by __repr__ and __str__ to display minimal info about the node
-    _repr_keys = ["source", "interpolation"]
+    format = tl.Unicode(default_value="geotiff")
+    crs = tl.Unicode(default_value="EPSG:4326")
```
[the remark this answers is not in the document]
jmilloy: I don't think so. Using a different format or crs wouldn't change the data.

podpac/core/data/ogc.py
mpu-creare: Throw `NotImplementedError` for now.

podpac/core/data/ogc.py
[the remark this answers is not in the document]
mpu-creare: Good point.

podpac/core/data/ogc.py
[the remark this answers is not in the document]
mpu-creare: Just keep it the way you have it now.

podpac/core/data/ogc.py
jmilloy: Actually, I just made the trait an enum with only 'geotiff' and 'geotiff_byte'. We can add more later if we need to.

PR #438, opened 2020-12-08
WIP: Remove InterpolationMixin and update some of the tests.

jmilloy:


jmilloy:
Removing the InterpolationMixin is a bit of a challenge, the tests rely on interpolation all over the place, and I've started to get into the algorithms which maybe have more problems. I made this a separate branch because I'm not entirely convinced that we should still do this.

mpu-creare:
Sounds like we should talk.

jmilloy:
I think we should keep the `InterpolationMixin`. It feels pretty weird to eval and get back different coordinates. I think we should rename `RasterioBase` etc to `RasterioRaw`, which feels more intentional.

I think we should also rename `DataCompositor` to `DataCompositorRaw` and `InterpDataCompositor` to `DataCompositor`.

Actually, I think we should remove the existing `TileCompositor` that never really worked, and rename the new data compositors to `TileCompositorRaw` and `TileCompositor`.

mpu-creare:
I agree with the name suggestions. Please make it so!

jmilloy:
We may not remove this after-all. Iceboxing.

mpu-creare:
shall I close this?

jmilloy:
Yep

PR #452, opened 2021-03-23
Remove support for numpy return type in Algorithms

jmilloy:
Backwards Incompatible changes
 * removes support for numpy return type from `Algorithm.algorithm`
 * passes `coordinates` into `Algorithm.algorithm`. Using `self._requested_coordinates` is no longer necessary.

Fixes
 * handles extra, missing, and transposed dimensions returned from `Algorithm.algorithm` correctly when `output` is provided.

Testing
 * more comprehensive `Algorithm.algorithm` testing
 * test the base `Algorithm` class without using any `Algorithm` subclasses such as `Arange` and `Arithmetic`

Additional Features
 * `Coordinates.from_xarray` accepts `DataArray` and `Dataset` types, extracting the `coords` and `crs` as needed. You can still pass in `DataArrayCoordinates` and `DatasetCoordinates`.

### DIFF
--- a/podpac/core/coordinates/coordinates.py
+++ b/podpac/core/coordinates/coordinates.py
@@ -307,14 +307,14 @@ class Coordinates(tl.HasTraits):
         return cls([stacked], crs=crs)
 
     @classmethod
-    def from_xarray(cls, xcoord, crs=None):
+    def from_xarray(cls, x, crs=None):
         """
         Create podpac Coordinates from xarray coords.
 
         Arguments
         ---------
-        xcoord : xarray.core.coordinates.DataArrayCoordinates
-            xarray coords
+        x : DataArray, Dataset, DataArrayCoordinates, DatasetCoordinates
+            DataArray, Dataset, or xarray coordinates
         crs : str, optional
             Coordinate reference system. Supports any PROJ4 or PROJ6 compliant string (https://proj.org/).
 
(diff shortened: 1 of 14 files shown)


jmilloy:
I need this fix for geowatch, but it's big enough that I thought I'd run it by you. You can basically look at `algorithm.py` and then peek at `test_algorithm.py` or some of the algorithms to see the effect that it has.

podpac/core/interpolation/nearest_neighbor_interpolator.py
```diff
@@ -80,8 +80,8 @@ def interpolate(self, udims, source_coordinates, source_data, eval_coordinates,
         def is_stacked(d):
             return "_" in d
 
-        if hasattr(source_data, "attrs"):
-            bounds = source_data.attrs.get("bounds", {d: None for d in source_coordinates.udims})
+        if hasattr(source_data, "attrs") and "bounds" in source_data.attrs:
+            bounds = source_data.attrs["bounds"]
```
mpu-creare, line 84: This will/should break something. You need to initialize bounds in an else branch if you want to eliminate the 'dense' statement. 
```suggestion
            bounds = source_data.attrs["bounds"]
        else:
            bounds = {d: None for d in source_coordinates.udims}          
```

mpu-creare (approved):
Much cleaner. Thanks @jmilloy

podpac/core/coordinates/coordinates.py
```diff
@@ -324,13 +324,19 @@ def from_xarray(cls, xcoord, crs=None):
...
 
-        if not isinstance(
-            xcoord, (xarray.core.coordinates.DataArrayCoordinates, xarray.core.coordinates.DatasetCoordinates)
-        ):
-            raise TypeError("Coordinates.from_xarray expects xarray DataArrayCoordinates, not '%s'" % type(xcoord))
+        if isinstance(x, (xr.DataArray, xr.Dataset)):
+            coords = x.coords
+            if crs is None:
+                crs = x.attrs.get("crs")
+        elif isinstance(x, (xarray.core.coordinates.DataArrayCoordinates, xarray.core.coordinates.DatasetCoordinates)):
```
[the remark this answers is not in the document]
jmilloy: That was the main motivation for passing the DataArray directly. I think *every* use was either `from_xarray(data.coords)` (forgetting the crs) or `from_xarray(data.coords, crs=data.attrs.crs)` (the same DataArray for the xcoords and the crs). I'll add the warning, too.

podpac/core/coordinates/coordinates.py
```diff
@@ -324,13 +324,19 @@ def from_xarray(cls, xcoord, crs=None):
...
+            coords = x
```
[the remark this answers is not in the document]
jmilloy: I think you're right, I changed the local variable name to `xcoord` at least.

podpac/core/interpolation/nearest_neighbor_interpolator.py
jmilloy: The else statement is already is there!

mpu-creare: Oh well.. I suppose that's what happens when you only look at the diffs. 😝

PR #455, opened 2021-04-07
BUGFIX: Nearest Preview Select Coordinates

jmilloy:
1. Fixes the tests for `test_nearest_preview_select`
    * check both lat and lon coordinates
    * check the returned coordinates index

2. Fixes the `InterpolationManager.select_coordinates` index reconstruction
    * need to reconstruct from the *source* coordinates, not the already selected coordinates
    * this fix applies to all interpolators

3. Fixes the `NearestPreview.select_coordinates` coordinates selection.
    * It appears that the UniformCoordinates1d `stop` sometimes needs a padding included, but not always. We need a test for both cases.

### DIFF
--- a/podpac/core/interpolation/interpolation_manager.py
+++ b/podpac/core/interpolation/interpolation_manager.py
@@ -481,7 +481,7 @@ class InterpolationManager(object):
                 selected_coords[d] = source_coordinates[d]
             # np.ix_ call doesn't work with slices, and fancy numpy indexing does not work well with mixed slice/index
             if isinstance(selected_coords_idx[d], slice) and index_type != "slice":
-                selected_coords_idx[d] = np.arange(selected_coords[d].size)
+                selected_coords_idx[d] = np.arange(source_coordinates[d].size)[selected_coords_idx[d]]
 
         selected_coords = Coordinates(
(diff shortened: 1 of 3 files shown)


jmilloy:
@mpu This fixes the failing `nearest_preview_select` test, but by removing the padding here:

```
c = UniformCoordinates1d(
                        src_start,
                        # src_stop + ndelta * src_delta / 2,  # The delta/2 ensures the endpoint is included
                        src_stop,
                        ndelta * src_delta,
                        **src_coords.properties
                    )
```

The test fails with the padding. Can you provide an example that requires that padding? I guess we need to cover both cases.

mpu-creare:
I'll have a look at it. The src_delta / 2 was a hard-won change. I'll add a relevant test and fix the failing one... hopefully in the morning

mpu-creare:
Thanks for fixing the other bug. Much appreciated.

mpu-creare:
@jmilloy Have a quick look at c94adc4 . I'm happy otherwise. 

Thanks so much for reaching out about that line of code instead of just making the change to get the unit test to pass. That really helped to all of the details sorted out.

mpu-creare (approved):
Good to go.

PR #463, opened 2021-04-19
Feature/node probe

mpu-creare:
Adds a utility method to probe a node/pipeline at a point. It evaluates EVERY node in the pipeline, even if it is not being used. 

See the unit tests for expected usage.

### DIFF
--- a/podpac/core/interpolation/interpolation.py
+++ b/podpac/core/interpolation/interpolation.py
@@ -28,12 +28,15 @@ class InterpolationMixin(tl.HasTraits):
         return super()._repr_keys + ["interpolation"]
 
     def _eval(self, coordinates, output=None, _selector=None):
-        node = Interpolate(interpolation=self.interpolation)
+        node = Interpolate(interpolation=self.interpolation, source_id=self.hash)
         node._set_interpolation()
         selector = node._interpolation.select_coordinates
         node._source_xr = super()._eval(coordinates, _selector=selector)
         self._interp_node = node
-        return node.eval(coordinates, output=output)
+        r = node.eval(coordinates, output=output)
+        # Helpful for debugging
+        self._from_cache = node._from_cache
+        return r
 
 
 class Interpolate(Node):
@@ -95,6 +98,7 @@ class Interpolate(Node):
     """
 
     source = NodeTrait(allow_none=True).tag(attr=True)
+    source_id = tl.Unicode(allow_none=True).tag(attr=True)
     _source_xr = tl.Instance(UnitsDataArray, allow_none=True)  # This is needed for the Interpolation Mixin
 
     interpolation = InterpolationTrait().tag(attr=True)
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -424,8 +424,16 @@ class Node(tl.HasTraits):
 
             if (
                 isinstance(value, Node)
-                or (isinstance(value, (list, tuple, np.ndarray)) and all(isinstance(elem, Node) for elem in value))
-                or (isinstance(value, dict) and all(isinstance(elem, Node) for elem in value.values()))
+                or (
+                    isinstance(value, (list, tuple, np.ndarray))
+                    and (len(value) > 0)
+                    and all(isinstance(elem, Node) for elem in value)
+                )
+                or (
+                    isinstance(value, dict)
+                    and (len(value) > 0)
+                    and all(isinstance(elem, Node) for elem in value.values())
+                )
             ):
                 inputs[name] = value
             else:
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -26,6 +26,7 @@ from podpac.core.utils import JSONEncoder
 from podpac.core.utils import cached_property
 from podpac.core.utils import trait_is_defined
 from podpac.core.utils import _get_query_params_from_url, _get_from_url, _get_param
+from podpac.core.utils import probe_node
 from podpac.core.coordinates import Coordinates
 from podpac.core.style import Style
 from podpac.core.cache import CacheCtrl, get_default_cache_ctrl, make_cache_ctrl, S3CacheStore, DiskCacheStore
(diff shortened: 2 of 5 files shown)


podpac/core/interpolation/interpolation.py
```diff
@@ -28,12 +28,15 @@ def _repr_keys(self):
         return super()._repr_keys + ["interpolation"]
 
     def _eval(self, coordinates, output=None, _selector=None):
-        node = Interpolate(interpolation=self.interpolation)
+        node = Interpolate(interpolation=self.interpolation, source_id=self.hash)
```
mpu-creare, line 31: This was an oversight on my part and fixes a bug. The old line 36 would return from cache with a certain set of settings for ALL Nodes. (whoops).

podpac/core/node.py
```diff
@@ -424,8 +424,16 @@ def _base_definition(self):
 
             if (
                 isinstance(value, Node)
-                or (isinstance(value, (list, tuple, np.ndarray)) and all(isinstance(elem, Node) for elem in value))
-                or (isinstance(value, dict) and all(isinstance(elem, Node) for elem in value.values()))
+                or (
+                    isinstance(value, (list, tuple, np.ndarray))
+                    and (len(value) > 0)
```
mpu-creare, line 429: Added this part of the if statement. Doesn't to any damage and helps with the node_probe implementation.

mpu-creare: If this is not present, empty lists and dictionaries end up in the "inputs" field.

podpac/core/utils.py
```diff
@@ -475,3 +475,70 @@ def resolve_bbox_order(bbox, crs, size):
...
+            if isinstance(ll, list):
+                lll = flatten_list(ll)
+                nl.extend(lll)
+            else:
+                nl.append(ll)
+        return nl
+
+    c = [(v, d) for v, d in zip([lat, lon, time, alt], ["lat", "lon", "time", "alt"]) if v is not None]
+    coords = podpac.Coordinates([[v[0]] for v in c], [[d[1]] for d in c])
+    with podpac.settings:
```
mpu-creare, line 502: I'm abusing the cache here to make my life easy for figuring out the actual "active" state below. If anyone can think of a better way to do this, I'm all ears.

mpu-creare:
@jmilloy could you please spare a few minutes to review and merge when happy.

podpac/core/utils.py
```diff
@@ -475,3 +475,70 @@ def resolve_bbox_order(bbox, crs, size):
     size = size[::r]
 
     return {"lat": [lat_start, lat_stop, size[0]], "lon": [lon_start, lon_stop, size[1]]}
+
+
+def probe_node(node, lat=None, lon=None, time=None, alt=None, nested=False):
```
jmilloy, line 480: Can you add documentation here? I don't know at a first glance what `active` means, for example, or why `test_composited_prober_nested` has `"inputs": { "inputs": [ ...] }`.

jmilloy: This feels to me like it could be a method `Node.probe`.

podpac/core/utils.py
mpu-creare: @jmilloy pointed out a much better way to this.

podpac/core/utils.py
mpu-creare: Added documentation and the Node.probe method -- good suggestion.

mpu-creare:
@jmilloy let me know if/when you are happy and I'll merge.

podpac/core/utils.py
```diff
@@ -541,4 +544,33 @@ def get_entry(key, out, definition):
...
+            "inputs": inputs,
+            "name": n.style.name if n.style.name else key,
+            "node_hash": n.hash,
+        }
+        # Fix sources for Compositors
+        if isinstance(n, podpac.compositor.OrderedCompositor):
+            for inp in inputs:
+                out[inp]["active"] = False
+                if out[inp]["value"] == out[key]["value"] and np.isfinite(out[inp]["value"]):
+                    out[inp]["active"] = True
```
jmilloy, line 572: Hmm, I expected a break here. Only the first non-nan source is active, right?

jmilloy (commented):
This looks good. I just had one question about the OrderedCompositer active loop.

podpac/core/utils.py
mpu-creare: Yes, but by default I'm setting everything to `False`. So any layers after the `active` one still needs their default changed from `True` to `False`.

mpu-creare: Thanks for reviewing!

jmilloy:
I added a test for the case I'm concerned about. Now I think I get what you are doing (you check the source values and mark the source active if it matches the orderedcompositor value), so I was able to make a test that fails. See d812b926.

jmilloy:
Hold on, I jumped the gun. Just a sec.

mpu-creare:
Oh True... I see what you mean... Yeah, if two sources give the same value they would both be marked active. That does need a fix.

mpu-creare:
ok, let me fix it.

jmilloy:
It was a bit tricky because if two OrderedCompositor sources are the same (not object, but definition/hash), only one of them shows up in the compositor definition. So of course only one shows up in the probe output. I'm not sure if that should be considered a bug or a feature!

```
>>> a = podpac.data.Array(source=np.ones(3), coordinates=podpac.Coordinates([[0, 1, 2]], dims=['time'])) 
>>> b = podpac.data.Array(source=np.ones(3), coordinates=podpac.Coordinates([[0, 1, 2]], dims=['time'])) 
>>> c = podpac.compositor.OrderedCompositor(sources=[a, b])
>>> c.probe(time=1)
OrderedDict([('Array', {'active': True, 'value': 1.0, 'inputs': [], 'name': 'Array', 'node_hash': 'e032d95041f8076a2a6317c079ace615'}), ('OrderedCompositor', {'active': True, 'value': 1.0, 'inputs': ['Array', 'Array'], 'name': 'OrderedCompositor', 'node_hash': '98354d0bc306ec5c8657590e04c0969a'})])
```

jmilloy:
Here's another fun one. Fails because the class isn't available in the global scope:

```
>>> def fn():
...     class AnotherOne(podpac.algorithm.Algorithm):
...         def algorithm(self, inputs, coordinates):
...             return self.create_output_array(coordinates, data=1)
...     c = podpac.compositor.OrderedCompositor(sources=[a, b, AnotherOne()])
...     c.probe(time=1)
... 
>>> fn()
  File "/home/jmilloy/Creare/Pipeline/podpac/podpac/core/node.py", line 816, in from_definition
    node_class = getattr(module, node_name)
AttributeError: module '__main__' has no attribute 'AnotherOne'

During handling of the above exception, another exception occurred:

  File "<stdin>", line 6, in fn
  File "/home/jmilloy/Creare/Pipeline/podpac/podpac/core/node.py", line 418, in probe
    return probe_node(self, lat, lon, time, alt)
  File "/home/jmilloy/Creare/Pipeline/podpac/podpac/core/utils.py", line 556, in probe_node
    n = podpac.Node.from_definition(d)
  File "/home/jmilloy/Creare/Pipeline/podpac/podpac/core/node.py", line 818, in from_definition
    raise ValueError(
ValueError: Invalid definition for node 'AnotherOne': class 'AnotherOne' not found in module '__main__'
```

In the test, I had to define the `AnotherOne` class globally instead of in the test method.

Again, not sure if this is a bug or just the way things are...

PR #465, opened 2021-04-20
Feature/none interpolation

mpu-creare:
Adding a new `NoneInterpolation` interpolator. 

This allows users to get raw data from DataSource nodes with little additional effort. 

This implementation supports partial `none` interpolation -- that is, you can interpolate lat/lon using nearest neighbor, without interpolating time, e.g. 
```json
interpolation=[
    {"method": "nearest", "dims": ["lat", "lon"]}, 
    {"method": "none", "dims": ["time"]},
] 
```

This has NOT been tested with any real datasources yet.


jmilloy (approved):

podpac/core/interpolation/none_interpolator.py
```diff
@@ -0,0 +1,68 @@
...
+        udims_subset = self._filter_udims_supported(udims)
+
+        return udims_subset
+
+    @common_doc(COMMON_INTERPOLATOR_DOCS)
+    def interpolate(self, udims, source_coordinates, source_data, eval_coordinates, output_data):
+        """
+        {interpolator_interpolate}
+        """
+        # Note, some of the following code duplicates code in the Selector class.
```
mpu-creare, line 47: Stray comment here.

PR #470, opened 2021-04-21
Replace Style enumeration tuple traits with dict traits.

jmilloy:


jmilloy:
@mpu-creare I replaced the tuples traits with dictionaries, but I think I can easily support both if you want this to be backwards compatible.

mpu-creare (approved):
The tests really make the need for it clear.

PR #473, opened 2021-04-28
FIX: Use abs. value for step comparison

### LINKED ISSUE #472: Interpolation selector uses full coordinate range with negative step
When interpolating, if the step of the source coordinates is larger than the step of the requested coordinates, the request dataset is evaluated at all source coordinates. However, if the request coordinates have a negative step that is larger in magnitude than the source coordinate step, the request dataset is unnecessarily evaluated at all source coordinates rather than at a subset, impacting performance. 

https://github.com/creare-com/podpac/blob/develop/podpac/core/interpolation/selector.py#L138

rmeyers4:
Closes #472

### DIFF
--- a/podpac/core/interpolation/nearest_neighbor_interpolator.py
+++ b/podpac/core/interpolation/nearest_neighbor_interpolator.py
@@ -283,7 +283,7 @@ class NearestNeighbor(Interpolator):
                 step = self._time_to_float(source.step, source, request)
             else:
                 step = source.step
-            rindex[np.abs(index - rindex) * step > tol] = -1
+            rindex[np.abs(index - rindex) * np.abs(step) > tol] = -1
 
         return rindex
 
--- a/podpac/core/interpolation/selector.py
+++ b/podpac/core/interpolation/selector.py
@@ -135,7 +135,7 @@ class Selector(tl.HasTraits):
 
     def _select_uniform(self, source, request, index_type):
         crds = request[source.name]
-        if crds.is_uniform and crds.step < source.step and not request.is_stacked(source.name):
+        if crds.is_uniform and abs(crds.step) < abs(source.step) and not request.is_stacked(source.name):
             return np.arange(source.size)
 
         index = (crds.coordinates - source.start) / source.step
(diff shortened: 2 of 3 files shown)


podpac/core/interpolation/nearest_neighbor_interpolator.py
```diff
@@ -283,7 +283,7 @@ def _get_uniform_index(self, dim, source, request, bounds=None):
                 step = self._time_to_float(source.step, source, request)
             else:
                 step = source.step
-            rindex[np.abs(index - rindex) * step > tol] = -1
+            rindex[np.abs(index - rindex) * np.abs(step) > tol] = -1
```
mpu-creare, line 286: I found one more stray absolute value... that's why I wanted to do a thorough review.

podpac/core/interpolation/selector.py
```diff
@@ -98,11 +98,11 @@ def select(self, source_coords, request_coords, index_type="numpy"):
         coords = []
         coords_inds = []
         for coord1d in source_coords._coords.values():
-            c, ci = self._select1d(coord1d, request_coords, index_type)
+            ci = self._select1d(coord1d, request_coords, index_type)
```
mpu-creare, line 101: Small refactor here, we weren't actually returning a modified coordinate so that was confusing before.

podpac/core/interpolation/selector.py
```diff
@@ -135,7 +135,7 @@ def _merge_indices(self, indices, source_dims, request_dims):
 
     def _select_uniform(self, source, request, index_type):
         crds = request[source.name]
-        if crds.is_uniform and crds.step < source.step and not request.is_stacked(source.name):
+        if crds.is_uniform and abs(crds.step) < abs(source.step) and not request.is_stacked(source.name):
```
mpu-creare, line 138: @rmeyers4 Nice fix -- thanks!

podpac/core/interpolation/test/test_selector.py
```diff
@@ -149,6 +149,81 @@ def test_bilinear_selector(self):
                     ),
                 )
 
+    def test_bilinear_selector_negative_step(self):
```
mpu-creare, line 152: I added a bunch more tests to make sure all the cases are covered.

mpu-creare (approved):
Nice bugfix -- thanks for the contribution.

PR #475, opened 2021-05-12
WIP: Implementing direct reading from Rasterio Overviews.

mpu-creare:
Adding option to pull data from overviews when the request is coarser than the data resolution. 

To use this new functionality, set the `prefer_overviews` attribute to `True`.

### DIFF
--- a/podpac/core/data/datasource.py
+++ b/podpac/core/data/datasource.py
@@ -137,7 +137,7 @@ class DataSource(Node):
     nan_vals : List, optional
         List of values from source data that should be interpreted as 'no data' or 'nans'
     coordinate_index_type : str, optional
-        Type of index to use for data source. Possible values are ``['slice', 'numpy']``
+        Type of index to use for data source. Possible values are ``['slice', 'numpy', 'xarray']``
         Default is 'numpy', which allows a tuple of integer indices.
     cache_coordinates : bool
         Whether to cache coordinates using the podpac ``cache_ctrl``. Default False.
@@ -372,7 +372,10 @@ class DataSource(Node):
         data = rsd
         if output is None:
             if requested_coordinates.crs.lower() != coordinates.crs.lower():
-                data = self.create_output_array(rsc, data=data.data)
+                if rsc.shape == data.shape:
+                    data = self.create_output_array(rsc, data=data.data)
+                else:
+                    crds = Coordinates.from_xarray(data, crs=data.attrs.get("crs", None))
             output = data
         else:
             output.data[:] = data.data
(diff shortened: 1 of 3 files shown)


podpac/core/data/datasource.py
```diff
@@ -372,7 +372,11 @@ def _eval(self, coordinates, output=None, _selector=None):
         data = rsd
         if output is None:
             if requested_coordinates.crs.lower() != coordinates.crs.lower():
-                data = self.create_output_array(rsc, data=data.data)
+                if rsc.shape == data.shape:
+                    data = self.create_output_array(rsc, data=data.data)
+                else:
+                    crds = Coordinates.from_xarray(data, crs=data.attrs.get("crs", None))
+                    data = self.create_output_array(crds.transform(rsc.crs), data=data.data)
```
mpu-creare, line 379: I think this piece is needed in general... since we've relaxed the requirement that get_data returns exactly what was asked...

podpac/core/data/rasterio_source.py
```diff
@@ -140,10 +155,65 @@ def get_data(self, coordinates, coordinates_index):
         data.data.ravel()[:] = raster_data.ravel()
         return data
 
+    def get_data_overviews(self, coordinates, coordinates_index):
+        # Figure out how much coarser the request is than the actual data
+        reduction_factor = np.inf
+        for c in ["lat", "lon"]:
```
mpu-creare, line 161: I'm assuming uniform lat/lon resolution

mpu-creare:
I merged rapidly, but would still like someone to look at the code. I needed this on geowatch.

jmilloy:
I'll take a look on Monday.

podpac/core/data/rasterio_source.py
jmilloy: What do you mean here? You are assuming that the requested coordinates are uniform lat/lon, or that the overview coordinates are? Can we just check that and raise a RuntimeError if they aren't uniform?

podpac/core/data/rasterio_source.py
```diff
@@ -140,10 +155,65 @@ def get_data(self, coordinates, coordinates_index):
...
+            crd = coordinates[c]
+            if isinstance(crd, UniformCoordinates1d):
```
jmilloy, line 163: You should use `if crd.is_uniform` here. The ArrayCoordinates1d have a `step` property.

jmilloy (commented):
Shall we add some tests?

mpu-creare:
> 
> 
> Shall we add some tests?

Yes I/we should... but I didn't have a geotiff with overviews handy.

PR #477, opened 2021-05-18
Discard extra dimensions when caching datasource and compositor outputs

### LINKED ISSUE #293: Cache coordinates for a datasource should depend only on the native_coordinates.dims
**Description**
Right now if I cache a node with only lat-lon coordinates, but I use a lat-lon-time request, every new time will cache a new object. 

For datasources, the node should be cached based on only coordinate dimensions that are a part of the native coordinates

**Sample Code**
```python
node = podpac.data.Array(
    source=np.ones((3, 4)),
    coordinates=podpac.Coordinates([range(3), range(4)], ['lat', 'lon']),
    cache_ctrl=['ram']
)
coords1 = podpac.Coordinates([range(3), range(4), '2012-05-19'], ['lat', 'lon', 'time'])
coords2 = podpac.Coordinates([range(3), range(4), '2019-09-10'], ['lat', 'lon', 'time'])
node.eval(coords1)   # Output gets cached
node._from_cache # False
node.eval(coords1) # Retrieved from cache
node._from_cache # True
node.eval(coords2)
node._from_cache # False <-- This should be True after this enhancement
```

jmilloy:
Closes #293

### DIFF
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -281,6 +281,10 @@ class Node(tl.HasTraits):
         key = "output"
         cache_coordinates = coordinates.transpose(*sorted(coordinates.dims))  # order agnostic caching
 
+        if isinstance(self, (podpac.data.DataSource, podpac.core.compositor.BaseCompositor)) and self.dims:
+            extra_dims = [dim for dim in cache_coordinates.dims if dim not in self.dims]
+            cache_coordinates = cache_coordinates.drop(extra_dims)
+
         if not self.force_eval and self.cache_output and self.has_cache(key, cache_coordinates):
             data = self.get_cache(key, cache_coordinates)
             if output is not None:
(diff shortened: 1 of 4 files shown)


podpac/core/node.py
```diff
@@ -281,6 +281,10 @@ def eval(self, coordinates, **kwargs):
         key = "output"
         cache_coordinates = coordinates.transpose(*sorted(coordinates.dims))  # order agnostic caching
 
+        if isinstance(self, (podpac.data.DataSource, podpac.core.compositor.BaseCompositor)) and self.dims:
+            extra_dims = [dim for dim in cache_coordinates.dims if dim not in self.dims]
+            cache_coordinates = cache_coordinates.drop(extra_dims)
+
```
jmilloy, line 287: You'll notice that I did both datasources and compositors. Let me know if you think I should only do it for datasources and not compositors.

jmilloy, line 287: @mpu-creare You'll notice that I did both datasources and compositors. Let me know if you think I should only do it for datasources and not compositors.

jmilloy:
Superseded by #478

Issue #471, opened 2021-04-23
cache documentation


jmilloy:
**Description**
The cache documentation is sparse. At the very least, add documentation about cache expiration.

Issue #436, opened 2020-12-05
Caching doesn't discriminate between evals with and without selector


mpu-creare:
**Description**
When I eval a pipeline: `coords --> interpolate --> source` or when I just eval the source `coords --> source` the cache will think it is the same. 

Thus, when I eval one of the two first, evaluating the second one gives me the wrong data. 

**Steps to Reproduce**
Sorry for the long example:
```python
point_interpolation = {
    "method": "nearest",
    "params": {
        "use_selector": False,  # By default, only the required data is retrieved remotely. This retrieves all the available data
        "remove_nan": True,     # Any nans will be removed prior to interpolation
        "time_scale": "1,M",    # When determining the closest neighbor, 1 month is considered equivalent to 1 degree
        "respect_bounds": False,# Extrapolation allowed
    },
}

soilscape = podpac.datalib.soilscape.SoilSCAPE20min(data_key="soil_moisture", site="Kendall_AZ", interpolation=point_interpolation)

time2 = podpac.crange('2015-12-01', '2016-02-01', '3,D', 'time')  # (start, stop, step)

# Make coordinates using the soilscape node - this selects soilscape probe coordinates 
time_series = soilscape.make_coordinates(depth=4.0, time=time2.coordinates)
time_series

cosmos_raw = podpac.datalib.cosmos_stations.COSMOSStationsRaw()
cosmos_daily = podpac.interpolators.Interpolate(source=cosmos_daily, interpolation=point_interpolation)
```
Here's the actual issue:
```python
o1 = cosmos_daily.eval(time_series)  # This passes a selector created form the interpolate node which subselects the source data
o2 = cosmos_raw.eval(time_series)  # This returns from the cache, but it shouldn't
```

**Expected Behavior**
The cache should take into account selectors passed through the eval.

**Observed Behavior**
The cache does not take into account selectors passed through the eval.

jmilloy:
The `DataCompositor` makes this difficult. The bug can be fixed in a DataSource by defining a custom `DataSource.eval` that does the selection and then calls super to get  to the `Node.eval` with the caching, but if you use a `DataCompositor` it still caches before even getting to its sources.

I fixed *this* example by not letting the `DataCompositor` cache (force `cache_output=False`) and having the COSMOSStationsRaw *sources* cache their output: 

```
    def sources(self):
        return np.array([COSMOSStation(station_data=item, cache_output=True) for item in self.stations_data["items"]])[self.has_data]
```

[...]

PR #478, opened 2021-05-27
Cache improvements

### LINKED ISSUE #432: CRS-agnostic caching
`Node.eval` transposes coordinates so that requests that differ only in dims-order can still use the cache. This is "dims order agnostic caching". We should try to do the same for CRS.

jmilloy:
Features
 - CRS-agnostic output caching
 - Discard extra dimensions when caching DataSource and Compositor outputs

Bugfixes
 - Output is cached depending on `selector`
 - StackedCoordinates `simplify` method (`Coordinates.simplify` now works with stacked coordinates)

Other
  - Cache documentation is improved

closes #432 #293 #436 #471

### DIFF
--- a/podpac/core/compositor/compositor.py
+++ b/podpac/core/compositor/compositor.py
@@ -217,6 +217,13 @@ class BaseCompositor(Node):
 
     @common_doc(COMMON_COMPOSITOR_DOC)
     def eval(self, coordinates, **kwargs):
+        """
+        Wraps the super Node.eval method in order to cache with the correct coordinates.
+
+        The output is independent of any extra dimensions, so this removes extra dimensions before caching in the
+        super eval method.
+        """
+
         super_coordinates = coordinates
 
         # remove extra dimensions
@@ -229,6 +236,7 @@ class BaseCompositor(Node):
             ]
             super_coordinates = super_coordinates.drop(extra)
 
+        # note: super().eval (not self._eval)
         output = super().eval(super_coordinates, **kwargs)
 
         if settings["DEBUG"]:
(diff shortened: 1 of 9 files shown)


jmilloy:
This is now all set. Not complicated.

podpac/core/compositor/compositor.py
```diff
@@ -215,6 +215,27 @@ def iteroutputs(self, coordinates, _selector=None):
...
+        if self.dims:
+            extra = [
+                c.name
+                for c in coordinates.values()
+                if (isinstance(c, Coordinates1d) and c.name not in self.dims)
+                or (isinstance(c, StackedCoordinates) and all(dim not in self.dims for dim in c.dims))
+            ]
+            super_coordinates = super_coordinates.drop(extra)
+
+        output = super().eval(super_coordinates, **kwargs)
```
mpu-creare, line 232: THAT is fantastic! Get infront of the eval that actually does the caching! Brilliant!

mpu-creare: Actually, that's tricky enough that you should add a comment explaining why you suddenly have an `eval` function showing up instead of the usual `_eval` function.

podpac/core/data/datasource.py
```diff
@@ -275,6 +281,45 @@ def _get_data(self, rc, rci):
...
+            if (isinstance(c, Coordinates1d) and c.name not in self.coordinates.udims)
+            or (isinstance(c, StackedCoordinates) and all(dim not in self.coordinates.udims for dim in c.dims))
+        ]
+        coordinates = coordinates.drop(extra)
+
+        # transform coordinates into native crs if different
+        if coordinates.crs.lower() != self.coordinates.crs.lower():
+            coordinates = coordinates.transform(self.coordinates.crs)
+
+        output = super().eval(coordinates, **kwargs)
```
mpu-creare, line 311: Again, please add a comment about why we are calling the super here instead of using `_eval`

mpu-creare (changes requested):
THis looks great. Pretty simple. Sorry I missed it for a few months. Please just add the comment to the compositors and datasource so we don't undo the change sometime in the future.

Issue #451, opened 2021-03-19
Propagate traits to sources/inputs


jmilloy:
Some traits, such as the caching, need to be propagated to Compositor sources, Algorithm inputs, or the Interpolator source. Currently, nodes handle this individually and incompletely.

We will use a tag, such as `sync` or `propagate`, to mark which traits need to be propagated.

```
tl.bool(cache_ctrl).tag(propagate=True)
```

We discussed propagating these tagged traits at eval time if possible, ideally in `Node.eval` or in the various `_eval` methods if necessary. Note that Compositor sources are often instantiated as needed in different ways, so we need to make sure those cases are covered.

We also discussed specifying not just that a trait needs to be propagated, but *where* it needs to be propagated to for fine-grained control.

```
tl.bool(cache_ctrl).tag(propagate='sources')
tl.bool(cache_output).tag(propagate=['a', 'b'])
```

Traits should only propagate if it exists but has not been defined in the target node.

jmilloy:
@mpu-creare This still bites me sometimes. I'm looking forward to fixing this. I'm okay with the `propagate` flag that we speced together, but I wonder what you think about gathering these traits into a single configuration `environment` that can be propagated. This would be a some (but not all) of the `settings` as an object that can be easily copied and modified and propagated. Each node would just have a single 'env' trait.

```python3
env = podpac.Environment(cache_ctrl=['ram'])
node = MyCompositor(env=env)
node.sources[0].env.cache_ctrl # ['ram']

# env traits can also be set like this
env = podpac.Environment()
env.multithreading = False
```

A context manager would be nice:

```python3
with podpac.Environment(cache_ctrl=['ram']) as env:
    node = MyNode(env=env)
    node.eval(coords)
```

Especially if the environment was automatically populated (not sure if this is feasible):

```python3
with podpac.Environment(cache_ctrl=['ram']):
    node = MyNode()
    node.eval(coords)
```

I'm not sure if this is feasible, either, but there is potential to evaluate the same node with different configuration environments:

```python3
node = MyCompositor()
node.eval(coords)
with podpac.Environment(cache_ctrl=['ram']):
    node.eval(coords)
with podpac.Environment(cache_ctrl=['disk']):
    node.eval(coords)
```

Keyword arguments for ease of use. These set up the `environment` under the hood:

```
node = MyNode(multithreading=True)

# same as
with podpac.Environment(mulithreading=True) as env:
    node1 = MyNode() 
```

[...]

PR #479, opened 2021-06-01
Performance Optimizations + A few Bug Fixes

mpu-creare:
A few performance optimizations and bug fixes.

This PR also add the following features:
* `Compositers.eval_serial`: For some compositors, it's important to actually evaluate the nodes in serial for performance reasons. So, even when multi-threading is used, when this flag is `True` the datasources will be serially evaluated.
* `RasterioSource.prefer_overview_closests`: when selecting overview levels, we can either select the coarsest overview smaller than the user-specified step size (based on eval coordinates) OR we can select the overview with the closest step size to the eval coordinates... this may be coarser than the specified coordinates. Setting this attr to `True` will select the closest overview instead of the closests higher resolution overview.

### DIFF
--- a/podpac/core/compositor/compositor.py
+++ b/podpac/core/compositor/compositor.py
@@ -32,6 +32,8 @@ class BaseCompositor(Node):
         Source nodes.
     source_coordinates : :class:`podpac.Coordinates`
         Coordinates that make each source unique. Must the same size as ``sources`` and single-dimensional. Optional.
+    eval_serial : bool, optional
+        Default is False. If True, will always evaluate the compositor in serial, ignoring any MULTITHREADING settings
 
     Notes
     -----
@@ -48,6 +50,7 @@ class BaseCompositor(Node):
 
     sources = tl.List(trait=NodeTrait()).tag(attr=True)
     source_coordinates = tl.Instance(Coordinates, allow_none=True, default_value=None).tag(attr=True)
+    eval_serial = tl.Bool(False)
 
     dims = tl.List(trait=Dimension()).tag(attr=True)
     auto_outputs = tl.Bool(False)
@@ -192,14 +195,14 @@ class BaseCompositor(Node):
             yield self.create_output_array(coordinates)
             return
 
-        if settings["MULTITHREADING"]:
+        if settings["MULTITHREADING"] and not self.eval_serial:
             n_threads = thread_manager.request_n_threads(len(sources))
             if n_threads == 1:
                 thread_manager.release_n_threads(n_threads)
         else:
             n_threads = 0
 
-        if settings["MULTITHREADING"] and n_threads > 1:
+        if settings["MULTITHREADING"] and n_threads > 1 and not self.eval_serial:
             # evaluate nodes in parallel using thread pool
             self._multi_threaded = True
             pool = thread_manager.get_thread_pool(processes=n_threads)
--- a/podpac/core/compositor/compositor.py
+++ b/podpac/core/compositor/compositor.py
@@ -32,7 +32,7 @@ class BaseCompositor(Node):
         Source nodes.
     source_coordinates : :class:`podpac.Coordinates`
         Coordinates that make each source unique. Must the same size as ``sources`` and single-dimensional. Optional.
-    eval_serial : bool, optional
+    multithreading : bool, optional
         Default is False. If True, will always evaluate the compositor in serial, ignoring any MULTITHREADING settings
 
     Notes
@@ -50,7 +50,11 @@ class BaseCompositor(Node):
 
     sources = tl.List(trait=NodeTrait()).tag(attr=True)
     source_coordinates = tl.Instance(Coordinates, allow_none=True, default_value=None).tag(attr=True)
-    eval_serial = tl.Bool(False)
+    multithreading = tl.Bool(False)
+
+    @tl.default("multithreading")
+    def _default_multithreading(self):
+        return settings["MULTITHREADING"]
 
     dims = tl.List(trait=Dimension()).tag(attr=True)
     auto_outputs = tl.Bool(False)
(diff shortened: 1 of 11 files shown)


podpac/core/coordinates/coordinates.py
```diff
@@ -856,7 +857,7 @@ def json(self):
 
         return json.dumps(self.definition, separators=(",", ":"), cls=podpac.core.utils.JSONEncoder)
 
-    @property
+    @cached_property
```
mpu-creare, line 860: @jmilloy I ended up caching this because it did cause quite a bit of slowdown -- it unfortunately leads to some hard-to-find caching bugs when running interactively.

podpac/core/coordinates/uniform_coordinates1d.py
```diff
@@ -215,7 +216,10 @@ def from_definition(cls, d):
     def __getitem__(self, index):
         # fallback for non-slices
         if not isinstance(index, slice):
-            return ArrayCoordinates1d(self.coordinates, **self.properties)[index]
+            # The following 3 lines is copied from ArrayCoordinates1d.__getitem__
```
mpu-creare, line 219: These lines were added so that I could cache the `self.coordinates` call. Otherwise these were constantly being recreated.

podpac/core/coordinates/uniform_coordinates1d.py
```diff
@@ -269,7 +273,7 @@ def __contains__(self, item):
     # Properties
     # ------------------------------------------------------------------------------------------------------------------
 
-    @property
+    @cached_property
```
mpu-creare, line 276: I'm pretty sure caching this is fine.

mpu-creare (commented):
Just a few things to look at to make sure we're not doing anything too too strange.

podpac/core/compositor/compositor.py
```diff
@@ -48,6 +50,7 @@ class BaseCompositor(Node):
 
     sources = tl.List(trait=NodeTrait()).tag(attr=True)
     source_coordinates = tl.Instance(Coordinates, allow_none=True, default_value=None).tag(attr=True)
+    eval_serial = tl.Bool(False)
```
jmilloy, line 53: I'd love to name this `multithreading`, like the setting. It should use the value from the setting by default, e.g.

```python3
multithreading = tl.Bool()

tl.default('multithreading')
def _default_multithreading(self):
    return settings['MULTITHREADING']
```

(And the OrderedCompositor can still use a different default value). Then you can just use `if self.multithreading` below.

This is related to #451, where it would be great if *all* evaluation controls like this could be overriden by any Node. If we use the same name for this trait now, we don't have to change it later.

podpac/core/coordinates/coordinates.py
```diff
@@ -360,7 +360,7 @@ def from_xarray(cls, x, crs=None):
                 # unstacked
                 d[dim] = ArrayCoordinates1d.from_xarray(xcoords[dim])
 
-        coords = cls(list(d.values()), crs=crs)
+        coords = cls(list(d.values()), crs=crs, validate_crs=False)
```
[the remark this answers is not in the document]
jmilloy: I think so.

If the crs comes from the DataArray attrs, then presumably it's already been validated. Can we validate if the user specifies the crs and skip if it comes from the DataArray, or do you need to skip it in both cases for the optimization?

```python3
validate_crs = False
if isinstance(x, (xr.DataArray, xr.Dataset)):
    xcoords = x.coords
    if crs is None:
        crs = x.attrs.get("crs")
        validate_crs = True
```

Or, could we provide `validate_crs` as a keyword argument to `from_xarray`, so you can disable it when necessary (like `from_geotransform` below)?

Ultimately, I don't we ever *have* to validate crs at instantiation. We could turn that off everywhere. If it's going to fail, it will fail when transforming during eval, and we can handle that exception then and there.

podpac/core/coordinates/coordinates.py
jmilloy: Yeah... hmmm. I'm sort of surprised about the caching bugs, and very concerned about introducing them. I'm surprised because `cached_property` doesn't use the podpac cache (right?), it just uses a private attribute in the class. Can you demo what kind of bug you were seeing?

What part is slow? Is it the hash_alg, or hex_digest, or what? I wonder if we can locally cache the results of the hash_alg or whatever the sticking point is in order to avoid the slow down and the bugs. E.g....

```python3
hash_alg_cache = {}
...
def cached_hash_alg(s):
    global hash_alg_cache
    if s not in hash_alg_cache:
        hash_alg_cache[s] = hash_alg(s)
    return hash_alg_cache[s]
```

podpac/core/coordinates/uniform_coordinates1d.py
jmilloy: Yeah, I think so.

podpac/core/interpolation/interpolation_manager.py
```diff
@@ -485,7 +485,10 @@ def select_coordinates(self, source_coordinates, eval_coordinates, index_type="n
                 selected_coords_idx[d] = np.arange(source_coordinates[d].size)[selected_coords_idx[d]]
 
         selected_coords = Coordinates(
-            [selected_coords[k] for k in source_coordinates.dims], source_coordinates.dims, crs=source_coordinates.crs
+            [selected_coords[k] for k in source_coordinates.dims],
+            source_coordinates.dims,
+            crs=source_coordinates.crs,
+            validate_crs=False,
```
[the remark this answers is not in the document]
jmilloy: Agreed.

jmilloy (commented):
I didn't look through the rasterio changes. Everything else seems okay, with a few comments that we can discuss now or maybe leave for after the release, up to you.

podpac/core/coordinates/coordinates.py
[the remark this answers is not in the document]
mpu-creare: Yes, I'd rather pass it as a flag and have the user make the decision. By default I'll make it False. Thanks! Great suggestion.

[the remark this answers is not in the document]
mpu-creare: I suppose we could also make that a setting so we can globally disable crs validation for optimization reasons...

podpac/core/coordinates/coordinates.py
mpu-creare: I didn't see any actual bugs... but I recall when making ui's it's convenient to overwrite certain traits (in which case the user supposedly knows what they are doing). But in those cases no modification actually changed the output of eval because it was just grabbing data from cache (because the hash stayed the same... in my memory example it was actually the definition that was cached.)... I'm not sure exactly what part was slow, but caching this saved 0.5 a second on a request.

podpac/core/compositor/compositor.py
mpu-creare: You always do a much better job with names than I do. Certainly I'll make that change.

PR #485, opened 2021-08-02
Feature: get_source_data

### LINKED ISSUE #468: get source data for Datasource and TileCompositors
### Description

It should be easy to get source data (for given bounds) for Datasources and TileCompositors.

### Current

You can get source data from any `Datasource` using the source `coordinates`:

```python3
# all source data
node.eval(node.coordinates)

# within bounds
coords = node.coordinates.select(my_bounds)
data = node.eval(coords)
```

### Proposal

I propose a method to get source data for `Datasource` *and* `TileCompositor` nodes (optionally given bounds for some or all dimensions).

```python3
# all source data
data = node.get_source_data()

# within bounds
data = node.get_source_data(my_bounds)
```

### Implementation

A universal `Datasource.get_source_data` can be implemented that selects and evaluates the source `coordinates`, but the value here (in addition to a more direct API) is that specific subclasses can and should define a more direct implementation whenever possible that avoids the evaluattion and interpolation machinery completely.

jmilloy:
### Datasource

```python3
>>> node = podpac.data.Array(
...             source=np.ones((3, 4)),
...             coordinates=podpac.Coordinates([range(3), range(4)], ["lat", "lon"]),
...         )
>>> 
>>> node.get_source_data()
<xarray.UnitsDataArray (lat: 3, lon: 4)>
array([[1., 1., 1., 1.],
       [1., 1., 1., 1.],
       [1., 1., 1., 1.]])
Coordinates:
  * lat      (lat) float64 0.0 1.0 2.0
  * lon      (lon) float64 0.0 1.0 2.0 3.0
Attributes:
    layer_style:   <podpac.core.style.Style object at 0x7f7c09fcb730>
    crs:           +proj=longlat +datum=WGS84 +no_defs +vunits=m
    geotransform:  (-0.5, 1.0, 0.0, -0.5, 0.0, 1.0)

```

### TileCompositor

```python3
>>> a = ArrayRaw(source=np.arange(5) + 100, coordinates=podpac.Coordinates([[0, 1, 2, 3, 4]], dims=["lat"]))
>>> b = ArrayRaw(source=np.arange(5) + 200, coordinates=podpac.Coordinates([[5, 6, 7, 8, 9]], dims=["lat"]))
>>> c = ArrayRaw(source=np.arange(5) + 300, coordinates=podpac.Coordinates([[10, 11, 12, 13, 14]], dims=["lat"]))
>>> node = TileCompositorRaw(sources=[a, b, c])
>>> node.get_source_data()
<xarray.UnitsDataArray (lat: 15)>
array([100., 101., 102., 103., 104., 200., 201., 202., 203., 204., 300.,
       301., 302., 303., 304.])
Coordinates:
  * lat      (lat) float64 0.0 1.0 2.0 3.0 4.0 5.0 ... 10.0 11.0 12.0 13.0 14.0
Attributes:
    layer_style:  <podpac.core.style.Style object at 0x7f7bb627a340>
    crs:          +proj=longlat +datum=WGS84 +no_defs +vunits=m
```

### Bounds

```python3
>>> node.get_source_data({"lon": (1.5, 4.5)})
<xarray.UnitsDataArray (lat: 3, lon: 2)>
array([[1., 1.],
       [1., 1.],
       [1., 1.]])
Coordinates:
  * lat      (lat) float64 0.0 1.0 2.0
  * lon      (lon) float64 2.0 3.0
Attributes:
    layer_style:   <podpac.core.style.Style object at 0x7f7bb6566190>
    crs:           +proj=longlat +datum=WGS84 +no_defs +vunits=m
    geotransform:  (1.5, 1.0, 0.0, -0.5, 0.0, 1.0)
```

closes #468

### DIFF
--- a/doc/source/overview.md
+++ b/doc/source/overview.md
@@ -59,6 +59,16 @@ node = podpac.datalib.TerrainTiles(tile_format='geotiff', zoom=8)
 # ... and more each release
 ```
 
+Retrieve the raw source data array at full/native resolution.
+
+```python
+# retrieve full source data
+node.get_source_data()
+
+# retrieve bounded source data
+node.get_source_data(bounds={'lat': (40, 45), 'lon': (-70, -75)})
+```
+
 ## Coordinates
 
 Define geospatial and temporal dataset coordinates.
--- a/doc/source/overview.md
+++ b/doc/source/overview.md
@@ -59,7 +59,7 @@ node = podpac.datalib.TerrainTiles(tile_format='geotiff', zoom=8)
 # ... and more each release
 ```
 
-Retrieve the raw source data array at full/native resolution.
+Retrieve the raw source data array at full/native resolution. **Note**: Some data source are too large to fit in RAM, and calling this function can crash Python. 
 
 ```python
 # retrieve full source data
--- a/podpac/core/compositor/tile_compositor.py
+++ b/podpac/core/compositor/tile_compositor.py
@@ -83,7 +83,7 @@ class TileCompositorRaw(BaseCompositor):
         """
 
         if any(not hasattr(source, "get_source_data") for source in self.sources):
-            raise ValueError("Cannot get composited source data; all sources must be a DataSource or TileCompositor")
+            raise ValueError("Cannot get composited source data; all sources must have `get_source_data` implemented (such as nodes derived from a DataSource or TileCompositor node).")
 
         coords = None  # n/a
         source_data_arrays = (source.get_source_data(bounds) for source in self.sources)  # generator
--- a/podpac/core/compositor/tile_compositor.py
+++ b/podpac/core/compositor/tile_compositor.py
@@ -67,6 +67,28 @@ class TileCompositorRaw(BaseCompositor):
             return result
         return res
 
+    def get_source_data(self, bounds={}):
+        """
+        Get composited source data, without interpolation.
+
+        Arguments
+        ---------
+        bounds : dict
+            Dictionary of bounds by dimension, optional.
+
+        Returns
+        -------
+        data : UnitsDataArray
+            Source data
+        """
+
+        if any(not hasattr(source, "get_source_data") for source in self.sources):
+            raise ValueError("Cannot get composited source data; all sources must be a DataSource or TileCompositor")
+
+        coords = None  # n/a
+        source_data_arrays = (source.get_source_data(bounds) for source in self.sources)  # generator
+        return self.composite(coords, source_data_arrays)
+
 
 class TileCompositor(InterpolationMixin, TileCompositorRaw):
     pass
--- a/podpac/core/compositor/tile_compositor.py
+++ b/podpac/core/compositor/tile_compositor.py
@@ -75,6 +75,7 @@ class TileCompositorRaw(BaseCompositor):
         ---------
         bounds : dict
             Dictionary of bounds by dimension, optional.
+            Keys must be dimension names, and values are (min, max) tuples, e.g. ``{'lat': (10, 20)}``.
 
         Returns
         -------
@@ -83,7 +84,9 @@ class TileCompositorRaw(BaseCompositor):
         """
 
         if any(not hasattr(source, "get_source_data") for source in self.sources):
-            raise ValueError("Cannot get composited source data; all sources must have `get_source_data` implemented (such as nodes derived from a DataSource or TileCompositor node).")
+            raise ValueError(
+                "Cannot get composited source data; all sources must have `get_source_data` implemented (such as nodes derived from a DataSource or TileCompositor node)."
+            )
 
         coords = None  # n/a
         source_data_arrays = (source.get_source_data(bounds) for source in self.sources)  # generator
--- a/podpac/core/data/datasource.py
+++ b/podpac/core/data/datasource.py
@@ -289,6 +289,7 @@ class DataSource(Node):
         ---------
         bounds : dict
             Dictionary of bounds by dimension, optional.
+            Keys must be dimension names, and values are (min, max) tuples, e.g. ``{'lat': (10, 20)}``.
 
         Returns
         -------
--- a/podpac/core/data/datasource.py
+++ b/podpac/core/data/datasource.py
@@ -281,6 +281,12 @@ class DataSource(Node):
     # Methods
     # ------------------------------------------------------------------------------------------------------------------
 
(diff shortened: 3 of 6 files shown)


jmilloy:
@mpu-creare This is a simple enhancement. It is just a general method implemented in the `DataSource` class and the `TileCompositor` class. Neither method calls `eval`, so it is a true enhancement and not just an API wrapper. However, some child datasource or datalib nodes could probably override `get_source_data` to be more efficient, such as perhaps the `Rasterio` node. At this point, I haven't done that (we can do that as needed as optimization), but I want to know if you would prefer to do some of that for this PR.

doc/source/overview.md
```diff
@@ -59,6 +59,16 @@ node = podpac.datalib.TerrainTiles(tile_format='geotiff', zoom=8)
...
 
+Retrieve the raw source data array at full/native resolution.
+
+```python
+# retrieve full source data
+node.get_source_data()
+
+# retrieve bounded source data
+node.get_source_data(bounds={'lat': (40, 45), 'lon': (-70, -75)})
+```
```
mpu-creare, line 70: Thanks for updating the documentation.

doc/source/overview.md
```diff
@@ -59,6 +59,16 @@ node = podpac.datalib.TerrainTiles(tile_format='geotiff', zoom=8)
 # ... and more each release
 ```
 
+Retrieve the raw source data array at full/native resolution.
```
mpu-creare, line 62:
```suggestion
Retrieve the raw source data array at full/native resolution. **Note**: Some data source are too large to fit in RAM, and calling this function can crash Python. 
```

podpac/core/compositor/tile_compositor.py
```diff
@@ -67,6 +67,28 @@ def composite(self, coordinates, data_arrays, result=None):
...
+            Dictionary of bounds by dimension, optional.
+
+        Returns
+        -------
+        data : UnitsDataArray
+            Source data
+        """
+
+        if any(not hasattr(source, "get_source_data") for source in self.sources):
+            raise ValueError("Cannot get composited source data; all sources must be a DataSource or TileCompositor")
```
mpu-creare, line 86:
```suggestion
            raise ValueError("Cannot get composited source data; all sources must have `get_source_data` implemented (such as nodes derived from a DataSource or TileCompositor node).")
```

podpac/core/data/datasource.py
```diff
@@ -281,6 +281,24 @@ def _get_data(self, rc, rci):
...
+        bounds : dict
+            Dictionary of bounds by dimension, optional.
+
+        Returns
+        -------
+        data : UnitsDataArray
+            Source data
+        """
+
+        coords, I = self.coordinates.select(bounds, return_index=True)
```
mpu-creare, line 299: We could add a test here on the size of `I` and throw an logger warning in case of requests that exceed something like 8 GB? What do you think?

podpac/core/data/datasource.py
```diff
@@ -281,6 +281,24 @@ def _get_data(self, rc, rci):
...
     # ------------------------------------------------------------------------------------------------------------------
 
+    def get_source_data(self, bounds={}):
+        """
+        Get source data, without interpolation.
+
+        Arguments
+        ---------
+        bounds : dict
+            Dictionary of bounds by dimension, optional.
```
mpu-creare, line 291: It looks like the bound have to be in the format `{"dim", (min, max)}` and `(max,min)` does not work). Is that correct?

podpac/core/data/datasource.py
```diff
@@ -281,6 +281,24 @@ def _get_data(self, rc, rci):
...
+
+        Returns
+        -------
+        data : UnitsDataArray
+            Source data
+        """
+
+        coords, I = self.coordinates.select(bounds, return_index=True)
+        return self._get_data(coords, I)
```
mpu-creare, line 300: This is a great implementation. I think it's also very safe and general. I looked to see if any source over-writes `_get_data` and only the ogc & ogr sources do. The ogc data source should work fine, but the ogr one has a completely different method signature -- not sure if you want to handle that here...

mpu-creare (approved):
I'm fine approving this. I did have a few suggestions on wording and potential improvements, but I'll leave it up to you. 

The major thing is the OGR node which will NOT work with this implementation. It will throw an exception. Any ideas?

podpac/core/data/datasource.py
jmilloy: Doesn't numpy throw it's own warning if you try to make an array that is too large?

podpac/core/data/datasource.py
jmilloy: cc59d811

podpac/core/data/datasource.py
jmilloy: I was just walking the dogs and decided that I should implement `get_source_data` itself for the WCS node for this PR. You're right about the ogr node, too. In fact, it subclasses `Node`, not `Datasource`, so doesn't even have `get_source_data defined`.

jmilloy: Well, it's not really possible for the ogc node. There is no native resolution.

jmilloy: I think it's best to leave both ogr and ogc how they are. The only thing I might do is implement the ogc get_source_data in order to throw a more user-friendly exception.

podpac/core/data/datasource.py
jmilloy:
```
>>> np.empty((10000, 1000, 1000))
numpy.core._exceptions.MemoryError: Unable to allocate 74.5 GiB for an array with shape (10000, 1000, 1000) and data type float64
```

mpu-creare: That's a good point, but I don't think all the underlying sources use numpy (I'm thinking rasterio) and the error might not get thrown at the index level since the indices might be orthogonal.

podpac/core/data/datasource.py
mpu-creare: That makes sense. I didn't realize they didn't inherit from DataSource. Although, the OGC source sort-of has a native resolution which *may* be provided by the server.

jmilloy: My mistake. The ogc node is fine: it does inherit from Datasource and we do pull the native resolution. It's not possible for the ogr node. That's the one that has no native coordinates.

PR #486, opened 2021-08-03
ENH: Node get_bounds method.

### LINKED ISSUE #467: Node bounds attribute
The bounds for a node should be easily accessible.

We should implement a universal `Node.bounds` property that uses `find_coordinates`, and then override where possible with more efficient properties, for example `Datasource.bounds` property that uses `coordinates` and `Rasterio.bounds` property that uses `dataset`. Specific datalib wrappers should further override when possible, especially large `TileCompositors` or remote datasources where the bounds are known.

In addition, `find_coordinates` should be implemented in child classes where appropriate, particularly in large `TileCompositors` and remote datasources. Wherever `find_coordinates` can be hardcoded or otherwise implemented efficiently in child classes, the base class `bounds` property can be used as-is.

jmilloy:
## Any node

```python
    def test_get_bounds(self):
        class MyNode(Node):
            def find_coordinates(self):
                return [
                    podpac.Coordinates([[0, 1, 2], [0, 10, 20]], dims=["lat", "lon"], crs="EPSG:2193"),
                    podpac.Coordinates([[3, 4], [30, 40]], dims=["lat", "lon"], crs="EPSG:2193"),
                ]

        node = MyNode()

        with podpac.settings:
            podpac.settings["DEFAULT_CRS"] = "EPSG:4326"

            # specify crs
            bounds, crs = node.get_bounds(crs="EPSG:2193")
            assert bounds == {"lat": (0, 4), "lon": (0, 40)}
            assert crs == "EPSG:2193"

            # default crs
            bounds, crs = node.get_bounds()
            assert bounds == {
                "lat": (-75.81397534013118, -75.81362774074242),
                "lon": (82.92787904584206, 82.9280189659297),
            }
            assert crs == "EPSG:4326"
```

## Special case for Datasources

```python
    def test_get_bounds(self):
        node = podpac.data.Array(
            source=np.ones((3, 4)),
            coordinates=podpac.Coordinates([range(3), range(4)], ["lat", "lon"], crs="EPSG:2193"),
        )

        with podpac.settings:
            podpac.settings["DEFAULT_CRS"] = "EPSG:4326"

            # specify crs
            bounds, crs = node.get_bounds(crs="EPSG:3857")
            assert bounds == {
                "lat": (-13291827.558247399, -13291815.707967814),
                "lon": (9231489.26794932, 9231497.142754894),
            }
            assert crs == "EPSG:3857"

            # native/source crs
            bounds, crs = node.get_bounds(crs="source")
            assert bounds == {"lat": (0, 2), "lon": (0, 3)}
            assert crs == "EPSG:2193"

            # default crs
            bounds, crs = node.get_bounds()
            assert bounds == {
                "lat": (-75.81365382984804, -75.81362774074242),
                "lon": (82.92787904584206, 82.92794978642414),
            }
            assert crs == "EPSG:4326"
```

closes #467

### DIFF
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -353,6 +353,37 @@ class Node(tl.HasTraits):
 
         raise NotImplementedError
 
+    def get_bounds(self, crs="default"):
+        """Get the full available coordinate bounds for the Node.
+
+        Arguments
+        ---------
+        crs : str
+            Desired CRS for the bounds.
+            If not specified, the default CRS in the podpac settings is used. Optional.
+
+        Returns
+        -------
+        bounds : dict
+            Bounds for each dimension. Keys are dimension names and values are tuples (hi, lo).
+        crs : str
+            The CRS for the bounds.
+        """
+
+        if crs == "default":
+            crs = podpac.settings["DEFAULT_CRS"]
+
+        bounds = {}
+        for coords in self.find_coordinates():
+            ct = coords.transform(crs)
+            for dim, (lo, hi) in ct.bounds.items():
+                if dim not in bounds:
+                    bounds[dim] = (lo, hi)
+                else:
+                    bounds[dim] = (min(lo, bounds[dim][0]), max(hi, bounds[dim][1]))
+
+        return bounds, crs
+
     @common_doc(COMMON_DOC)
     def create_output_array(self, coords, data=np.nan, attrs=None, **kwargs):
         """
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -365,7 +365,7 @@ class Node(tl.HasTraits):
         Returns
         -------
         bounds : dict
-            Bounds for each dimension. Keys are dimension names and values are tuples (hi, lo).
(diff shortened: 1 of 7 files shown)


podpac/core/node.py
```diff
@@ -353,6 +353,37 @@ def find_coordinates(self):
...
+        Arguments
+        ---------
+        crs : str
+            Desired CRS for the bounds.
+            If not specified, the default CRS in the podpac settings is used. Optional.
+
+        Returns
+        -------
+        bounds : dict
+            Bounds for each dimension. Keys are dimension names and values are tuples (hi, lo).
```
mpu-creare, line 368:
```suggestion
            Bounds for each dimension. Keys are dimension names and values are tuples (min, max).
```

mpu-creare (approved):
Nice! Just fix a few of the doc strings and you can go ahead and merge.

podpac/core/node.py
mpu-creare: @jmilloy It looks like you're actually returning (lo, hi), not (hi, lo) as the comment mentions. Also above.

PR #493, opened 2022-02-08
Feature/fortify scan fixes

mpu-creare:
This is a large, unorganized update driven by specific project needs. 

# Changes
* Removed the `dist` folder -- no longer supporting easy lambda function setup. This *may* eventually be done through a Docker file, but in any case we'll make a separate repo for it if there is demand

# Features
* Improved error message on coord_select node when that node is unable to infer the coordinate step size
* Node prober can now handle nodes with multiple values for a single point (i.e. because of an `ExpandCoordinates` node for statistical operations)
* `Convolution` can now handle requests with dimensions not in the source
* `Convolution` can now handle sources with dimensions not in the kernel_dims
* `Convolution` node can now extend `ArrayCoordinates1d` to do edges properly -- it does assume a uniform step in the extended direction

# Bug Fixes
* Added `Percentile` to list of algorithm classes
* There was a tricky floating point error in the `signal.py` file. When you made an xarray from `UniformCoordinates` and then re-make the `UniformCoordinates` from the xarray, the coordinates are mismatched by floating point precision. This causes xarray to no longer match the coordinates -- which causes all kinds of problems. 
* `Convolution` now works for `ArrayCoordinates1d` with only 1 element


mpu-creare:
@BAyotte could you please review this PR?

BAyotte (approved):
Looks good to me. I just had a couple of very minor suggestions. Still not entirely sure what I am doing but I am selecting approve even with some of my comments outstanding.

podpac/core/algorithm/coord_select.py
```diff
@@ -225,15 +225,21 @@ def get_modified_coordinates1d(self, coords, dim):
             # no selection in this dimension
             return coords1d
 
-        if len(selection) == 1:
+        if len(selection) == 1 or ((len(selection) == 2) and (selection[0] == selection[1])):
```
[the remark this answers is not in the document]
mpu-creare: The format for `selection` in this case is `[start_offset, end_offset, step]` where the last two entries are optional. When we create these nodes programmatically, it sometimes happens that `end_offset == start_offset`, which is essentially a single coordinate value. The new part of the if statement handles this case.

podpac/core/algorithm/signal.py
```diff
@@ -59,14 +59,15 @@ class Convolution(UnaryAlgorithm):
...
     kernel : np.ndarray, optional
         The convolution kernel. This kernel must include the dimensions of source node outputs. The dimensions for this
-        array are labelled by `kernel_dims`. Any dimensions not in the soucr nodes outputs will be summed over.
+        array are labelled by `kernel_dims`. Any dimensions not in the source nodes outputs will be summed over.
     kernel_dims : list, optional
-        A list of the dimensions for the kernel axes. The dimensions in this list must match the
-        coordinates in the source, or contain additional dimensions, and the order does not need to match.
-        Any extra dimensions are summed out.
+        A list of the dimensions for the kernel axes. If the dimensions in this list do not match the
+        coordinates in the source, then any extra dimensions in the kernel are summed out and any extra
```
mpu-creare, line 65:
```suggestion
        coordinates in the source, then any extra dimensions in the kernel are removed by adding all the values over that axis
```

mpu-creare: @BAyotte does that make more sense?

podpac/core/algorithm/signal.py
```diff
@@ -59,14 +59,15 @@ class Convolution(UnaryAlgorithm):
...
+        dimensions in the source are not convolved with any kernel.
+
     kernel_type : str, optional
         If kernel is not defined, kernel_type will create a kernel based on the inputs, and it will have the
-        same number of axes as kernel_ndim.
+        same number of axes as kernel_dim.
```
mpu-creare, line 70:
```suggestion
        same number of axes as kernel_dims.
```

podpac/core/test/test_authentication.py
```diff
@@ -34,17 +35,17 @@ def test_set_credentials(self):
             # test input/getpass
             # TODO: how do you test this?
 
-            # set both username and password
-            set_credentials(hostname="test.com", username="testuser", password="testpass")
+            # set both username and pw
+            set_credentials(hostname="test.com", uname="testuser", password="testpass")
```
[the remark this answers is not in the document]
mpu-creare: I think (strangely) the `password` in the comment was flagged as a problem by the static code analyzer, but the one in the code was not. From memory, I think the issue was "password added to comments in source file" or something like that. :)

Issue #494, opened 2022-07-26
Fix JSON data format


mpu-creare:
**Description**
The PODPAC JSON format for pipelines is NOT (in fact) JSON. PODPAC requires ORDERING of the key-value pairs -- this worked through `Python` using the `OrderedDict` trait. While working in `JavaScript` we couldn't consistently serialize ordered pairs, which led to actually reading the JSON spec. Whoops. 

**Expected Behavior**
The `PODPAC` `JSON` pipeline format should NOT rely on ordering of keys.

mpu-creare:
Original JSON (relies on the output Node `Arithmetic` to be the LAST key defined):
```json
{
    "SinCoords": {
        "node": "core.algorithm.utility.SinCoords"
    },
    "Arange": {
        "node": "core.algorithm.utility.Arange"
    },
    "Arithmetic": {
        "node": "core.algorithm.generic.Arithmetic",
        "attrs": {
            "eqn": "a+b",
            "params": {}
        },
        "inputs": {
            "a": "SinCoords",
            "b": "Arange"
        }
    }
}
```

## Possible fixes
### Use a list
```json
{   "nodes": [
        {
            "node_name": "SinCoords",
            "node": "core.algorithm.utility.SinCoords"
        },
        {
            "node_name": "Arange",
            "node": "core.algorithm.utility.Arange"
        },
        {
            "node_name": "Arithmetic",
            "node": "core.algorithm.generic.Arithmetic",
            "attrs": {
                "eqn": "a+b",
                "params": {}
            },
            "inputs": {
                "a": "SinCoords",
                "b": "Arange"
            }
        }
    ]
}
```

* Upsides: 
    * Now follows json Spec
* Downsides
    * Harder to work with once it's in Python -- you have to loop through all elements in the list to find the "node_name". Requires a bigger refactor on the `podpac` side. 
    * Not really backwards compatible

### Add an ordered list entry in JSON
```json
{
    "Arithmetic": {
        "node": "core.algorithm.generic.Arithmetic",
        "attrs": {
            "eqn": "a+b",
            "params": {}
        },
        "inputs": {
            "a": "SinCoords",
            "b": "Arange"
        }
    },
    "SinCoords": {
        "node": "core.algorithm.utility.SinCoords"
    },
    "Arange": {
        "node": "core.algorithm.utility.Arange"
    },
    "podpac_order": ["SinCoords", "Arange", "Arithmetic"]
}
```
* Upsides: 
    * Now follows json Spec
[...]

PR #501, opened 2023-01-17
Feature/json fix

CFoye-Creare:
Addresses issue #494.

### DIFF
--- a/ex_json.JSON
+++ b/ex_json.JSON
@@ -1,0 +1,29 @@
+{
+	"Arithmetic": {
+		"node": "core.algorithm.generic.Arithmetic",
+		"attrs": {
+			"eqn": "a+b",
+			"params": {
+
+			}
+		},
+		"inputs": {
+			"a": "SinCoords",
+			"b": "Arange"
+		}
+	},
+    "SinCoords": {
+		"node": "core.algorithm.utility.SinCoords",
+		"style": {
+			"colormap": "jet",
+			"clim": [
+				-1.0,
+				1.0
+			]
+		}
+	},
+	"Arange": {
+		"node": "core.algorithm.utility.Arange"
+	},
+	"podpac_version": "3.2.0"
+}
--- a/podpac/core/node.py
+++ b/podpac/core/node.py
@@ -875,10 +875,10 @@ class Node(tl.HasTraits):
                 kwargs[k] = v
 
             for k, v in d.get("inputs", {}).items():
-                kwargs[k] = _lookup_input(nodes, name, v)
+                kwargs[k], nodes = _lookup_input(nodes, name, v, definition)
 
             for k, v in d.get("lookup_attrs", {}).items():
-                kwargs[k] = _lookup_attr(nodes, name, v)
+                kwargs[k] = _lookup_attr(nodes, name, v, definition)
 
             if "style" in d:
                 style_class = getattr(node_class, "style", Style)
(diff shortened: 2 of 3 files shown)


podpac/core/node.py
```diff
@@ -1208,33 +1208,97 @@ def get_ui_spec(cls, help_as_html=False):
...
         )
+    # node not yet discovered yet
+    if not value in nodes:
+        # Look for it in the definition items:
+        for found_name, d in definition.items():
+            if value != found_name:
+                continue
+
+             # get node class
+            module_root = d.get("plugin", "podpac")
```
mpu-creare, line 1233: Is the code below here copied and pasted from somewhere else?

podpac/core/node.py
```diff
@@ -1208,33 +1208,97 @@ def get_ui_spec(cls, help_as_html=False):
...
+            nodes[found_name] = node_class(**kwargs)
+            # found!
+            break
 
     if not value in nodes:
         raise ValueError(
             "Invalid definition for node '%s': reference to nonexistent node '%s' in inputs" % (name, value)
         )
-
+    print(value)
```
mpu-creare, line 1294: Errant print statement

ex_json.JSON
```diff
@@ -0,0 +1,29 @@
+{
```
mpu-creare, line 1: Is this file being used anywhere?

mpu-creare (changes requested):
Looks like this will be a great backwards compatible solution, but let's discuss the code structure a bit before proceeding.

podpac/core/node.py
CFoye-Creare: Yes. The code from `from_definition` worked here as well, with a few modifications. Happy to make a function to encapsulate this loop and call it in both locations.?

CFoye-Creare: Attempted to fix this with commit 3fb5048ef08d00fd69e95d05343ebe02db73fc98

ex_json.JSON
CFoye-Creare: Deleted

podpac/core/node.py
CFoye-Creare: Deleted in 3fb5048ef08d00fd69e95d05343ebe02db73fc98

CFoye-Creare:
Added `podpac_output_node` functionality in 00f053004d786edffe9a3aef432c08051181389a

CFoye-Creare:
Added a unit test and checked for missing node in 9d1b5e9d3b069d29ace4b8c27e85ac31e707076c for `podpac_output_node` functionality.

CFoye-Creare:
fixed bug here: 33ff5aaa4f206bc365ba561ab43a7d4885ecdf0e

podpac/core/node.py
```diff
@@ -863,6 +863,11 @@ def from_definition(cls, definition):
         if output_node is None:
             return list(nodes.values())[-1]
         
+
+        if output_node not in nodes:
+                raise ValueError(
+                    "Invalid definition for value 'podpac_output_node': reference to nonexistent node '%s' in lookup_attrs" % (output_node)
+                )
```
mpu-creare, line 870: Good job adding the error message.

podpac/core/node.py
```diff
@@ -1258,7 +1263,7 @@ def _process_kwargs(name, d, definition, nodes):
         kwargs[k] = v
 
     for k, v in d.get("inputs", {}).items():
-        kwargs[k], nodes = _lookup_input(nodes, name, v, definition)
+        kwargs[k]= _lookup_input(nodes, name, v, definition)
```
mpu-creare, line 1266: does `_lookup_input` change nodes? I do prefer returning both if that's the case. Otherwise a comment in the code would be quite helpful.

mpu-creare (approved):
I'd add a comment or two in the code, but I tried this out. Seems to work great. You can merge when ready.

CFoye-Creare:
Will add some comments and then merge.

PR #503, opened 2023-01-20
API Audit

CFoye-Creare:


CFoye-Creare:
Running tests after changing all appropriate modules to plural.

CFoye-Creare:
merging from develop undid previous changes. reinstating refactoring of modules.

CFoye-Creare:
All tests that should pass do.

mpu-creare:
Also, I think the `version` module should be an exception? It should be singular, because that seems to be the common convention across Python libraries. @jxm, care to weigh in?

CFoye-Creare:
I renamed modules inside podpac.* to plural, but changed modules inside podpac.*.* to singular. So, `podpac.interpolator.py` became `podpac.interpolators.py` while `podpac.interpolation.interpolator.interpolator.py` was kept singular. Want me to change `podpac.interpolation.*` to `podpac.interpolator`?

Issue #502, opened 2023-01-20
Helper Function to determine resolution of coordinates


mpu-creare:
**Description**
When working with lat/lon or other projected coordinates, it's hard to know the real nominal resolution of your coordinate system.

**Describe the solution you'd like**
I want a helper function on coordinates that give me the resolution. 

**Sample Code**
``` python
>>> coords = podpac.Coordinates([...], dims=['lat', 'lon']
>>> coords.horizontal_resolution(units='m', type='nominal')
{"lat": 10,
"lon": 20, 
"units": "m"
}
```

mpu-creare:
This code might be useful: https://github.com/creare-com/pydem/blob/master/pydem/utils.py#L145

mpu-creare:
This might also be useful: https://pyproj4.github.io/pyproj/stable/api/crs/crs.html

CFoye-Creare:
Left to-do
- [ ] Finish Unit Test
- [ ] Write code to convert degrees to metres
- [ ] Stacked Conversion
- [ ] Unstacked Conversion
- [ ] Use podpac.units implementation

CFoye-Creare:
- [ ] cKDTree with boxsize wraparound implementation for stacked NearestNeighbor

mpu-creare:
For `coords.horizontal_resolution(units='m', type='nominal')`, let's talk about the different types

types = 
* "nominal" <-- this is wrong but cheap to calculate. Give a 'nominal' resolution over the entire domain
* "summary" <-- This is still approximate, gives the mean and standard deviation to each point
* "full" <-- min distance between every point (dx, dy, basically)
* "full_exact" <-- uses brute-force for the stacked coordinates

This is a proposal, feel free to propose something better.

CFoye-Creare:
Implemented 3 of the 4 types:

- "nominal" <-- this is wrong but cheap to calculate. Give a 'nominal' resolution over the entire domain
- "summary" <--- Returns average and std.dev for each point.
- "full" <--- returns the exact distance between every point, or every dx,dy if coordinates are unstacked

CFoye-Creare:
9851b1cb5ce4b32bc00e96e500ffc8c78efd7376

CFoye-Creare:
e05a662688b5d600aeb29784d43b5acb8c466cbe

CFoye-Creare:
[...]

PR #505, opened 2023-01-30
Add Coordinate Resolution

CFoye-Creare:
# Description
New feature branch to address issue #502. We are looking to implement a helper function to return the resolution of coordinate systems at different levels of accuracy.
# Proposed Changes
For a `Coordinate` object, a method `horizontal_resolution` to return the resolution of horizontal coordinate system. This should support both stacked and unstacked coordinates. We are offering three kinds of resolutions:
- "nominal" <-- this is wrong but cheap to calculate. Give a 'nominal' resolution over the entire domain
- "summary" <--- Returns average and std.dev
- "full" <--- returns the exact distance between every point, or every dx, dy if coordinates are unstacked
We also let the user specify their desired return units. We support any unit that `podpac.units` supports.
## Stacked Coordinates
If the horizontal coordinates are stacked, then we should return information based upon the set of points.
- "nominal" resolution: returns the average min distance from each point to the next
- "summary" resolution: returns the average min distance from each point to the next, and the standard deviation of the min distances
- "full" resolution: returns every distance between every point in a distance matrix
## Unstacked Coordiantes
If the horizontal coordinates are unstacked, then we should return information based upon the grid space
- "nominal" resolution: returns the average dx,dy.
- "summary" resolution: returns average dx,dy and their standard deviations
- "full" resolution: returns the exact dx,dy of each grid square.


CFoye-Creare:
Currently assumes the inputted data will be in valid lat/lon degrees

podpac/core/coordinates/coordinates.py
```diff
@@ -19,6 +19,10 @@
 from six import string_types
 import pyproj
 import logging
+from scipy import spatial
+from geopy.distance import geodesic
```
mpu-creare, line 23: Whenever you add a new package, you MUST add it to the setup.py file. Otherwise I can't run the code. I'll make this fix. 

Also, when it's an optional dependency, it's better to lazy-load the function or module. In this case I think the dependency should be part of the core functionality.

podpac/core/coordinates/coordinates.py
```diff
@@ -19,6 +19,10 @@
 from six import string_types
 import pyproj
 import logging
+from scipy import spatial
+from geopy.distance import geodesic
+import math
+import utm
```
mpu-creare, line 25: Same comment about adding an import as above.

podpac/core/coordinates/coordinates.py
```diff
@@ -1501,12 +1505,152 @@ def issubset(self, other):
...
+    '''
+    Calculate horizontal resolution of coordinate system.
+    Assumes horizontal coords are in degrees.
+    
+    Support different types!
+        - "nominal" <-- this is wrong but cheap to calculate. Give a 'nominal' resolution over the entire domain
+        - "summary" <-- This is still approximate, gives the mean and standard deviation to each point
+        - "full" <-- min distance between every point (dx, dy, basically)
+    '''
+    def horizontal_resolution(self, units = "metre", type="nominal"):
```
mpu-creare, line 1526: We use the Numpy style docstrings. Please reformat your doc strings to comply with those.

mpu-creare: Also looks like you're not using the `black` formatting git hook... I'll help get that set up for you.

podpac/core/coordinates/test/test_coordinates.py
```diff
@@ -2202,3 +2204,40 @@ def test_affine_to_uniform(self):
 
         # unstacked uniform -> unstacked uniform
         assert c3.simplify() == c3
+
+
+class TestResolutions(object):
+    def test_resolution(self):
+        
+        '''Unstacked Coordinates'''
+        # Nominal
```
mpu-creare, line 2213: Please add some tests for me to review.

mpu-creare (changes requested):
Please make the requested changes and then I'll review again.

mpu-creare:
Could you also edit the MR and add a short description of the changes so we can remember what they were next time we look at this. Also add a link to the relevant issue.

podpac/core/coordinates/test/test_coordinates.py
CFoye-Creare: Sounds good.

podpac/core/coordinates/coordinates.py
```diff
@@ -1501,12 +1505,254 @@ def issubset(self, other):
...
+            dim : str
+                The dimension to return the resolution of. Should be unstacked.
+
+            Returns
+            -------
+            tuple
+                the average distance between grid squares
+                the standard deviation of those distances
+            """
+            diff = np.zeros(len(self[dim].coordinates) - 1)
```
mpu-creare, line 1689: again, vectorization here would be good.

mpu-creare (changes requested):
Pretty nice implementation -- I'm asking for two changes: 
1. if else instead of match to support Python 3.9
2. vectorization of some operations (not strictly necessary but a nice learning opportunity.)

CFoye-Creare:
See https://github.com/creare-com/podpac/issues/502#issuecomment-1425955172 for some info on how I implemented the unstacked coords this time.

jmilloy:
I haven't been following this but I get all the notificatons and took a brief look today. I'm not going to make a thorough review, but I notice that this new method has logical cases for different types of coordinates which is something I tried pretty hard to avoid when developing the coordinates module. So something like this would be more in line with the style of the rest of the module:

```python3
def horizontal_resolution(self, units, type):
    resolution_dict = {}
    for dim, c in self.items():
        if 'lat' in c.dims or 'lon' in c.dims:
            resolution_dict[dim] = c.horizontal_resolution(units, type)
```

or maybe

```python3
def horizonal_resolution(self, units, type):
    return {dim: c.horizontal_resolution(units, type) for dim, c in self.items() if c.is_horizontal}
```

Then the four methods are defined each in Coordinates1d or StackedCoordinates instead. Similarly, I think there are some benefits to then testing these methods individually in test_coordinates1d and test_stacked_coordinates and then lightly testing the Coordinates horizontal_resolution class that pulls them together. Anyways, just a note.

mpu-creare:
@jmilloy this would typically work but to get the resolution you need both the latitude and longitude, even in the unstacked case -- so it can't be handled in `Coordinates1d`. Unless you have a better suggestion? 

I really do appreciate the note on the design.

jmilloy:
@mpu-creare Yeah, I wondered about that and wasn't thorough enough. I'd consider passing the lat in, which would ignored in some cases, but make it very clear what's needed.

```
for dim, c in self.items():
    resolution[dim] = c.horizontal_resolution(units, type, lat=self['lat'])
```

In fact, you can technically make Coordinates with a lon dimension but not a lat dimension, right? And then calling horizontal_resolution on coordinates without a 'lat' should be caught and raised as an error, right? I didn't look super closely but I don't think it is. I think making it it explicit helps point that out and reminds you to test for it.

jmilloy:
(Also, don't use `type` as an argument/variable)

CFoye-Creare:
@jmilloy `restype` work?

CFoye-Creare:
Wrote a check for unstacked "lat" or unstacked "lon" without a counterpart. Now throwing a ValueError. Also deleted some lingering commented code, and changed the "type" parameter to "restype".

CFoye-Creare:
I realized that the lat value actually matters for longitude as well, because the distances are greater near the equator due to the bump at the equator.

podpac/core/coordinates/coordinates.py
```diff
@@ -1790,28 +1738,37 @@ def full_unstacked_resolution(dim):
...
             # Is the dim lat/lon?
             if not check_horizontal(dim):
                 continue
+            # Put this check inside the loop to avoid checking if stacked coords
+            if dim == "lat":
+                if "lon" not in self.dims:
+                    return ValueError("Need both lat and lon for resolution.")
+            if dim == "lon":
+                if "lat" not in self.dims:
+                    return ValueError("Need both lat and lon for resolution.")
```
jmilloy, line 1754: Hmmm, I would just check `if lat not in self.udims or lon not in self.udims` once outside of the loop.

What if the dims are `['lat_time']`? That case should fail. Or `['lat_time', 'lon']` which should work? I'm not sure your loop and functions handle the `['lat_time', 'lon']` case at all. I know these cases seem weird, maybe you guys can decide not to handle them. @mpu-creare are these sorts of coordinates still allowed? @CFoye-Creare Am I right that this case isn't handled?

Something like this would be more general:

```
def horizontal_resolution(self, units, restype):
    if 'lat' not in self.udims or 'lon' not in self.udims:
        raise ValueError("horizontal_resolution requires lat and lon dimensions.")
    d = {}
    for dim, c in self.items():
        if 'lat' in c.dims and 'lon' in c.dims:
            d[dim] = stacked_horizontal_resolution(self['lat'], self['lon'], units, restype)
        elif 'lat' in c.dims:
            d[dim] = unstacked_lat_horiziontal_resolution(self['lat'], self['lon'], units, restype)
        elif 'lon' in c.dims:
            d[dim] = unstacked_lon_horiziontal_resolution(self['lat'], self['lon'], units, restype)
    return d
```

I still prefer a version that moves those functions to the Coordinates1d and StackedCoordinates classes, but regardless, at least passing the coordinates in as an argument is better in principle because it makes the behavior explicit rather than implicit. And at least moving the helper functions out of horizontal_resolution would be better because then you can't use the other dimention in the "wrong scope", plus you can unit test them. In general, the inline functions lead you to make assumptions that aren't true and makes readability and testing harder.

CFoye-Creare: I agree. We talked about it this morning and are going to go with the 1dCoordinates and StackedCoordinates implementation.

CFoye-Creare: I'm not sure I understand how the `'lat_time'` case would work. It was my understanding that we only on horizontal resolution.

CFoye-Creare: Got it.

podpac/core/coordinates/coordinates.py
```diff
@@ -1501,12 +1504,77 @@ def issubset(self, other):
 
         return all(c.issubset(other) for c in self.values())
 
-    def is_stacked(self, dim):
-        if dim not in self.udims:
+    def is_stacked(self, dim):  # re-wrote to be able to iterate through c.dims
```
mpu-creare, line 1507: Is there a reason you're not using the individual coordinate's `is_stacked` method?

podpac/core/coordinates/coordinates1d.py
```diff
@@ -423,3 +429,118 @@ def issubset(self, other):
...
+            Returns
+            --------
+            The average distance between each grid square for this dimension
+            """
+            if self.name == "lat":
+                return (
+                    calculate_distance(
+                        (self.bounds[0], 0), (self.bounds[1], 0), ellipsoid_tuple, coordinate_name, units
+                    ).magnitude
+                    / self.size
```
mpu-creare, line 472: Gotta watch those fence-posts! should be (self.size -1 )

mpu-creare:
@CFoye-Creare Nice first PR. Thanks for navigating all the trickiness behind this implementation. Merging. 

Closes #502 .

Issue #512, opened 2023-09-07
Allow custom coordinate dimensions


mpu-creare:
**Description**
Allow downstream applications to specify arbitrary coordinate dimensions.

**Describe the solution you'd like**
I might have an application where I want to pull data from a dataset based on coordinates with dimensions other than ["lat", "lon", "alt", "time"]

**Sample Code**
```
import podpac
podpac.coordinates.add_dimension('my_custom_dimension')
coords = podpac.Coordinates([[1, 2, 3]], dims=["my_custom_dimension"])
node = MyCustomNode()  # Defined elsewhere
o = node.eval(coords)
```

mpu-creare:
Probably relevant: https://github.com/creare-com/podpac/blob/develop/podpac/core/coordinates/utils.py#L503

CFoye-Creare:
I want to create a utility function `add_valid_dimension`, so we can add a dimension to the list of valid dimensions. I wonder if we want a `rm_valid_dimension` function with protections on `lat`, `lon`, `time`, `alt`

CFoye-Creare:
See https://github.com/creare-com/podpac/blob/develop/podpac/core/common_test_utils.py. This is an odd example of hardcoding. This only seems to be used for testing the coordinate creation? I don't really understand the point of this test or why we would make a hardcoded list of permutations and not use a library (itertools.permutations for example)

mpu-creare:
This is just fine because it lives in the test, and the base `podpac` library will never have any dimensions other than these. The reason we didn't use itertools is because of combinations of stacked dimensions. Perhaps there was a way to do it, but it was faster to just do a manually.

CFoye-Creare:
Would like to meet about this. 

- [ ] It seems the main updating we have to do is include dataset keys for non-default dimension. It seems like Dataset has a "selection" key that supports other dims, but it isn't clear to me how widespread this is. Would like to talk about it.
- [ ] The "polar coordinates" object currently hardcode checks for 'lat/lon' as the dimension. Wondering if we should keep this/what it is intending and a way we can do it without hardcode check.

CFoye-Creare:
Both seem okay, and a change is not needed right now.

CFoye-Creare:
Import of `VALID_DIMENSION_NAMES` seems to be updating.
```
In [1]: from podpac.core.coordinates.utils import VALID_DIMENSION_NAMES
INFO:numexpr.utils:NumExpr defaulting to 8 threads.

[...]

PR #513, opened 2023-09-12
CODE: add_valid_dimension utility function #512

CFoye-Creare:
See issue #512. We want to allow downstream applications to specify arbitrary coordinate dimensions.


CFoye-Creare:
Incorrect rebasing. Cherry picked commits onto `feature/custom_coord_dims`

PR #527, opened 2024-10-10
HOTFIX Release 3.5.2

dhinckley-creare:
This relates to #525 

# Release

How to release `podpac`

## Branch Notes
There are currently two actively maintained versions of PODPAC:
* 3.x version which uses main-3.X and develop-3.X
* 4.x version which uses main and develop

The 3.x is maintained for projects that have not migrated to the new 4.x interface. 

The repercussions of this are as follows:
* Most of the code bases are still compatible, so any new features or bug fixes can go to both branches
* Process is: Feature or hotfix branch off develop-3.X, then when done coding, merge into both develop-3.X and develop

## Updating main

1. Ensure your local main / main-3.X branch is synced to upstream:

```bash
$ git pull upstream main  # or just git pull
```

2. Create a release branch: `release/3.x.y`

3. Update [`version.py`](podpac/version.py) `MAJOR`, `MINOR`, and `HOTFIX` to the right semantic version

4. Run unit tests

```bash
$ # Activate Python environment
$ pytest podpac             
```

4. Review the [CHANGELOG](CHANGELOG.md) and update
    - Prefix should be a [numpy commit prefix](https://docs.scipy.org/doc/numpy/dev/gitwash/development_workflow.html#writing-the-commit-message)
    - Convention:
```markdown
- <prefix>: <short description> ([<github issue>](https://github.com/creare-com/podpac/issues/<issue#>))
```

5. Merge changes into main / main-3.X branch

6. On the main branch, Tag the release:

```bash
$ git tag -a X.Y.Z -m 'X.Y.Z'
```

6. Push your changes to main / main-3.X:

```bash
$ git push upstream main
$ # OR 
$ git push upstream main-3.X
$ git push upstream --tags
```

7. Build source and binary wheels for pypi (you have to have the `wheels` package installed):

```bash
$ git clean -xdf  # this deletes all uncommited changes!
$ python setup.py bdist_wheel sdist
```

8. Upload package to [TestPypi](https://packaging.python.org/guides/using-testpypi/). You will need to be listed as a package owner at
https://pypi.python.org/pypi/podpac for this to work. You now need to use a pypi generated token, can no longer use your password. 

```bash
$ twine upload --repository-url https://test.pypi.org/legacy/ dist/podpac-X.Y.Z*
```

9. Use twine to register and upload the release on pypi. Be careful, you can't
take this back! You will need to be listed as a package owner at
https://pypi.python.org/pypi/podpac for this to work.

```bash
$ twine upload dist/podpac-X.Y.Z*
```

11. Issue the release announcement (via github)

# Checklist
* [x] Ensure your local main / main-3.X branch is synced to upstream:
* [x] Create a release branch: `release/3.x.y`
* [x] Update [`version.py`](podpac/version.py) `MAJOR`, `MINOR`, and `HOTFIX` to the right semantic version
* [x] **Run unit tests** and **fix anything that doesn't pass**
* [x] Review the [CHANGELOG](CHANGELOG.md) and update
* [x] Merge changes into main / main-3.X branch
* [x] On the main branch, Tag the release.
* [x] Push your changes to main / main-3.X:
* [x] Build source and binary wheels for pypi (you have to have the `wheels` package installed):
* [x] Upload package to pypi. Be careful, you can't take this back! * Issue the release announcement (via github)


podpac/core/algorithm/signal.py
```diff
@@ -246,7 +246,13 @@ def _make_kernel(kernel_type, ndim):
             k = np.ones([size] * ndim)
         else:
             args = [float(a) for a in kernel_type.split(",")[2:]]
-            f = getattr(scipy.signal, ktype)
+            if ktype=='gaussian':
```
dhinckley-creare: @mpu-creare given the nature of the change, this seemed a little more stable than the try-except but correct me if I'm wrong.

podpac/core/coordinates/test/test_coordinates1d.py
```diff
@@ -108,15 +111,22 @@ def test_horizontal_resolution(self):
...
+        lat_compare = (
+                4442569.935968436 * podpac.units("meter"),
+                13040.905617921147 * podpac.units("meter"),
+            )
+        lon_compare = (
+                5558704.3695234 * podpac.units("meter"),
+                3399219.0171971265 * podpac.units("meter"),
+            )
+        for i in range(2):
+            assert lat_summary[i].units==lat_compare[i].units
```
dhinckley-creare, line 126: @mpu-creare the approximate compare couldn't be applied to the `pint.utils.Quantity` directly, this was my attempt at achieving the same result while also respecting units (the units of assignment, not converting to a common unit).

podpac/core/data/test/test_datasource.py
```diff
@@ -546,6 +546,8 @@ def test_eval_get_cache_extra_dims(self):
     def test_eval_get_cache_transform_crs(self):
         with podpac.settings:
             podpac.settings["DEFAULT_CACHE"] = ["ram"]
+            podpac.settings["DEFAULT_CRS"] = podpac.core.settings.DEFAULT_SETTINGS["DEFAULT_CRS"]
```
dhinckley-creare, line 549: @mpu-creare this was added as a precaution given that this function deals with CRS but it didn't resolve the test's failure.

This test still fails at ln560 `assert not node._from_cache` and while I understand that it shouldn't fail at that line (the node should not have originated from cache) I can't puzzle out the exact reason it fails.

@rparker-creare runs this test without it failing so I would suspect something related to python environment as the cause of that difference.

podpac/core/algorithm/signal.py
mpu-creare: Agreed. 

I think you can get rid of the outer `if ktype==gaussian`, the `hasattr` check is good enough. That way, I bet this fix will apply to more windows.

podpac/core/coordinates/test/test_coordinates1d.py
mpu-creare: Great fix.

podpac/core/data/test/test_datasource.py
mpu-creare: Mostly like a difference in settings between you and @rparker-creare  -- I would start looking there.

mpu-creare: Probably related to one of the caching settings.

podpac/core/algorithm/signal.py
dhinckley-creare: Roger that.

podpac/core/data/test/test_datasource.py
dhinckley-creare: The issue seems to be what DEFAULT_CRS setting you have in your `/home/<account>/.config/podpac/settings.json` file.

When I have `"EPSG:4326"`, the test fails but both `"+proj=longlat +datum=WGS84 +no_defs +vunits=m"` and `"EPSG:25832"` work.

@mpu-creare can you explain why might this be? My guess is it has something to do with some operations done using the default crs on initialization that are causing the nodes to be cached before expected (hence the test failure) but I don't understand enough about the crs' to guess what that might be.

mpu-creare: Hmm... this is a bit mysterious to me. I think it would require digging to figure out what's going on.

dhinckley-creare: Is resolving this something to be done before the release? Would it be improper for there be a means to prevent the default settings file adoption process and should the unit tests exercise it?

dhinckley-creare: This issue has been resolved. The problem was caused by another unit test creating the same node which caused it to be cached if the default CRS matched that of the test. The test has been changed to define a different node, preventing the issue regardless of CRS setting.

rparker-creare (approved):

podpac/core/algorithm/algorithm.py
```diff
@@ -130,6 +130,9 @@ def f(node):
                 inputs[key] = node.eval(coordinates, output=output, _selector=_selector)
             self._multi_threaded = False
 
+        if settings["ALGORITHM_XARRAY_FLOATING_POINT_CORRECTION"]:
```
mpu-creare: I wouldn't do this SOLELY through a global setting. It's best if individual nodes can over-ride this behavior if needed. 

```suggestion
        if self.xarray_floating_point_correction is None:
             if settings["ALGORITHM_XARRAY_FLOATING_POINT_CORRECTION"]:
                  inputs = align_xarray_dict(inputs)
       elif self.xarray_floating_point_correction:
```

mpu-creare: You'll have to add the traited attribute `xarray_floating_point_correction` to the algorithm class as well.

podpac/core/algorithm/signal.py
```diff
@@ -246,7 +246,10 @@ def _make_kernel(kernel_type, ndim):
             k = np.ones([size] * ndim)
         else:
             args = [float(a) for a in kernel_type.split(",")[2:]]
-            f = getattr(scipy.signal, ktype)
+            if hasattr(scipy.signal,ktype):
```
mpu-creare: This if block looks INSANE without context -- please comment the code so future developers can know why you're doing this.

podpac/core/test/test_utils.py
```diff
@@ -757,3 +758,39 @@ def test_prober_with_enumerated_legends(self):
...
+    
+
+    inputs = align_xarray_dict(inputs)
+
+    for k in ['lat','lon']:
+        assert(np.all(inputs['A'][k]==inputs['B'][k]))
+        assert(np.all(inputs['A'][k]==inputs['C'][k]))
+    assert(np.all(inputs['A'].data==data_a))
+    assert(np.all(inputs['B'].data==data_b))
+    assert(np.all(inputs['C'].data==data_c))
```
mpu-creare, line 796: I would add an assert to test the use-case as it ends up in an algorithm... something like: 
```
assert (np.all(np.array((a + b + c).shape) == a.shape))
```

mpu-creare (changes requested):
I found a few things worth changing. Pretty close though. Thanks for fixing all those tests!

podpac/core/test/test_utils.py
dhinckley-creare: Added in recent push. Due to object behavior, I had to reference the versions that existed inside the inputs dict instead of the originals.

podpac/core/algorithm/signal.py
dhinckley-creare: Comment added

podpac/core/algorithm/algorithm.py
dhinckley-creare: I put the trait in the Algorithm class and it defaults to the global settings value. Will this cause issues with settings load behavior? Would the settings every get loaded after this trait is used in normal operations?

mpu-creare: Much better fix than my suggestion. I think that's actually what we typically do in PODPAC.

mpu-creare (approved):

PR #529, opened 2025-01-15
Option to use AWS defaults to login to S3

jdw-creare:
This PR introduces a podpac config file option called `AWS_GET_AUTH_FROM_ENV`, affecting the S3-backed node types `ZarrRaw` and `RasterioRaw` (as well as their derivatives).  If absent or set to `false`, authentication behaves exactly as it did before. If present and set to `true`, S3 authentication will fall back to the `boto3` default methods for finding authentication.  These methods are detailed [in the boto3 documentation](https://boto3.amazonaws.com/v1/documentation/api/latest/guide/credentials.html) - in short, `boto3` will check the system environment variables, then fall back through several AWS CLI configuration files.

I found that when I used `rasterio` version 1.3.8, `RasterioRaw` was unable to access S3 even when the credentials were working fine.  Upgrading to `rasterio` version 1.4.3 solved the issue.

### DIFF
--- a/podpac/core/authentication.py
+++ b/podpac/core/authentication.py
@@ -190,6 +190,7 @@ class S3Mixin(tl.HasTraits):
     """Mixin to add S3 credentials and access to a Node."""
 
     anon = tl.Bool(False).tag(attr=True)
+    aws_get_auth_from_env = tl.Bool()
     aws_access_key_id = tl.Unicode(allow_none=True)
     aws_secret_access_key = tl.Unicode(allow_none=True)
     aws_region_name = tl.Unicode(allow_none=True)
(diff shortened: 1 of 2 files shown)


mpu-creare (approved):
Changes look pretty unobtrusive. I don't have any concerns. nice work. 

This looks like the feature is branched off develop -- it will have to get into develop3.X as well (as a reminder).

jdw-creare:
> This looks like the feature is branched off develop -- it will have to get into develop3.X as well (as a reminder).

Thanks @mpu-creare ! I almost accidentally merged into `develop`, that wouldn't've been great.

PR #532, opened 2025-05-15
Fix sonarqube findings

jdw-creare:
This PR fixes some "blocker" level findings in SonarQube.  

The most contentious change here is probably the change to `coordinates.py`.  Sonarqube [took issue with](https://sonarcloud.io/project/issues?impactSeverities=BLOCKER&issueStatuses=OPEN%2CCONFIRMED&id=creare-com_podpac&open=AZFxutjSkYO_0R1F0Oz-) having both a property called `CRS` and a data member called `crs` (differing only by case).  I could not find anything in Creare code referencing the `CRS` property.  So I changed the `CRS` property to `pyproj_crs` and  updated its test. 

We should think a little about whether this counts as an API change.  Depending on how closely we want to follow the [semantic versioning guidelines](https://semver.org/), we might want to update the major version number in whatever release includes this PR.

I think we will have PRs following this one to address Sonarqube findings that are categorized as High or Medium, so we may have more changes like this later that affect our decision to bump the major version number or not.

### DIFF
--- a/podpac/core/test/test_authentication.py
+++ b/podpac/core/test/test_authentication.py
@@ -19,9 +19,6 @@ class TestAuthentication(object):
                 del settings["password@test.com"]
 
             # require hostname
-            with pytest.raises(TypeError):
-                set_credentials()
-
             with pytest.raises(ValueError):
                 set_credentials(None, uname="test", password="test")
 
--- a/podpac/core/cache/file_cache_store.py
+++ b/podpac/core/cache/file_cache_store.py
@@ -308,7 +308,7 @@ class FileCacheStore(CacheStore):
     # file storage abstraction
     # -----------------------------------------------------------------------------------------------------------------
 
-    def _save(self, path, s):
+    def _save(self, path, s, metadata=None):
         raise NotImplementedError
 
     def _load(self, path):
--- a/podpac/core/coordinates/coordinates.py
+++ b/podpac/core/coordinates/coordinates.py
@@ -682,7 +682,7 @@ class Coordinates(tl.HasTraits):
 
         # properties
         # TODO check transform instead
-        if self.CRS != other.CRS:
+        if self.pyproj_crs != other.pyproj_crs:
             return False
 
         # full check of underlying coordinates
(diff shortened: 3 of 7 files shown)


jdw-creare:
Hey @mpu-creare , do you have time to take a look at this for me?  I don't believe these changes will affect external users, but it's hard for me to tell for sure.

mpu-creare:
@jdw-creare can it wait till next week?

jdw-creare:
@mpu-creare it absolutely can!  Thanks Matt!

podpac/core/test/test_authentication.py
```diff
@@ -19,9 +19,6 @@ def test_set_credentials(self):
                 del settings["password@test.com"]
 
             # require hostname
-            with pytest.raises(TypeError):
-                set_credentials()
```
mpu-creare, line 23: What's the reason for removing this unit test? Was it broken?

mpu-creare (approved):
Seems fine. I don't think the "CRS" property is widely used, at least I don't think I use it frequently. the `pyproj_crs` is a more descriptive name anyway, so a good change.

podpac/core/test/test_authentication.py
jdw-creare: It worked fine, just SonarQube was flagging line 23 as missing required arguments.

kxm-creare: Would another way to address this be to pass `None` to the test?  It's funny SonarQube is complaining about "mistakes" when this is exactly the behavior we want to test - that the code catches mistakes.

Something to consider in the future.

jdw-creare: Right, I totally see a couple of valid perspectives here.  I think I tried passing in `None`, and it raised a `ValueError`.

A whole argument went through my head that ended in me removing these lines:

* So on the one side: you're absolutely right, this test (lines 22 and 23) confirm that a mistake raises an error.  Which is, of course, a good thing, and a core part of good testing.

* Another perspective: yes it's an error, but the error is _outside_ of `set_credentials()` - python itself raises the `TypeError` before calling into `set_credentials()`, our code does nothing to raise that error.  Therefore lines 22 and 23 aren't testing our code, they're testing the python interpreter itself.  And we don't really need to confirm that it's still working.

* A response: But why bother to remove those lines, their inclusion hurts nothing?

* An answer: because SonarQube says so, and we have a customer requirement to pass SonarQube scans.  And it's not asking anything particularly egregious of us.

jmilloy: I haven't looked at the code, but I think you're totally right here John. If this was an exception that our code raised, then I think we would want to keep it (if for no other reason than code coverage, which sonarqube also checks), and I suspect sonarqube wouldn't flag it either. But since it's not covering any of our code, we don't need to test it.

PR #533, opened 2025-05-28
Fix low-impact sonarqube findings

jdw-creare:

PR #534, opened 2025-06-04
Fixed sonarqube findings on the develop branch, by bringing in fixes from the `develop-3.X` branch.

jdw-creare:
Fixed sonarqube findings on the develop branch, by bringing in fixes from the `develop-3.X` branch.

I took the changes from these PRs:
* #531
* #532
* #533

which were made on the branch `develop-3.X`, and cherry-picked their changes into the `develop` branch.  Naturally there were some merge conflicts with that, which I then resolved.

There were 4 failing unit tests prior to this change, and the same 4 tests fail after this change, while the pass count remains at 960.

I also noticed some unit tests with a `skip` mark, which were being erroneously skipped.  Fixing the skip mark adds 5 passing tests.

### DIFF
--- a/podpac/core/data/test/test_csv.py
+++ b/podpac/core/data/test/test_csv.py
@@ -5,6 +5,8 @@ import numpy as np
 
 from podpac.core.data.csv_source import CSV
 
+_CRS = "+proj=merc +vunits=m"
+INVALID_DATA_KEY = "Invalid data_key"
 
 class TestCSV(object):
     """test csv data source"""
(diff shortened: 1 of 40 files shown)


podpac/core/data/test/test_csv.py
```diff
@@ -190,7 +195,7 @@ def test_get_data_multiple(self):
         np.testing.assert_array_equal(out.sel(output="b"), self.other)
 
         # default
-        node = CSV(source=self.source_multiple, alt_key="altitude", crs="+proj=merc +vunits=m")
+        node = CSV(source=self.source_multiple, alt_key="altitude", crs=_CRS)
```
mpu-creare, line 198: Just a curiosity -- why didn't it flat "altitude"

mpu-creare (approved):
Looks the same as the v3 side -- approved!

podpac/core/data/test/test_csv.py
jdw-creare: I found that confusing as well!  From the SonarQube documentation:

No issue will be raised on:
* duplicated string in decorators
* strings with less than 5 characters
* strings with only letters, numbers and underscores

The exception makes some sense, since you wouldn't want to flag things like `mydict['key']`.  On the other hand, we clearly aren't addressing the spirit of the rule by overlooking `"altitude"` - it's a constant just as much as that CRS string.  On the gripping hand: normally we'd exclude test files from sonarqube scans anyway, so here we just want to do the bare minimum to pass deployment requirements.

jdw-creare:
Thanks matt!!

PR #535, opened 2025-06-18
Fixes for "high" sonarqube findings, mostly cognitive complexity

jdw-creare:
Lots of substantial changes in this PR:
- The majority of these are refactors of methods, to bring their cognitive complexity down to 25 or less.
- I've tried to make each commit contain a single method refactor (or related change).  So the easiest way to review this may be a single commit at a time.

Some notes on unit tests:
- When I started this branch, there were 4 failing unit tests.
- I didn't want to creep the scope of this PR, so I did not fix those tests in this PR.
- During this work, I found what seems like a bug, worked around it, and added a failing unit test to address that bug (See [`test_add_custom_dimension()`](https://github.com/creare-com/podpac/blob/445267a982ab50c128bca74f78c01ac847897869/podpac/core/coordinates/test/test_coordinates.py#L638),  
- I also found that one file of test cases had syntax errors, so the test collection skipped them entirely.  I'm not sure how this happened - some indentation changed, possibly during a merge of one of my previous PRs.  When I corrected the indentation, it added about 25 working tests and 5 failing tests. 
- I didn't want to make this huge PR any bigger, so I left those tests failing.
- All in all, I have added 6 new failing unit tests, bringing the total fail count to 10.  All should be addressed soonish and I don't mind being the one to fix them, I just wanted to get this PR wrapped up before it creeps further.
- The quality gate fails because it introduces new issues at the "Medium" and "low" level.  I plan to address the "medium" issues in bulk during a future PR.


podpac/core/utils.py
```diff
@@ -14,10 +14,8 @@
 from copy import deepcopy
 from hashlib import sha256 as hash_alg
 
-try:
-    import urllib.parse as urllib
-except:  # Python 2.7
-    import urlparse as urllib
+import urllib.parse as urllib
```
scranford1, line 17: Are we dropping support for Python2.7 here? (Should we add a python requirement in setup.py?)

podpac/core/algorithm/test/test_coord_select.py
```diff
@@ -16,6 +16,9 @@ def setup_module(module):
     COORDS = podpac.Coordinates(
         ["2017-09-01", podpac.clinspace(45, 66, 4), podpac.clinspace(-80, -70, 5)], dims=["time", "lat", "lon"]
     )
+    global _rand
+    # Set up the PRNG with a seed to stay deterministic
+    _rand = np.random.default_rng(0xC * ord('r') + 0xea + ord('r') * 0xe)
```
scranford1, line 21: Out of curiosity is there a specific reason for this value?

podpac/core/algorithm/stats.py
```diff
@@ -1079,24 +1079,65 @@ class DayOfYearWindow(Algorithm):
...
+
+        dims = ["lat", "lon", "alt"]
+        for alt in alt_f:
+            for lat in lat_f:
+                for lon in lon_f:
+                    # _log.debug(f'lat, lon, alt = {lat}, {lon}, {alt})
+                    loc_dict = {k: v for k, v in zip(dims, [lat, lon, alt]) if v is not None}
+
+                    data = source.sel(time=time_in_window_flags, **loc_dict).dropna("time").data
+                    if not np.all(np.isnan(data)):
```
scranford1: May just be personal preference but thoughts on `if np.any(~np.isnan(data))`?

podpac/core/coordinates/base_coordinates.py
```diff
@@ -125,7 +125,6 @@ def __eq__(self, other):
         return NotImplemented
 
     # python 2 compatibility
-    if sys.version < "3":
-
+    if sys.version_info[0] < 3:
```
scranford1: Using .major may be nice for readability (a couple other places).

podpac/core/interpolation/interpolation_manager.py
```diff
@@ -601,11 +685,7 @@ def interpolate(self, source_coordinates, source_data, eval_coordinates, output_
         # TODO handle stacked issubset of unstacked case
         #      this case is currently skipped because of the set(eval_coordinates) == set(source_coordinates)))
         if eval_coordinates.issubset(source_coordinates) and set(eval_coordinates) == set(source_coordinates):
-            if any(isinstance(c, StackedCoordinates) and c.ndim > 1 for c in eval_coordinates.values()):
-                # TODO AFFINE
-                # currently this is bypassing the short-circuit in the shaped stacked coordinates case
-                pass
-            else:
+            if not any(isinstance(c, StackedCoordinates) and c.ndim > 1 for c in eval_coordinates.values()):
```
scranford1: Do we need to update this to have `or c.ndim <= 1`?

scranford1 (commented):
Looks great to me! Just a few comments and then some additional questions (out of curiosity).

podpac/core/algorithm/stats.py
jdw-creare: I like that; seems like it'd work the same without needing to check as many values.

podpac/core/algorithm/test/test_coord_select.py
jdw-creare: Since the seed just has to be an arbitrary fixed number, and `0` or `1` felt boring, I wanted to try to put something clever in but didn't get very far. 

My first thought was `0xCREARE` but since of course R isn't a hex digit, I went with `ord('r')`.  Then it seemed boring to just sum them up, so I multiplied some of them to make a slightly larger arbitrary number.  

It's really all quite arbitrary and I'm all ears if you can think of something cooler!

podpac/core/utils.py
jdw-creare: Great idea to add the python requirement!  

My plan was indeed to stop supporting Python 2.7 with this change.  I figure since python 2.7 is now [6 years past end of life](https://devguide.python.org/versions/), we can safely drop support (and this particular compatibility method gets flagged by sonarqube as a defect).  I know academic projects and others like that still probably use versions of Python 2, but I feel like it is better at this stage to be part of the pressure to upgrade projects rather than keep trying to maintain backwards compatibility.

We've also been running all our tests with python 3, so it's possible we've introduced code dependent on python 3 features already anyway.

podpac/core/interpolation/interpolation_manager.py
jdw-creare: Great question! i'll look at it

jdw-creare: You're absolutely right; great catch!

jdw-creare:
@scranford1 , I've finished incorporating fixes based on your notes.  Thanks for the review!

podpac/core/algorithm/stats.py
```diff
@@ -1127,45 +1168,18 @@ def algorithm(self, inputs, coordinates):
             # If either the start or end runs over the year, we need to do an OR on the bool index
             # ----->s....<=e------   .in -out
             # ..<=e----------->s..
-            do_or = False
-
-            start = doy - win
-            if start < 1:
-                start += 365
-                do_or = True
+            window_is_split = doy - win < 1 or doy + win > self._DAYS_PER_YEAR
```
mpu-creare: Even though eliminating the more verbose 'if' statements, I would argue this actually increases MY cognitive complexity.

podpac/core/cache/zarr_cache.py
```diff
@@ -275,7 +275,7 @@ def subselect_has(self, request_coords):
         if np.all(bool_data):
             return None  # or any other indicator that all data is present
 
-        false_indices = np.where(bool_data == False)
+        false_indices = np.nonzero(bool_data == False)
```
mpu-creare, line 278: Is there a good reason for this somewhat arbitrary change? Just curious.

podpac/core/coordinates/base_coordinates.py
```diff
@@ -125,7 +125,6 @@ def __eq__(self, other):
         return NotImplemented
 
     # python 2 compatibility
-    if sys.version < "3":
-
+    if sys.version_info.major < 3:
```
mpu-creare: We can probably drop Python 2 compatibility entirely, so could just remove this branch entirely. In fact, I doubt that the current version of PODPAC would work in Python 2.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -455,33 +452,35 @@ def select(self, bounds, outer=False, return_index=False):
...
 
+    def _and_indices(self, indices: List["Coordinates"]):
+        """logical AND of the selection in each dimension"""
         if all(isinstance(index, slice) for index in indices):
             index = slice(max(index.start or 0 for index in indices), min(index.stop or self.size for index in indices))
             # for consistency
             if index.start == 0 and index.stop == self.size:
                 if self.ndim > 1:
-                    index = [slice(None, None) for dim in self.dims]
+                    index = [slice(None, None)] * len(self.dims)
```
mpu-creare: Oooh... there be dragons. These two codes do not do the same thing. I'd have to carefully look at surrounding code to convince myself that this doesn't introduce a bug. Here's an example of the problem -- the slice instance is the same for all list elements in the new code. So, change index[0], changes all of them. 

```python
In [1]: a = [slice(None) for _ in range(2)]

In [2]: b = [slice(None)] * 2

In [3]: id(a[0]), id(a[1])
Out[3]: (126093725304704, 126093674613888)

In [4]: id(b[0]), id(b[1])
Out[4]: (126093674614336, 126093674614336)
```

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -496,7 +495,7 @@ def _index_len(index):
             # for consistency
             if np.all(index):
                 if self.ndim > 1:
-                    index = [slice(None, None) for dim in self.dims]
+                    index = [slice(None, None)] * len(self.dims)
```
mpu-creare: Another case, see comment above.

podpac/core/interpolation/interpolation_manager.py
```diff
@@ -601,11 +685,7 @@ def interpolate(self, source_coordinates, source_data, eval_coordinates, output_
         # TODO handle stacked issubset of unstacked case
         #      this case is currently skipped because of the set(eval_coordinates) == set(source_coordinates)))
         if eval_coordinates.issubset(source_coordinates) and set(eval_coordinates) == set(source_coordinates):
-            if any(isinstance(c, StackedCoordinates) and c.ndim > 1 for c in eval_coordinates.values()):
-                # TODO AFFINE
-                # currently this is bypassing the short-circuit in the shaped stacked coordinates case
-                pass
-            else:
+            if not any(isinstance(c, StackedCoordinates) or c.ndim <= 1 for c in eval_coordinates.values()):
```
mpu-creare, line 688: This refactor eliminated a TODO... not sure if the TODO will ever be done... but it might have been a helpful hint to a future developer -- or not. It's not helpful to me right now, so who know. 

Just a note.

podpac/core/interpolation/test/test_interpolation_manager.py
```diff
@@ -174,7 +174,7 @@ def test_init_interpolators(self):
             InterpolationManager([{"method": "nearest", "params": {"spatial_tolerance": "tol"}}])
 
         with pytest.raises(AttributeError):
-            assert interp.config[("default",)]["interpolators"][0].myarg == "tol"
+            interp.config[("default",)]["interpolators"][0].myarg
```
mpu-creare, line 177: Why are we no longer checking for the correct value?

mpu-creare (changes requested):
Yep. Looks like a refactor. Mostly code moved around and.. well.. refactored. 

My only notable comment was the dragon one. Please consider that one carefully.

podpac/core/interpolation/test/test_interpolation_manager.py
jdw-creare: I wouldn't have noticed this, but SonarQube flagged it.  This call to `interp.config()` is supposed to raise (and does raise) an `AttributeError`.  So it never returns, and the `assert ... == "tol"` check is never evaluated.  The `with pytest.raises(AttributeError):` bit catches the  `AttributeError` and the test continues.  So the `assert ` statement in this case ends up being another form of unreachable code.

podpac/core/interpolation/interpolation_manager.py
jdw-creare: I'd be happy to put it back in, but i couldn't tell what the TODO was saying should be done.  Like, I assume there's another short-circuit case when something is affine?  I'm not sure what would be affine here though.  Do you remember what it was for?  It looks like it's been 5+ years without any strong enough impetus to complete the todo, so it may not really matter either way.

podpac/core/interpolation/test/test_interpolation_manager.py
mpu-creare: HA! That's a great catch. +1 for SonarQube.

podpac/core/interpolation/interpolation_manager.py
mpu-creare: Agreed

podpac/core/cache/zarr_cache.py
jdw-creare: We spoke on the phone about this, so for any future readers, here's the backstory behind this change:

https://sonarcloud.io/organizations/creare-com/rules?open=python%3AS6729&rule_key=python%3AS6729

podpac/core/algorithm/stats.py
jdw-creare: Fixed!

PR #537, opened 2025-09-25
Sonarqube coverage reporting and migration to GitHub actions

scranford1:


.github/workflows/gitlab-python-worflow.yml
```diff
@@ -0,0 +1,132 @@
...
+
+      - name: Install Dependencies
+        run: |
+          pip install --upgrade pip setuptools && pip install .[devall]
+
+      - name: Run Pytest with Coverage
+        run: |
+          coverage run --branch \
+          -m pytest -m "not integration" --continue-on-collection-errors || \
+          { echo "WARNING: Some tests have failed. (Temporarily ignored)"; true; }
```
jdw-creare, line 63: I think as soon as we get unit tests to pass - and maybe they already do? we should take out this suppression line.  I'd love to do whatever we can to get people in the habit of expecting all the unit tests to pass, and fixing them any time they break, and requiring them to pass before a merge request goes in

jdw-creare (approved):
This looks great, thanks Sam!

.github/workflows/gitlab-python-worflow.yml
scranford1: Agree, I added the warning for now as we currently have the below results. But I can also remove it so that it will fail properly in the future.
`= 26 failed, 972 passed, 49 skipped, 9 deselected, 2 xfailed, 2115 warnings in 44.41s =`

jdw-creare: Makes sense!  Yeah feel free to leave it in now and we can take it out as soon as we get everything passing again

PR #539, opened 2026-01-20
General Bug Fixes

dhinckley-creare:
This branch contains various behavior fixes/adjustments that were discovered when working through the transition from podpac3 to podpac4 with another program that makes heavy use of podpac

### DIFF
--- a/podpac/core/data/datasource.py
+++ b/podpac/core/data/datasource.py
@@ -424,7 +424,7 @@ class DataSource(Node):
             # use the original coords if there was no intersection
             if rsc.size != 0:
                 temp_selector = Selector(method="nearest")
-                (rsc, rsci) = temp_selector.select(coordinates, rsc, index_type=self.coordinate_index_type)
+                (rsc, rsci) = temp_selector.select(self.coordinates, rsc, index_type=self.coordinate_index_type)
 
         # if requested coordinates and coordinates do not intersect, shortcut with nan UnitsDataArary
         if rsc.size == 0:
@@ -462,7 +462,6 @@ class DataSource(Node):
         # get indexed boundary
         rsb = self._get_boundary(rsci)
         output.attrs["boundary_data"] = rsb
-        output.attrs["bounds"] = self.coordinates.bounds
 
         # save output to private for debugging
         if settings["DEBUG"]:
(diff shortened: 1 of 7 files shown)


podpac/core/data/datasource.py
```diff
@@ -468,6 +474,36 @@ def _eval(self, coordinates, output=None, _selector=None):
 
         return output
 
+    @common_doc(COMMON_DATA_DOC)
+    def create_output_array(self, coords, data=np.nan, attrs=None, outputs=None, **kwargs):
+        """
+        Initialize an output data array. This adds bounds to the output attrs
```
mpu-creare:
```suggestion
        Initialize an output data array. This adds `bounds` and `boundary_data` to the output attrs
```

podpac/core/data/datasource.py
```diff
@@ -468,6 +474,36 @@ def _eval(self, coordinates, output=None, _selector=None):
...
+
+        The boundary_data output.attrs is set to match this node's boundary.
```
mpu-creare:
```suggestion
        The `boundary_data` output.attrs is set to match this node's polygonal (i.e. non-rectangular) boundary.
```

podpac/core/utils.py
```diff
@@ -515,28 +515,42 @@ def _get_entry(key, out, definition):
     entry = OrderedDict()
     entry["name"] = out[key]["name"]
     entry["value"] = str(out[key]["value"])
-    if out[key]["units"] not in [None, ""]:
-        entry["value"] = entry["value"] + " " + str(out[key]["units"])
+    # if out[key]["units"] not in [None, ""]:
+    #     entry["value"] = entry["value"] + " " + str(out[key]["units"])
+    entry['label'] = out[key]['label']
```
mpu-creare, line 519: Did we just get rid of units in the node prober?

podpac/core/utils.py
```diff
@@ -515,28 +515,42 @@ def _get_entry(key, out, definition):
...
     entry["params"] = {}
     entry["inputs"] = {"inputs": [_get_entry(inp, out, definition) for inp in out[key]["inputs"]]}
     if len(entry["inputs"]["inputs"]) == 0:
         entry["inputs"] = {}
     return entry
-
-def _format_value(value, style, add_enumeration_labels):
-    """Helper for probe_node()."""
+    
+def _get_label(value, style, add_enumeration_labels):
```
mpu-creare, line 531: Please comment this function. I don't know; 
* Why it exists
* What it is doing
* How it is doing it...

mpu-creare (changes requested):
A few minor comments. 

Major thing as you already guessed is the lack of code coverage. Please add some tests that cover the new code.

podpac/core/utils.py
dhinckley-creare: Label now represents both enumeration labels and units.

dhinckley-creare: This was part of a change to make it so the probe_node doesn't combine the value and units in the return structure. This allows for more flexibility by whomever calls the function (no string parsing a combined return).

podpac/core/utils.py
dhinckley-creare: Comment added

mpu-creare: SOO much better. Thanks!

dhinckley-creare:
@mpu-creare When adding unit tests to improve coverage, I noticed that [this catch](https://github.com/creare-com/podpac/blob/554fbc8a88c19a807508bae7df624f7b629a91f4/podpac/core/utils.py#L349) doesn't have a well defined return.

Should we return `None` like in the other catch?

mpu-creare:
> @mpu-creare When adding unit tests to improve coverage, I noticed that [this catch](https://github.com/creare-com/podpac/blob/554fbc8a88c19a807508bae7df624f7b629a91f4/podpac/core/utils.py#L349) doesn't have a well defined return.
> 
> Should we return `None` like in the other catch?

Very yes. Nice catch.

dhinckley-creare:
> > @mpu-creare When adding unit tests to improve coverage, I noticed that [this catch](https://github.com/creare-com/podpac/blob/554fbc8a88c19a807508bae7df624f7b629a91f4/podpac/core/utils.py#L349) doesn't have a well defined return.
> > Should we return `None` like in the other catch?
> 
> Very yes. Nice catch.

Roger that. I went ahead and committed that change already (it made the unit test easier to write).

PR #540, opened 2026-02-06
Release/4.0.0

dhinckley-creare:
Merging updates from the release branch back into develop

### DIFF
--- a/podpac/core/authentication.py
+++ b/podpac/core/authentication.py
@@ -8,13 +8,15 @@ import logging
 
 import requests
 import traitlets as tl
-from lazy_import import lazy_module, lazy_function
+from lazy_import import lazy_module
 
 from podpac.core.settings import settings
 from podpac.core.utils import cached_property
 
 # Optional dependencies
-pydap_setup_session = lazy_function("pydap.cas.urs.setup_session")
+# see pydap_source.py for import note
+# pydap_setup_session = lazy_function("pydap.cas.urs.setup_session")
+from pydap.cas.urs import setup_session as pydap_setup_session
 
 _log = logging.getLogger(__name__)
 _USERNAME_AT = "username@{}"
--- a/podpac/core/cache/cache_ctrl.py
+++ b/podpac/core/cache/cache_ctrl.py
@@ -9,13 +9,15 @@ from podpac.core.cache.ram_cache_store import RamCacheStore
 from podpac.core.cache.disk_cache_store import DiskCacheStore
 from podpac.core.cache.s3_cache_store import S3CacheStore
 import traitlets as tl
+import logging
+
 
 
 _CACHE_STORES = {"ram": RamCacheStore, "disk": DiskCacheStore, "s3": S3CacheStore}
 
 _CACHE_NAMES = {RamCacheStore: "ram", DiskCacheStore: "disk", S3CacheStore: "s3"}
 
-_CACHE_MODES = ["ram", "disk", "network", "all"]
+_CACHE_MODES = ["ram", "disk", "s3", "all"]
 
 # Error messages used in 3 or more places
 _INVALID_NODE = "Invalid node (must be of type Node, not '%s')"
@@ -24,6 +26,7 @@ _INVALID_COORDS = "Invalid coordinates (must be of type 'Coordinates', not '%s')
 _INVALID_MODE = "Invalid mode (must be one of %s, not '%s')"
 _INVALID_ITEM_ASTERISK = "Invalid item ('*' is reserved)"
 
+_logger = logging.getLogger(__name__)
 
 def get_default_cache_ctrl():
     """
@@ -63,7 +66,15 @@ def make_cache_ctrl(names):
         if name not in _CACHE_STORES:
             raise ValueError("Unknown cache store type '%s', options are %s" % (name, list(_CACHE_STORES)))
 
-    return CacheCtrl([_CACHE_STORES[name]() for name in names])
+    cache_stores = []
+    for name in names:
+        try:
+            cache_store = _CACHE_STORES[name]()
+            cache_stores.append(cache_store)
+        except Exception as e:
+            _logger.warning("Cannot create cache_store of type {} -- error={}".format(name, e))
+
+    return CacheCtrl(cache_stores)
 
 
 def clear_cache(mode="all"):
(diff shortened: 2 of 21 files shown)


podpac/core/cache/cache_ctrl.py
```diff
@@ -63,7 +66,15 @@ def make_cache_ctrl(names):
         if name not in _CACHE_STORES:
             raise ValueError("Unknown cache store type '%s', options are %s" % (name, list(_CACHE_STORES)))
 
-    return CacheCtrl([_CACHE_STORES[name]() for name in names])
+    cache_stores = []
```
mpu-creare, line 69:
```suggestion
    # makes all requested cache stores and fails gracefully if one of the stores is unavailable
    cache_stores = []
```

podpac/core/data/csv_source.py
```diff
@@ -85,7 +85,7 @@ def _default_outputs(self):
     # -------------------------------------------------------------------------
 
     def open_dataset(self, f):
-        return pd.read_csv(f, parse_dates=True, infer_datetime_format=True, header=self.header)
```
mpu-creare, line 88: What was the problem with `infer_datetime_format` ? New pandas behavior? Seems like a potentially important flag...

podpac/core/data/datasource.py
```diff
@@ -263,7 +263,7 @@ def _get_data(self, rc, rci):
 
         """
         # get data from data source at requested source coordinates and requested source coordinates index
-        data = self.get_data(rc, rci)
+        data = deepcopy(self.get_data(rc, rci))
```
mpu-creare, line 266: Why the deepcopy? that's rarely needed...

podpac/core/data/h5py_source.py
```diff
@@ -52,6 +52,8 @@ class H5PY(FileKeysMixin, BaseFileSource):
     file_mode = tl.Unicode(default_value="r").tag(readonly=True)
     array_dims = tl.List(trait=tl.Unicode()).tag(readonly=True)
 
+    coordinate_index_type = "slice"
```
mpu-creare, line 55: I'm surprised that this was missed.

podpac/core/test/test_units.py
```diff
@@ -564,7 +567,7 @@ def test_to_geotiff_roundtrip_1band(self):
             fp.write(b"a")  # for some reason needed to get good comparison
 
             fp.seek(0)
-            rnode = Rasterio(source=fp.name, outputs=node.outputs)
+            rnode = Rasterio(source=fp.name, outputs=node.outputs, crs="EPSG:4326")
```
mpu-creare, line 570: The CRS should be picked up automatically from the file -- what was the issue here?

podpac/core/test/test_utils.py
```diff
@@ -831,7 +838,7 @@ def test_key_not_in_params_upper_not_in_params(self):
 class TestGetFromUrl:
     def test_raise_requests_error(self):
         mock_requests = MagicMock()
-        mock_requests.get.side_effect = ConnectionError("Test Connection Error")
+        mock_requests.get.side_effect = Exception("Test Connection Error")
```
mpu-creare, line 841: `ConnectionError` seems like a better `Exception` type -- why the change?

podpac/core/authentication.py
```diff
@@ -8,13 +8,15 @@
 
 import requests
 import traitlets as tl
-from lazy_import import lazy_module, lazy_function
+from lazy_import import lazy_module
```
mpu-creare, line 11: you can comment this import.

podpac/core/utils.py
```diff
@@ -341,11 +340,7 @@ def _get_from_url(url, session=None):
                     url, r.status_code, r.text
                 )
             )
-
-    except requests.ConnectionError as e:
-        _log.warning("Cannot connect to {}:".format(url) + str(e))
-        r = None
-    except RuntimeError as e:
+    except Exception as e:
```
mpu-creare, line 343: Same comment about `ConnectionError` being a better exception type.

mpu-creare (approved):
Couple of minor comments. Mostly looks good to me.

podpac/core/test/test_utils.py
dhinckley-creare: Pytest claimed that it was improper to catch something that did not inherit from Exception and would fail the test

podpac/core/utils.py
dhinckley-creare: Because pytest wouldn't permit the ConnectionError, it meant more than just RuntimeErrors could be raised. This change was to handle that, and acknowledged that only the wording of the error message changed with the different except cases.

podpac/core/test/test_units.py
dhinckley-creare: For these unit tests, if crs wasn't specified, Rasterio would end up with this very verbose CRS, and I'm not sure where it was coming from.

podpac/core/data/datasource.py
dhinckley-creare: This later line
```
        udata_array.data[np.isin(udata_array.data, self.nan_vals)] = self.nan_val
```
would get "read-only" permission errors if the original get_data was used for one of the unit tests.

podpac/core/data/csv_source.py
dhinckley-creare: For the version of pandas that was installed when pip installing podpac, that argument no longer existed.

I suspect me using python 3.11.12 for this effort may have caused me to do some unexpected "newer python" updating. This and the "zarr<3" in setup seem to both be python version upgrade-like changes.

podpac/core/cache/cache_ctrl.py
dhinckley-creare: I got the approval notice before these suggestions and jumped the gun a bit. I'll get this into the branch that is still merging into main.

podpac/core/authentication.py
dhinckley-creare: done

podpac/core/data/pydap_source.py
```diff
@@ -12,20 +12,26 @@
...
-pydap = lazy_module("pydap")
-lazy_module("pydap.client")
-lazy_module("pydap.model")
+# pydap = lazy_module("pydap")
+# lazy_module("pydap.client")
+# lazy_module("pydap.model")
+# lazy_class("pydap.__spec__")
+
+# Lazy loading was conflicting with xarray access of pyap.__spec__
+import pydap
```
mpu-creare, line 31: I just realized we can still make pydap an optional dependency using the pattern: 


```suggestion
try: 
    import pydap
    import pydap.model
    import pydap.client
except: ImportError
    class Dum:
        pass
    pydap = Dum()
    pydap.model =None
    pydap.client = None
```

podpac/core/test/test_utils.py
mpu-creare: REally?. I just checked the 'requests' code, and it seems like the inheritance is `Exception --> IOError --> RequestException --> ConnectionError`... so it DOES inherit from Exception. 

Also
```python
import requests
assert isinstance(requests.ConnectionError(), Exception)
```

PR #541, opened 2026-02-06
Release/4.0.0

dhinckley-creare:
Merging fixes from release branch into main for formal tagged release


dhinckley-creare:
@mpu-creare As far as I can tell, although the change log was updated, the 4.0.0 changes (hash_cache, zarr_cache, etc) never made their way to main before now.

So, should this still be the 4.0.0 tagged release because it is the first of this new form or, because this effort did follow a different effort that considered itself 4.0.0, should this be a 4.1.0 release?

mpu-creare:
Since we don't have a 4.0.0 tag at all, this should be the first 4.0.0 release.

.github/PULL_REQUEST_TEMPLATE/release_template.md
```diff
@@ -0,0 +1,95 @@
+# Release
+
+How to release `podpac`
+
+## Branch Notes
+There are currently two actively maintained versions of PODPAC:
```
mpu-creare: We need to fix this -- after this merge we will no longer be maintaining a 3.x version.

doc/source/api.rst
```diff
@@ -242,7 +242,6 @@ Utilities
     podpac.utils.create_logfile
     podpac.utils.clear_cache
     podpac.utils.cached_property
-    podpac.utils.NoCacheMixin
     podpac.utils.DiskCacheMixin
```
mpu-creare: Need to remove this from the api docs.

podpac/core/cache/utils.py
```diff
@@ -46,13 +46,13 @@ def expiration_timestamp(value):
     if isinstance(expires, string_types):
         try:
             expires = np.datetime64(expires).item()
-        except:
+        except Exception:
```
mpu-creare: I'd love to catch a more specific exception... I also would have loved a comment about why it's ok to pass

podpac/core/interpolation/test/test_interpolation_manager.py
```diff
@@ -174,7 +174,7 @@ def test_init_interpolators(self):
             InterpolationManager([{"method": "nearest", "params": {"spatial_tolerance": "tol"}}])
 
         with pytest.raises(AttributeError):
-            assert interp.config[("default",)]["interpolators"][0].myarg == "tol"
+            interp.config[("default",)]["interpolators"][0].myarg
```
mpu-creare, line 177: This test changed! Why are we checking for an attribute error when we should be checking if the tolerance we tried to set on the interpolator was actually set on the interpolator. This needs an update.

podpac/core/interpolation/test/test_interpolation_manager.py
```diff
@@ -334,7 +334,9 @@ def interpolate(self, udims, source_coordinates, source_data, eval_coordinates,
         assert np.all(outdata == srcdata)
 
 
-class TestHeterogenousInterpolation(object):
+class TestHeterogenousInterpolation(
+    object
+):  # I don't think heterogeneou interpolation works with current .interpolate() scheme.
```
mpu-creare: replying to the commented code -- the fact that the unit test works implies that heterogenous still works?

podpac/core/interpolation/interpolation.py
```diff
@@ -129,6 +95,15 @@ class will be used without modification.
     _requested_source_data = tl.Instance(UnitsDataArray)
     _evaluated_coordinates = tl.Instance(Coordinates)
 
+    # def __getattr__(self, name):
```
mpu-creare: Can we remove this comment block?

podpac/core/managers/test/test_parallel.py
```diff
@@ -40,7 +40,7 @@ def test_parallel_multi_thread_compute_fill_output2(self):
 
         np.testing.assert_array_equal(o, o_p)
 
-    @pytest.mark.skipif(sys.version < "3.7", reason="python < 3.7 cannot handle processes launched from threads")
+    @pytest.mark.skipif(sys.version_info[0:2] < (3,7), reason="python < 3.7 cannot handle processes launched from threads")
```
mpu-creare: Since Python 3.7 has reached end of life, perhaps we can remove these skip conditions.

podpac/core/test/test_utils.py
```diff
@@ -74,28 +79,28 @@ def test(self):
...
             d = OrderedDictTrait()
 
-        m = MyClass(d=OrderedDict([("a", 1)]))
+        MyClass(d=OrderedDict([("a", 1)]))
 
         with pytest.raises(tl.TraitError):
             MyClass(d=[])
 
-    @pytest.mark.skipif(sys.version < "3.6", reason="python < 3.6")
+    @pytest.mark.skipif(sys.version_info[0:2] < (3, 6), reason="python < 3.6")
```
mpu-creare: Since Python 3.6 is EOL, perhaps we can remove these checks as well.

podpac/core/utils.py
```diff
@@ -126,7 +125,7 @@ def create_logfile(
     return log, handler, formatter
 
 
-if sys.version < "3.6":
+if sys.version_info[0:2] < (3, 6):
```
mpu-creare: Since 3.6 is EOL, we can remove this branch.

CHANGELOG.md
```diff
@@ -1,11 +1,88 @@
...
+which nodes are cached, and how they are cached. The functionality moved away from the parent `Node` class, and
+now resides in the `CacheNode` child classes. Now caching is just part of a processing pipeline. For convenience,
+we added the `Node.cache` method, which will add a `CacheNode` to a `Node`, which basically restores the original
+functionality.
+
+We also removed the `InterpolationMixin` node that restored the functionality of every certain `DataSource`
+nodes being automatically interpolated on `eval`. Now users have to create `Interpolation` nodes explicitly. To make the transition smoother, we added a convenience function `Node.interpolate` that will add an `Interpolation` node to the pipeline, restoring the original functionality.
+
+### Features
+* Added the `ZarrCache`. This node enables partial caching of nodes based on the source's coordinate. This means you can
```
mpu-creare, line 15: Did we miss anything here? Should we mention the code refactor to reduce cognitive complexity? Anything else that's new? Bugfixes, etc.

mpu-creare (approved):
Again, I have a few minor comments. Mostly to clean things up. I think it's probably worth fixing the unit test that changed.

doc/source/api.rst
dhinckley-creare: I don't see this line in the release branch.

podpac/core/cache/utils.py
dhinckley-creare: From some brief testing, it seems failures here are always ValueErrors (I couldn't fashion a string that failed in any other way).

The catches just passing the value through is handled by the last isinstance check, which throws an error is a string went un-converted. I think a try-pass was adopted to handle the multiple different string parsing methods (because there are 2 that both fail via error, the try-pass avoids annoying case logic).

podpac/core/interpolation/test/test_interpolation_manager.py
dhinckley-creare: I think this test always only ever tested that attempted to access a non-existent attribute resulted in error. "myarg" was never defined and existed nowhere else in the repo.

I think the "tol" part may have always been pointless.

podpac/core/interpolation/test/test_interpolation_manager.py
dhinckley-creare: It seems so.

podpac/core/interpolation/interpolation.py
dhinckley-creare: Removed

.github/PULL_REQUEST_TEMPLATE/release_template.md
dhinckley-creare: Updated both release_template.md and RELEASE.md to remove mention of podpac3

podpac/core/managers/test/test_parallel.py
dhinckley-creare: Will remove.

Should setup.py update the python_requires as well? Perhaps to the oldest still supported python? Or should it be left to acknowledge that podpac is/may be compatible with older python?

podpac/core/test/test_utils.py
dhinckley-creare: Removed

podpac/core/utils.py
dhinckley-creare: Removed.

The `OrderedDictTrait = tl.Dict` has also been moved to the top of the file.

CHANGELOG.md
dhinckley-creare: I added the 1 new bugfix and 1 new breaking change of which I'm aware.

Most of the changes in the PR come from a previous effort that didn't make its way to a full formal/tagged release.

podpac/core/cache/utils.py
mpu-creare: Please add a comment so we don't have to figure this out again in the future.

doc/source/api.rst
mpu-creare: As far as I can tell, `podpac.utils.DiskCacheMixin` remains in this file?

podpac/core/cache/utils.py
dhinckley-creare: Is the one on line 46 sufficient or does it need more background?

doc/source/api.rst
dhinckley-creare: Ah, I thought you were pointing out the NoCacheMix (I saw the highlighting in the diff and mistook it for a github highlight thing). Removed

PR #542, opened 2026-02-13
Use pytest's built-in tmp_path

kxm-creare:
There are a few tests that create a temporary directory within the PODPAC package itself. These tests can't pass if the package's location is not writeable.  I've updated these tests to use pytest's built-in `tmp_path` instead.


dhinckley-creare:
All unit tests pass in a fresh pyenv, merging.

PR #544, opened 2026-03-05
Bugfix/interpolation dimensions

scranford1:
### Changes

- Remove squeeze from Stacked Coordinates to avoid issues with axes being removed unintentionally in the case of single value coordinates.
- Use _udrop_ instead of _drop_ for extra dimensions as the dimensions are found using _udims_ method.
- Allow rasterio interpolator to handle additional CRS formats as coordinates can use PROJ4 or WKT according to documentation.
```
crs : str, optional
            Coordinate reference system. Supports PROJ4 and WKT.
```

### DIFF
--- a/podpac/core/coordinates/stacked_coordinates.py
+++ b/podpac/core/coordinates/stacked_coordinates.py
@@ -329,7 +329,7 @@ class StackedCoordinates(BaseCoordinates):
             dtype = dtypes[0]
         else:
             dtype = object
-        return np.dstack([c.coordinates.astype(dtype) for c in self._coords]).squeeze()
+        return np.stack([c.coordinates.astype(dtype) for c in self._coords], axis=-1)
 
     @property
     def xcoords(self):
@@ -468,7 +468,6 @@ class StackedCoordinates(BaseCoordinates):
                 start = index.start
             return stop - start
         return len(index)
-
 
     def _and_indices(self, indices: List["Coordinates"]):
         """logical AND of the selection in each dimension"""
--- a/podpac/core/coordinates/test/test_stacked_coordinates.py
+++ b/podpac/core/coordinates/test/test_stacked_coordinates.py
@@ -783,12 +783,14 @@ class TestStackedCoordinatesMethods(object):
         sc_3 = StackedCoordinates([lat[::-1], lon], name="lat_lon")  # same coordinates, paired differently
         sc_t = sc.transpose("lon", "lat")
         sc_time = StackedCoordinates([lat, lon, time], name="lat_lon_time")
+        sc_single_value = StackedCoordinates([lat[0], lon[0]], name="lat_lon")
 
         assert sc.issubset(sc)
         assert sc[:2].issubset(sc)
         assert not sc.issubset(sc[:2])
         assert not sc_2.issubset(sc)
         assert not sc_3.issubset(sc)
+        assert sc_single_value.issubset(sc_single_value)
 
         assert sc_t.issubset(sc)
 
--- a/podpac/core/coordinates/test/test_stacked_coordinates.py
+++ b/podpac/core/coordinates/test/test_stacked_coordinates.py
@@ -791,6 +791,7 @@ class TestStackedCoordinatesMethods(object):
         assert not sc_2.issubset(sc)
         assert not sc_3.issubset(sc)
         assert sc_single_value.issubset(sc_single_value)
+        assert sc_single_value.issubset(sc)
 
         assert sc_t.issubset(sc)
 
--- a/podpac/core/interpolation/interpolation.py
+++ b/podpac/core/interpolation/interpolation.py
@@ -122,12 +122,12 @@ class Interpolate(Node):
 
     @property
     def outputs(self):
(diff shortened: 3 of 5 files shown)


scranford1:
Unit tests are now passing at same percentage as develop. (18 failed, 987 passed, 48 skipped, 9 deselected, 2 xfailed)

scranford1:
@dhinckley-creare Tagging you in this pull request as well, mostly related to stacked coordinate changes and one change for CRS formats.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -329,7 +329,7 @@ def coordinates(self):
             dtype = dtypes[0]
         else:
             dtype = object
-        return np.dstack([c.coordinates.astype(dtype) for c in self._coords]).squeeze()
+        return np.stack([c.coordinates.astype(dtype) for c in self._coords], axis=-1)
```
dhinckley-creare, line 332: @scranford1 this is the only line that I'm not too sure about.

What gives me pause here is that `squeeze` would crush a pointless dimension regardless of order but I'm not familiar enough with stacked_coordinates to know if `axis=-1` is always correct.

scranford1: From my understanding the squeeze is being used to remove the extra dimension that is introduced during the conversion to a 3D shape. However for single values it can end up removing a dimension which should be kept. With stack and axis=-1, 1D arrays are staked to go from (N,) -> (N,k) and 2-D Arrays are stacked to go from (N,M,) to (N,M,k) where k is number of arrays in stack.

podpac/core/coordinates/test/test_stacked_coordinates.py
```diff
@@ -783,12 +783,14 @@ def test_issubset(self):
...
         sc_t = sc.transpose("lon", "lat")
         sc_time = StackedCoordinates([lat, lon, time], name="lat_lon_time")
+        sc_single_value = StackedCoordinates([lat[0], lon[0]], name="lat_lon")
 
         assert sc.issubset(sc)
         assert sc[:2].issubset(sc)
         assert not sc.issubset(sc[:2])
         assert not sc_2.issubset(sc)
         assert not sc_3.issubset(sc)
+        assert sc_single_value.issubset(sc_single_value)
```
mpu-creare, line 795: Does this work? 
```suggestion
assert sc_single_values.issubset(sc)
```

podpac/core/coordinates/stacked_coordinates.py
mpu-creare: Looks good to me.

podpac/core/interpolation/interpolation.py
```diff
@@ -247,7 +247,7 @@ def _eval(self, coordinates, output=None, _selector=None):
         # if requested crs is differented than coordinates,
         # fabricate a new output with the original coordinates and new values
         if self._evaluated_coordinates.crs != coordinates.crs:
-            output = self.create_output_array(self._evaluated_coordinates.drop(extra_dims), data=output[:].values)
+            output = self.create_output_array(self._evaluated_coordinates.udrop(extra_dims), data=output[:].values)
```
mpu-creare, line 250: I'm surprised this was still in the code. It seems like the correct fix to me. Did you run all the unit tests and made sure this change didn't break something else?

podpac/core/interpolation/rasterio_interpolator.py
```diff
@@ -98,12 +98,12 @@ def interpolate(self, udims, source_coordinates, source_data, eval_coordinates,
 
         with rasterio.Env():
             src_transform = transform.Affine.from_gdal(*source_coordinates.geotransform)
-            src_crs = rasterio.crs.CRS.from_proj4(source_coordinates.crs)
+            src_crs = rasterio.crs.CRS.from_string(source_coordinates.crs)
```
mpu-creare, line 101: Nice, increased flexibility. Did you search podpac to see if we use from_proj4 anywhere else in the codebase? Worth doing.

mpu-creare (approved):
I had a couple of minor suggestions and comments. I don't need to see it again.

podpac/core/interpolation/interpolation.py
mpu-creare: I didn't see your comment about the unit tests passing rate being unchanged. Can you add a unit test that would expose this issue?

podpac/core/coordinates/test/test_stacked_coordinates.py
scranford1: I added it as another test case, it passes as expected.

podpac/core/interpolation/rasterio_interpolator.py
scranford1: Yes, `from_proj4` was only used in the rasterio interpolator.

podpac/core/interpolation/interpolation.py
scranford1: A unit test has been added to expose this for stacked coordinates.

PR #547, opened 2026-04-15
Updates to pacify linter and sonarqube

jdw-creare:
The vast bulk of this PR is removal of unused imports, and assignment operations that have no effect.

Five of these changes rise above that threshold, and deserve a little more attention on review:
* https://github.com/creare-com/podpac/commit/afcc5060638ee96cf4135e55e695c1bfbf384646
* https://github.com/creare-com/podpac/commit/716501293781f6166a15b87b4ee6999cdca48890
* https://github.com/creare-com/podpac/commit/461b344b6234216184163b02992ed7c4ae3fc6fd (Make improper formatting fail the pipeline)
* https://github.com/creare-com/podpac/commit/6085a9347a92f9644b100185c3a934f3c8444324
* https://github.com/creare-com/podpac/commit/fd6a0e674d557b79f88fd6f8fed5e62fc062a54e


podpac/core/algorithm/generic.py
```diff
@@ -113,7 +113,7 @@ def algorithm(self, inputs, coordinates):
         f_locals = dict(zip(fields, res))
 
         try:
-            from numexpr import evaluate  # Needed for some systems to get around lazy_module issues
+            import numexpr as ne  # Needed for some systems to get around lazy_module issues
```
mpu-creare: Wow! Weird longstanding bug. I bet this was never covered in a unit test

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -452,7 +452,7 @@ def select(self, bounds, outer=False, return_index=False):
         else:
             return self[index]
 
-    def _index_len(self, index: "Coordinates"):
+    def _index_len(self, index: "BaseCoordinates"):
```
mpu-creare: While we're updating type hints -- index seems to be a slice or "BaseCoordinates"... also, no return type hint, which should be an int.

podpac/core/coordinates/stacked_coordinates.py
```diff
@@ -469,7 +469,7 @@ def _index_len(self, index: "Coordinates"):
             return stop - start
         return len(index)
 
-    def _and_indices(self, indices: List["Coordinates"]):
+    def _and_indices(self, indices: List["BaseCoordinates"]):
```
mpu-creare: Same comment as for the above.

podpac/core/data/test/test_rasterio.py
```diff
@@ -78,23 +76,19 @@ def test_get_band_numbers(self):
...
-        for i,data in enumerate([new_coords_1,new_coords_2]) :
-            for a in ['lon','lat'] :
-                assert(np.isnan(data[a].coordinates).sum()==0) # nan check
-                assert(len(data[a])==expected_values[i][a]) # guard against old issue of return being trimmed
-
-
-        
\ No newline at end of file
+        window_1, new_coords_1 = node._get_window_coords(c2, c1)  # tests when 1 coord completely contains the other
+        window_2, new_coords_2 = node._get_window_coords(
```
mpu-creare: See SonarCloud finding.

podpac/core/data/test/test_wcs.py
```diff
@@ -89,6 +77,14 @@ def test_eval_nonuniform(self):
         assert output.shape == (100, 100)
         assert output.data.sum() == 1256581.0
 
+    def test_eval_nonuniform2(self):
+        c = COORDS[[0, 10, 99], [0, 99]]
+
+        node = MockWCS(source="mock", layer="mock").interpolate()
```
mpu-creare, line 83: Looks like this class was removed... where is it getting MockWCS from?

podpac/core/interpolation/interpolation.py
mpu-creare, line 168: I think this should be `selector` to match the main interface defined in `Node`.

podpac/core/interpolation/interpolation.py
```diff
@@ -203,7 +197,8 @@ def _eval(self, coordinates, output=None, _selector=None):
             Cannot evaluate these coordinates
         """
         self._set_interpolation()
-        _selector = self._interpolation.select_coordinates
+        if _selector is None:
+            _selector = self._interpolation.select_coordinates
```
mpu-creare: This is probably boilerplate from non-interpolation nodes, or not?!?

podpac/core/interpolation/interpolation.py
mpu-creare, line 213: This is probably the same selector variable that totally overwrites anything passed into the function... something's fishy here. Worth a closer look. There are some major subtleties with the selectors.

podpac/core/test/test_style.py
```diff
@@ -105,9 +104,9 @@ def test_eq(self):
         assert style1 is not style3
         assert style2 is not style3
 
-        assert style1 == style1
-        assert style2 == style2
-        assert style3 == style3
+        assert style1.__eq__(style1)
+        assert style2.__eq__(style2)
+        assert style3.__eq__(style3)
```
mpu-creare, line 109: What's up with this change?

mpu-creare (changes requested):
Please address SonarCloud findings (seem pretty minor). Bonus points if you add a unit test to cover that `ne.evaluate` branch of that one node. 

Put something on my calendar so we can look at the selector on the Interpolation node -- that requires some care.

podpac/core/test/test_style.py
jdw-creare: It's a bit annoying: SonarQube doesn't seem to understand the operator overloading here.  So it sees `assert style1 == style1` as a bug.  It thinks "oh you're checking an object against itself, that'll always evaluate to true".  But of course this test is a test of the `__eq__()` function, and we need to check that.  So this compromise makes SonarQube happy.

podpac/core/data/test/test_wcs.py
jdw-creare: It's up on line 37 of the original file, line 35 after the change.  It looks like someone copy-pasted, and maybe intended to make two fairly different versions of `TestWCS` and `MockWCS`, but never quite finished.

podpac/core/algorithm/generic.py
jdw-creare: It turns out this WAS covered by a unit test, but the reason it didn't fail was that there was a `lazy_module` import for `ne` at the top of the file.  So line 116 isn't actually necessary - I removed it in a recent commit!

jdw-creare:
@mpu-creare , I've made all the changes I plan to!  Did you want to take a look at it again before I merge it?  I also added VS Code dev container support to make it easier to run unit tests locally. (Since that was the easiest way to confirm unit test coverage for that thing we had questions about)

mpu-creare:
@jdw-creare I don't need to see it again. merge when ready.

jdw-creare:
Thanks Matt!

PR #549, opened 2026-04-22
Update formatting to pass flake8 rules, and added flake8 formatting gate

jdw-creare:
Added flake8 formatting gate, and used Claude Code to fix flake8's formatting findings

Note that in commit cca1356b61c91d17abeffe7a4e50a1f9dcb1e5fb, Claude found that complying with rules A001, A002, A003, and A005 would require changing the public-facing interface.  We may still want to make these changes, but i would like to propose that in a separate PR and discuss the pros and cons of that change. CC added those rules to the ignore list, but in the future I would prefer to remove them from the ignore list.

### DIFF
--- a/.github/workflows/gitlab-python-worflow.yml
+++ b/.github/workflows/gitlab-python-worflow.yml
@@ -25,7 +25,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          pip install --prefer-binary .[devall]
+          pip install --only-binary :all: .[devall]
 
       - name: Run Black Format Checks
         run: black --version; black --check --diff -l 120 . > black-report.txt
@@ -65,7 +65,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          pip install --prefer-binary .[devall]
+          pip install --only-binary :all: .[devall]
 
       - name: Run Pytest with Coverage
         run: |
@@ -106,7 +106,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          cd podpac && pip install --prefer-binary .[devall]
+          cd podpac && pip install --only-binary :all: .[devall]
 
       - name: Run doc tests
         run: |
--- a/.github/workflows/gitlab-python-worflow.yml
+++ b/.github/workflows/gitlab-python-worflow.yml
@@ -105,7 +105,7 @@ jobs:
 
       - name: Install Dependencies
         run: |
-          pip install --upgrade pip setuptools && pip install .[devall]
+          cd podpac && pip install --upgrade pip setuptools && pip install .[devall]
 
       - name: Run doc tests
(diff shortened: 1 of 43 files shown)


jdw-creare:
This branch also passes integration tests, if that affects anything

.github/workflows/gitlab-python-worflow.yml
```diff
@@ -90,11 +101,11 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@v3
         with:
-          python-version: "3.7"
+          python-version: "3.12"
 
       - name: Install Dependencies
         run: |
-          pip install --upgrade pip setuptools && pip install podpac[devall]
+          cd podpac && pip install --upgrade pip setuptools && pip install .[devall]
```
mpu-creare: The cd is a weird change -- not sure why that needed to be modified... but I'm sure it's needed. I'm just surprised that it used to work.

podpac/core/units.py
```diff
@@ -476,16 +476,16 @@ def create(cls, c, data=np.nan, outputs=None, dtype=float, **kwargs):
     meth = "__{:s}__".format(tp)
 
     def make_func(meth, tp):
-        def func(self, other):
-            x = getattr(super(UnitsDataArray, self), meth)(other)
-            x2 = self._apply_binary_op_to_units(getattr(operator, tp), other, x)
+        def func(self, other):  # noqa: B023 -- meth/tp bound via make_func params
```
mpu-creare: i don't follow the comments

setup.py
```diff
@@ -72,36 +68,27 @@
...
+    ],
+}
 
 # set long description to readme
 with open("README.MD") as f:
     long_description = f.read()
 
 all_reqs = []
 for key, val in extras_require.items():
-    if "key" == "dev":
```
mpu-creare, line 90: Ouch! How did that nasty bug get in there. Nice catch.

mpu-creare (approved):
Couple of comments, nothing major. Merge when ready.

setup.py
jdw-creare: Claude caught this one!

.github/workflows/gitlab-python-worflow.yml
jdw-creare: I found that, without the `cd`, the command failed when i switched from `pip install podpac` to `pip install .`.  I had some reason to suspect that that command was reaching out to the PyPi distribution of podpac, rather than the latest commit of podpac.  So I changed it to `pip install .` so there was no way it would consider reaching out to the PyPi version.

podpac/core/units.py
jdw-creare: I've updated the instances of `# noqa` to actually fix the finding

PR #548, opened 2026-04-21
Update formal support from python 3.7/3.8 to python 3.12

jdw-creare:

PR #550, opened 2026-04-24
Changes to remove shadowing of builtins, which makes breaking changes to the API

jdw-creare:
Note that this builds on #548 and #549, so those should be merged in prior to this one.  Merging this will need to be discussed with the whole team, and coordinated with projects that rely on it.

Noteworthy changes:
* 6c4973f8358d5709b5c5f95a0c31d5868ab2ce33 Rename `Node.eval` and `_eval` to `evaluate` and `_evaluate`
* cc8aee415a34d003c187e3455d4d81cc6a0308b7 Rename `hash` to `hash_id`
* 75d9374ec86668278b5c940895a5a79bf0c31e57 Rename `signal.py` to `convolution.py`
* 37202eef914a7a4985622bcf496105f0cf3c7f73 Renamed arguments named `format`, `type`, and `value`

This PR also updates Sphinx to version 3.5 to help with shadowing of the variable name `copyright`.


jdw-creare:
This is not necessary; we can use `noqa` on these errors.

PR #551, opened 2026-05-01
Added `noqa` to intentional "shadowing" of builtins

jdw-creare:
There are a few findings from the `flake8-builtins` plugin that seem unnecessary.  We think it makes sense to have, for example, methods named things like `eval` and `hash_id`.  In some cases (namely `hash_id`) these are overrides, not "shadowing".  

This PR adds `# noqa` comments to exclude those rules in specific locations where we think it makes sense to ignore those linting rules.  It also fixes some small errors highlighted by sonarqube and VS Code.


mpu-creare (approved):
Looks good to me. Please proceed.

jdw-creare:
Thanks @mpu-creare !

PR #552, opened 2026-05-11
Fix sonarqube findings

jdw-creare:
The most significant change here is to switch the example dockerfile from Conda to vanilla python.

This PR also fixes three kinds of findings:
* Commented-out code
* Branches in `if` statements that were only `pass`
* Variable assignment where the value is never subsequently referenced


podpac/core/coordinates/array_coordinates1d.py
```diff
@@ -62,18 +62,15 @@ def __init__(self, coordinates, name=None, **kwargs):
...
+        if self.coordinates.size == 1:
             self._is_monotonic = True
 
         elif self.coordinates.ndim > 1:
             self._is_monotonic = None
             self._is_descending = None
             self._is_uniform = None
 
-        else:
+        elif self.coordinates.size != 0:
```
mpu-creare, line 73: I personally find the if - pass more readable in this case... but oh well.

podpac/core/coordinates/coordinates.py
```diff
@@ -653,14 +653,13 @@ def __getitem__(self, index):
...
         # coerce
-        if isinstance(c, BaseCoordinates):
-            pass
-        elif isinstance(c, Coordinates):
-            c = c[dim]
-        elif "_" in dim:
-            c = StackedCoordinates(c)
-        else:
-            c = ArrayCoordinates1d(c)
+        if not isinstance(c, BaseCoordinates):
```
mpu-creare, line 656: oh interesting... in this case disallowing 'pass' creates an extra nest -- interesting.

podpac/core/coordinates/polar_coordinates.py
```diff
@@ -197,17 +197,3 @@ def full_definition(self):
 
     def copy(self):
         return PolarCoordinates(self.center, self.radius, self.theta, dims=self.dims)
-
-    # TODO return PolarCoordinates when possible
```
mpu-creare, line 201: I generally dislike getting rid of TODOs -- this is probably a great hint for some feature we didn't have time to implement but thought about when this was written... I suppose it's old enough that it no longer serves as a hint. Don't mind me, I'm just ranting today.

podpac/__init__.py
```diff
@@ -22,9 +22,7 @@ def makedirs(name, mode=511, exist_ok=False):
     try:
         _osmakedirs(name, mode)
     except OSError as e:
-        if exist_ok:
-            pass
-        else:
+        if not exist_ok:
```
mpu-creare, line 25: Here's an instance I wholeheartedly agree with!

Dockerfile
```diff
@@ -1,22 +1,17 @@
...
-
-# Create a podpac anaconda environment and activate
-RUN conda init bash && . ~/.bashrc \
-    && conda create --yes --name podpac python=3 anaconda \
-    && conda activate podpac
-
+    wget && \
+    apt-get clean
+COPY . /opt/podpac
+WORKDIR /opt/podpac
```
mpu-creare, line 15: Looks the old image gets podpac from PYPI, whereas the new version installs it from the local repository. Not sure I really care, just noting the difference.

mpu-creare (approved):
Just a few ranting comments, please proceed with the merge.

Dockerfile
jdw-creare: Absolutely!  I should've called that out in a commit message or something.  I wanted to make sure it built with the very latest code.  Of course, now it comes with the prerequisite that you build the dockerfile from within a full checkout of the repo, whereas before you could just build the dockerfile separately if you wanted to.

PR #553, opened 2026-05-20
Fixed timedelta typing issue, resolved warnings and errors.

dhinckley-creare:
This PR includes:
- removal of log messages about "dataset.crs.is_valid does nothing"
- replaced print statements with logging statements
- resolved issue where dtype wasn't specified in coordinates1d.select
- fixed timedelta handling in utils.py
- added unit test for exception case of to_geotiff with nonuniform coordinates
  - removed log warning message about limitation
- rolled back exception type specification in s3_cache_store because it cause implicit requirement for botocore install


podpac/core/cache/s3_cache_store.py
```diff
@@ -59,7 +58,7 @@ def __init__(self, s3_bucket=None, aws_region_name=None, aws_access_key_id=None,
 
         try:
             self._s3_client.head_bucket(Bucket=self._s3_bucket)
-        except (botocore.exceptions.BotoCoreError, botocore.exceptions.ClientError) as e:
+        except Exception as e:
             raise e
```
jdw-creare: I think this is equivalent to removing the whole `try` block, right? if we're re-raising every exception, may as well remove the `try` too.  (This will fix a flake8 B902 finding too)

podpac/core/interpolation/test/test_interpolation_manager.py
```diff
@@ -64,7 +67,7 @@ def test_dict_definition(self):
 
         # handle string methods
         interp = InterpolationManager({"method": "nearest", "dims": ["lat", "lon"]})
-        print(interp.config)
+        _log.debug(interp.config)
```
jdw-creare, line 70: excellent!

podpac/core/interpolation/test/test_interpolators.py
```diff
@@ -339,8 +342,8 @@ def test_spatial_tolerance(self):
         coords_dst = Coordinates([[1, 1.2, 1.5, 5, 9]], dims=["lat"])
         output = node.eval(coords_dst)
 
-        print(output)
-        print(source)
+        _log.debug(output)
+        _log.debug(source)
         assert isinstance(output, UnitsDataArray)
```
jdw-creare, line 347: love it

podpac/core/interpolation/test/test_interpolators.py
```diff
@@ -798,7 +801,7 @@ def test_interpolate_scipy_grid(self):
 
         assert isinstance(output, UnitsDataArray)
         assert np.all(output.lat.values == coords_dst["lat"].coordinates)
-        print(output)
+        _log.debug(output)
```
jdw-creare, line 804: a righteous extermination of `print` statements

podpac/core/managers/aws.py
```diff
@@ -1599,7 +1599,7 @@ def __init__(self, aws_access_key_id=None, aws_secret_access_key=None, region_na
         try:
             _ = self.get_account_id()
         except botocore.exceptions.ClientError as e:
-            _log.error(
+            _log.exception(
```
jdw-creare, line 1602: I think that with `exception`, you can leave out the part that incorporates the exception, and it'll do that automatically: https://docs.python.org/3/library/logging.html#logging.Logger.exception.

podpac/core/managers/test/test_multithreading.py
```diff
@@ -65,11 +68,11 @@ def test_request_release_threads_single_threaded(self):
 
     def test_request_release_threads_multi_threaded(self):
         def f(s):
-            print("In", s)
+            _log.debug("In", s)
             n1 = thread_manager.release_n_threads(s)
             time.sleep(0.05)
             n2 = thread_manager.release_n_threads(s)
-            print("Released", s)
+            _log.debug("Released", s)
```
jdw-creare: i'm pleased to see all these `print()` statements go away

dhinckley-creare:
The "Fixed timedelta" portion of this PR has been reverted as it was an improper band-aid I implemented after trying to eval a node which had timedelta coordinates using float coordinate values.

jdw-creare (approved):
There are two tiny tweaks that i'd suggest, along with pacifying the automatic checks.  I recently updated the formatting checker to look for additional things, so unfortunately it triggers much more often!

podpac/core/cache/s3_cache_store.py
dhinckley-creare: Yup, you are right. I was too focused on the line itself to see the context around it. Removing try.

podpac/core/managers/aws.py
dhinckley-creare: Yep, you appear to be correct. Originally this was going to be a `_log.error` but VS code complained and said I should use `_log.exception`. I was unaware of that option before that moment. Removing last sentence of error str that include the exception text.

jdw-creare:
@dhinckley-creare and @mpu-creare : I had Claude make unit tests for `managers/aws.py`, which addresses SonarQube's main complaint about coverage.  This file has gone uncovered for some time, so I figure some measure of test coverage is better than nothing. Take a look and feel free to merge when you feel ready!

PR #554, opened 2026-05-25
Change fake access key ID to avoid tripping Trufflehog

jdw-creare:
Trufflehog thinks this is a secret because it matches a regular expression:

https://github.com/trufflesecurity/trufflehog/blob/main/pkg/detectors/aws/access_keys/accesskey.go#L78

(At some point in the future maybe we want trufflehog to be part of our CI pipeline)

### DIFF
--- a/podpac/core/managers/test/test_aws.py
+++ b/podpac/core/managers/test/test_aws.py
@@ -60,8 +60,8 @@ class _MockBoto3Session(boto3_real.Session):
 
     def __init__(self):
         super().__init__(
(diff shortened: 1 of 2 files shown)


jdw-creare:
Hey @mpu-creare , this is a small one, let me know what you think!

mpu-creare (approved):
Looks fine to me. I do find it helpful to have a more realistic example key, but I understand the change.

jdw-creare:
Thanks Matt!

PR #555, opened 2026-06-01
Release/4.0.2

dhinckley-creare:
### Bugfixes
* Replaced `print` statements with `logging` statements.
* Fixed implicit `botocore` environment requirement. 
* Clarified error statement when input coordinates do not match expected `dtype`.
* Removed old `crs.is_valid` calls.
* Set Rasterio geotiff export `nodata` value to `np.nan` in `podpac.core.units.to_geotiff` for consistency.


doc/source/conf.py
```diff
@@ -67,7 +65,7 @@
 
 # General information about the project.
 project = "PODPAC"
-copyright = "2017-{}, Creare".format(datetime.datetime.now().year)
+project_copyright = "2017-{}, Creare".format(datetime.datetime.now().year)
```
mpu-creare, line 68: Is "project_copyright" the new "magic" variable name? Otherwise I'm expecting to see renaming of 'copyright' --> 'project_copyright' in at least 1 place...

podpac/alglib/__init__.py
```diff
@@ -6,3 +6,7 @@
 """
 
 from podpac.alglib import climatology
+
+__all__ = [
+    "climatology",
```
mpu-creare, line 11: Is this a trick to satisfy SonarQube?

podpac/core/cache/cache_ctrl.py
```diff
@@ -72,7 +70,7 @@ def make_cache_ctrl(names):
         try:
             cache_store = _CACHE_STORES[name]()
             cache_stores.append(cache_store)
-        except Exception as e:
+        except (ImportError, OSError, RuntimeError, TypeError, ValueError, CacheException) as e:
```
mpu-creare, line 73: How did we find all these exception types!

podpac/core/cache/s3_cache_store.py
```diff
@@ -52,16 +50,12 @@ def __init__(self, s3_bucket=None, aws_region_name=None, aws_access_key_id=None,
...
             aws_access_key_id=aws_access_key_id,
             aws_secret_access_key=aws_secret_access_key,
         )
         self._s3_bucket = s3_bucket
 
-        try:
-            self._s3_client.head_bucket(Bucket=self._s3_bucket)
-        except Exception as e:
-            raise e
+        self._s3_client.head_bucket(Bucket=self._s3_bucket)
```
mpu-creare, line 58: Why do we no longer need to catch the exception? 

it's pretty moot because I don't think anyone uses the S3 cache store.

podpac/core/cache/zarr_cache.py
```diff
@@ -109,7 +109,7 @@ def _default_group_data(self):
                 )  # adjust dtype as necessary
                 self._create_coordinate_zarr_dataset(group)
             return group
-        except Exception as e:
+        except (OSError, RuntimeError, ValueError, KeyError, tl.TraitError) as e:
```
mpu-creare, line 112: Again, how did we come up with this list?

podpac/core/coordinates/test/test_stacked_coordinates.py
```diff
@@ -783,12 +780,15 @@ def test_issubset(self):
         sc_3 = StackedCoordinates([lat[::-1], lon], name="lat_lon")  # same coordinates, paired differently
         sc_t = sc.transpose("lon", "lat")
         sc_time = StackedCoordinates([lat, lon, time], name="lat_lon_time")
+        sc_single_value = StackedCoordinates([lat[0], lon[0]], name="lat_lon")
```
mpu-creare: I seem to recall @dhinckley-creare  added this test because of a bug he found? It might be helpful to include a comment about that. Bonus points if it links to an issue.

podpac/core/data/rasterio_source.py
```diff
@@ -135,14 +131,11 @@ def get_coordinates(self):
...
-        elif isinstance(self.dataset.crs, rasterio.crs.CRS) and "init" in self.dataset.crs:
-            crs = self.dataset.crs["init"].upper()
-            if self.dataset.crs.is_valid:
-                validate_crs = False
-        elif isinstance(self.dataset.crs, dict) and "init" in self.dataset.crs:
+        elif (
+            isinstance(self.dataset.crs, rasterio.crs.CRS) or isinstance(self.dataset.crs, dict)
+        ) and "init" in self.dataset.crs:
             crs = self.dataset.crs["init"].upper()
-            if self.dataset.crs.is_valid:
```
mpu-creare, line 144: When combinding the two branches, it looks like this if statement got dropped, and it seems important?

podpac/core/interpolation/interpolation.py
```diff
@@ -247,7 +240,7 @@ def _eval(self, coordinates, output=None, _selector=None):
         # if requested crs is differented than coordinates,
         # fabricate a new output with the original coordinates and new values
         if self._evaluated_coordinates.crs != coordinates.crs:
-            output = self.create_output_array(self._evaluated_coordinates.drop(extra_dims), data=output[:].values)
+            output = self.create_output_array(self._evaluated_coordinates.udrop(extra_dims), data=output[:].values)
```
mpu-creare, line 243: This seems like a substantial change that warrants at least a comment in the code that explains why 'udrop' is correct and 'drop' is incorrect.

mpu-creare (approved):
I don't think I found everything. These are hard to review with formatting changes and non substantive changes mixed with small, subtle, relevant changes. 

I feel like I've reviewed a number of these types of changes recently. I hope we're close to the end with these... 

A few minor comments. There were a number of places where the generic "Exception" was replaces by a list of exceptions. I wasn't sure where those lists came from? I didn't flag them all. I point it out because that list could miss an exception that we should be catching, which can lead to the code breaking in production.

doc/source/conf.py
mpu-creare: I never saw any renames -- so either we need to rename something somewhere, or this is the new magic variable...

podpac/alglib/__init__.py
dhinckley-creare: As I understand it, yes. This gets around the "imported but unused" complaints.

podpac/core/cache/cache_ctrl.py
dhinckley-creare: @jdw-creare could you speak to this? Are these just the exceptions encountered by the existing unit tests?

podpac/core/cache/s3_cache_store.py
dhinckley-creare: The previous code caught the exception and then immediately raised it with no other actions.

I imagine we'd prefer to catch and then take some action but because this happens in the init, I can't think of what action should be taken.

podpac/core/coordinates/test/test_stacked_coordinates.py
dhinckley-creare: Nope, this is a @scranford1 update. @scranford1 can you describe what this addressed? At a quick glance, the naming suggests a previous issue with stacked coords of only 1 entry.

podpac/core/data/rasterio_source.py
dhinckley-creare: When the `self.dataset.crs.is_valid` ran, we saw a console log stating that the call was obsolete and didn't do anything (I don't have the exact text recorded anywhere).

podpac/core/interpolation/interpolation.py
dhinckley-creare: @scranford1 can you speak to this? I recall that it has something to do with stacked coordinates but that's as much as I know.

scranford1: This was a fix for stacked coordinates, see issue here with test added.
https://github.com/creare-com/podpac/pull/544
Comments: https://github.com/creare-com/podpac/pull/544/changes#r3053311977

Extra dimensions is found in the following manner:
`extra_dims = [d for d in coordinates.udims if d not in source_coords.udims]`

So dropping the coordinates required udims and not dims to match the expected dimensions.

podpac/core/coordinates/test/test_stacked_coordinates.py
scranford1: This was a fix for single stacked coordinates, see issue here with test added.
 https://github.com/creare-com/podpac/pull/544
Comments: https://github.com/creare-com/podpac/pull/544/changes#r2962114100

Stacked coordinates were being created with an inproper squeeze. This lead to values being scalar where an list was expected. The subset function would then error when it attempted to use iteration operations on a scalar.

mpu-creare: Can we drop a comment in the code? It's always helpful when a particular unit tests points to a specific bug.

podpac/core/data/rasterio_source.py
mpu-creare: That's fine. As long it was intentional. Resolving.

doc/source/conf.py
mpu-creare: Looks like a new magic variable name: https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-copyright  because of linting: https://github.com/sphinx-doc/sphinx/issues/8132

podpac/core/coordinates/test/test_stacked_coordinates.py
dhinckley-creare: Comment added. To keep the comment brief, I just referenced PR 544.

podpac/core/cache/cache_ctrl.py
jdw-creare: Ah yup, these lists were composed by Claude Code.  Unfortunately I don't know the criteria it used to determine the lists.  I do know that the unit tests still pass.  I did some spot checks too; descending through the calls to look for extraneous exceptions or missing ones, but it's hard to exhaustively check.

The motivation is based on a SonarQube rule.  The rule says something to the effect of "If you catch all exceptions, developers can make typos that go unnoticed."

jdw-creare:
@dhinckley-creare , this release is just the summation of changes that we've already reviewed, right?  Phrased differently, are there changes in this release that weren't introduced via a reviewed PR?  I figure those would be the sections (if any) that really need some attention.

dhinckley-creare:
> @dhinckley-creare , this release is just the summation of changes that we've already reviewed, right? Phrased differently, are there changes in this release that weren't introduced via a reviewed PR? I figure those would be the sections (if any) that really need some attention.

All of these changes have been reviewed in PRs for develop. My fear is that some changes were fine for develop but not so for main.

jdw-creare (approved):
OK great - I'm happy with all the PRs that have been merged in, and those have all been reviewed.

PR #557, opened 2026-07-23
Adding zarr3 support

BAyotte:


podpac/core/data/zarr_compat.py
```diff
@@ -0,0 +1,55 @@
...
+from lazy_import import lazy_module
+
+zarr = lazy_module("zarr")
+
+
+def _zarr_v3():
+    return int(zarr.__version__.split(".")[0]) >= 3
+
+
+def zarr_open(store, mode="r", **kwargs):
```
mpu-creare: This function doesn't actually do anything. I find including it to be confusing. I'd rather just take it out and use zarr.open directly.

podpac/core/data/zarr_compat.py
```diff
@@ -0,0 +1,55 @@
...
+    return zarr.open(store, mode=mode, **kwargs)
+
+
+def zarr_open_consolidated(store, mode="r", **kwargs):
```
mpu-creare: Same comment as above. I'm not convinced the exception catching give you anything more than what zarr would give you already. And returning "none" while swallowing the errors seems bad.

podpac/core/data/zarr_compat.py
```diff
@@ -0,0 +1,55 @@
...
+    try:
+        return zarr.open_consolidated(store, mode=mode, **kwargs)
+    except (KeyError, ValueError):
+        return None  # No consolidated metadata available
+
+
+def zarr_group():
```
mpu-creare: Same comment as above. Use zarr.group directly.

mpu-creare (commented):
It *looks* fine. I didn't actually test it, but the unit tests look good. I made some suggestions to help with code readability in the future. You can proceed.

PR #562, opened 2026-08-25
Fixing zarr3 issue for S3

BAyotte:
Fixing a type error (Filesystem needs to support async operations) related to zarr3 upgrade. Unit tests added as well.

### DIFF
--- a/podpac/core/data/zarr_compat.py
+++ b/podpac/core/data/zarr_compat.py
@@ -31,7 +31,7 @@ def create_zarr_array(group, name, chunks=None, **kwargs):
     return group.create_dataset(name, **kwargs)
 
 
-def _ensure_async_fs(fs: Any) -> Any:
+def _ensure_async_fs(fs: fsspec.spec.AbstractFileSystem) -> fsspec.spec.AbstractFileSystem:
     """Coerce an fsspec filesystem instance into one zarr 3's FsspecStore accepts.
 
     `S3Mixin` builds a synchronous `s3fs` instance, but zarr 3's `FsspecStore`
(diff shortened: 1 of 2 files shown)


podpac/core/data/zarr_compat.py
```diff
@@ -29,8 +29,24 @@ def create_zarr_array(group, name, chunks=None, **kwargs):
...
+    # which podpac's own "s3fs>=0.4" floor still technically allows), and warns even when the
+    # class is async-capable but this particular instance wasn't created with asynchronous=True.
+    # This mirrors zarr's own internal `_make_async` helper (used by FsspecStore.from_url/from_mapper).
+    if getattr(fs, "asynchronous", False):
+        return fs
+    if getattr(fs, "async_impl", False):
+        return type(fs)(*fs.storage_args, **{**fs.storage_options, "asynchronous": True})
+    from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
+
+    return AsyncFileSystemWrapper(fs, asynchronous=True)
```
scranford1, line 66: I am almost inclined to just have this function do the following. I assume the wrapper is harmless on already async filessytems?

```
def _ensure_async_fs(fs):
    from fsspec.implementations.asyn_wrapper import AsyncFileSystemWrapper
    return AsyncFileSystemWrapper(fs, asynchronous=True)
```

 Thoughts?

podpac/core/data/zarr_compat.py
```diff
@@ -29,8 +29,24 @@ def create_zarr_array(group, name, chunks=None, **kwargs):
     return group.create_dataset(name, **kwargs)
 
 
+def _ensure_async_fs(fs):
```
scranford1: Numpy style documentation and type hinting would be good for this and tests.

podpac/core/data/zarr_compat.py
scranford1: https://github.com/fsspec/filesystem_spec/blob/master/fsspec/implementations/asyn_wrapper.py

BAyotte: Its probably fine, I dont have a strong preference either way though. Which do you prefer?

scranford1 (approved):

PR #566, opened 2026-09-15
Bugfix/pydap dependency

scranford1:
pydap.cas.urs has been removed as of 3.5.11. Following guidance from https://urs.earthdata.nasa.gov/documentation/for_users/data_access/python, URS access is now done with a session that allows authorization headers to persistent to and from the NASA auth host.


dhinckley-creare (approved):
I don't see any issues.

PR #569, opened 2026-09-21
FIX: Use magic mock with root_marker

scranford1:
Fsspec's `AsyncFileSystemWrapper` reads `root_marker` via` inspect.getaattr_static`. This change ensures newer versions of rasterio using fsspec pass tests and do not raise AttributeError when `get_s3_store` tries to wrap the mock.


BAyotte (approved):
looks good!

